PR #4, opened 2024-07-10
Ghost Identity authentication

allouis:


src/app.ts
```diff
@@ -207,9 +209,12 @@ fedify.setObjectDispatcher(
 
 /** Hono */
 
+type GhostRole = 'Anonymous' | 'Owner' | 'Administrator' | 'Editor' | 'Author' | 'Contributor';
```
mike182uk, line 212: We could maybe use `enum` here so we don't have to worry about magic strings 

```ts
enum GhostRole {
  ANONYMOUS = 'Anonymous',
  OWNER = 'Owner',
  ADMINISTRATOR = 'Administrator',
  EDITOR = 'Editor',
  AUTHOR = 'Author',
  CONTRIBUTOR = 'Contributor'
}
```

src/app.ts
```diff
@@ -277,6 +282,61 @@ app.use(async (ctx, next) => {
...
+
+    const [match, token] = authorization.match(/Bearer\s+(.*)$/) || [null];
+
+    if (!match) {
+        throw new Error('Invalid Authorization header');
+    }
+
+    let protocol = 'https';
+    // We allow insecure requests when not in production for things like testing
+    if (process.env.NODE_ENV !== 'production' && !request.raw.url.startsWith('https')) {
```
mike182uk, line 308: Should the 2nd conditional be IF starts with `https`

src/app.ts
```diff
@@ -277,6 +282,61 @@ app.use(async (ctx, next) => {
...
+    });
+
+    const jwks = await jwksResponse.json();
+
+    const key = await jose.JWK.asKey(jwks.keys[0]);
+
+    try {
+        const claims = jwt.verify(token, key.toPEM());
+        if (typeof claims === 'string' || typeof claims.role !== 'string') {
+            return;
```
mike182uk, line 325: This will result in no role being set, we probably want it to be `Anonymous` like if an error was thrown right?

features/step_definitions/stepdefs.js
```diff
@@ -242,15 +253,36 @@ BeforeAll(async function () {
 });
 
 BeforeAll(async function () {
-    captain = new WireMock('http://wiremock:8080');
+    externalActivityPubCaptain = new WireMock('http://fake-external-activitypub');
+    ghostActivityPubCaptain = new WireMock('http://fake-ghost-activitypub');
```
mike182uk, line 257: Not a blocker - How would you feel about dropping the `Captain` suffix

```
ghostActivityPub.register
```

```
externalActivityPub.register
```

Feels nice

src/app.ts
allouis: Nope - we're allowing the protocol to be `http` but only if we're not in production - otherwise we force to `https`

src/app.ts
allouis: Yessir - thanks

src/app.ts
```diff
@@ -277,6 +282,61 @@ app.use(async (ctx, next) => {
...
+
+    let protocol = 'https';
+    // We allow insecure requests when not in production for things like testing
+    if (process.env.NODE_ENV !== 'production' && !request.raw.url.startsWith('https')) {
+        protocol = 'http';
+    }
+
+    const jwksURL = new URL('/ghost/.well-known/jwks.json', `${protocol}://${host}`);
+
+    const jwksResponse = await fetch(jwksURL, {
```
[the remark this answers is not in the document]
allouis: I'm going to add a separate issue for this for now

PR #5, opened 2024-07-24
Implemented Ghost site changed webhook handler

mike182uk:
refs [MOM-300](https://linear.app/tryghost/issue/MOM-300/update-actor-data-with-ghost-site-settings#comment-4e49803f)

Implemented Ghost site changed webhook handler so that when the settings for a Ghost instance are updated, the actor data is also updated in the database (name, description, icon). For this to work, the Ghost instance needs to setup a webhook for the `Site Changed` event and point it to the `<host>/.ghost/activitypub/webhooks/site/changed` endpoint

### DIFF
--- a/src/handlers.ts
+++ b/src/handlers.ts
@@ -7,6 +7,7 @@ import {
     Create,
     Note,
     Update,
+    PUBLIC_COLLECTION
 } from '@fedify/fedify';
 import { Context, Next } from 'hono';
 import ky from 'ky';
@@ -15,6 +16,12 @@ import { addToList } from './kv-helpers';
 import { toURL } from './toURL';
 import { ContextData, HonoContextVariables, fedify } from './app';
 import type { PersonData } from './user';
+import {
+    ACTOR_DEFAULT_HANDLE,
+    ACTOR_DEFAULT_ICON,
+    ACTOR_DEFAULT_NAME,
+    ACTOR_DEFAULT_SUMMARY
+} from './constants';
 
 type GhostSiteSettings = {
     site: {
@@ -22,6 +29,20 @@ type GhostSiteSettings = {
         icon: string;
         title: string;
     }
+}
+
+async function getGhostSiteSettings(host: string): Promise<GhostSiteSettings> {
+    const settings = await ky
+        .get(`https://${host}/ghost/api/admin/site/`)
+        .json<Partial<GhostSiteSettings>>();
+
+    return {
+        site: {
+            description: settings?.site?.description || ACTOR_DEFAULT_SUMMARY,
+            title: settings?.site?.title || ACTOR_DEFAULT_NAME,
+            icon: settings?.site?.icon || ACTOR_DEFAULT_ICON
+        }
+    };
 }
 
 async function postToArticle(ctx: RequestContext<ContextData>, post: any) {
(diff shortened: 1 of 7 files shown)


src/handlers.ts
```diff
@@ -121,6 +131,39 @@ export async function postPublishedWebhook(
...
+export async function siteChangedWebhook(
+    ctx: Context<{ Variables: HonoContextVariables }>,
+    next: Next,
+) {
+    // Retrieve site settings from Ghost
+    const host = ctx.req.header('host')
+
+    const settings = await ky
+        .get(`https://${host}/ghost/api/admin/site/`)
+        .json<GhostSiteSettings>();
```
mike182uk, line 143: This is sus, but not sure what the best thing to do here is as we don't know anything about the requesting client (so am just assuming that the request came from the Ghost instance on that is accessible on the same host 🙃)

src/handlers.ts
```diff
@@ -121,6 +131,43 @@ export async function postPublishedWebhook(
...
+    ctx: Context<{ Variables: HonoContextVariables }>,
+    next: Next,
+) {
+    try {
+        // Retrieve site settings from Ghost
+        const host = ctx.req.header('host');
+
+        const settings = await ky
+            .get(`https://${host}/ghost/api/admin/site/`)
+            .json<GhostSiteSettings>();
```
mike182uk, line 144: This is sus, but not sure what the best thing to do here is as we don't know anything about the requesting client (so am just assuming that the request came from the Ghost instance on that is accessible on the same host 🙃)

allouis: Yeah, agree this is sus! Maybe something like:
```ts
async function getGhostSiteSettings(host: string): Promise<GhostSiteSettings> {
  const url = new URL(`https://${host}/ghost/api/admin/site/`);
  const settings = await ky
            .get(`https://${host}/ghost/api/admin/site/`)
            .json<unknown>();
  
  // Some assertions here so TS doesn't complain about `unknown`
  
  return {
    site: {
      description: settings?.site?.description || 'This is a summary',
      title: settings?.site?.title || 'index',
      icon: settings?.site?.icon || 'https://ghost.org/favicon.ico',
    }
  };
}
```

allouis: The above couples site settings -> default actor data, which isn't great, but it's a step forward. Would be better to have `null` as the default instead and then default in the handler probs

mike182uk: For me the sus part was using the `host` header and assuming thats where the original request was coming from in order to make the request to get the site settings 😄 

But good to know you think the rest is sus 😂

Is there a need for the defaults? The retrieved settings are merged with the existing data - If any of the expected settings are not present or `null`, the existing data would be used?


🤔 I liked the typing of the response, but understand that that might not actually be what we get, was optimising for the happy path 😄 We could add a validation step? (i.e `isSiteSettings()`) or do you think its not worth the hassle over optional chaining?

allouis: Yeah, I think typing the response is nice - but we don't actually know what we get, so I was thinking that the lil wrapper gives us the type AND verifies it

RE: The `host` header - that's all good IMO, that's how we're differentiating between different sites

Defaults - hmm, I think we should have some defaults - because we don't want empty profiles, I don't think the object spread operator will ignore `null` values?

mike182uk: Ah yeh ofc the `null` values wont get dropped 🤦‍♂️ Have gone with this approach and added in the defaults, nice one

PR #6, opened 2024-07-25
Use constants for common strings

mike182uk:
In https://github.com/TryGhost/ActivityPub/pull/5 we introduced constants to keep track of common strings. This PR updates the codebase to make use of these constants


src/user.ts
```diff
@@ -61,18 +66,6 @@ export async function getUserData(ctx: RequestContext<ContextData>, handle: stri
         ),
     };
 
-    const dataToStore: PersonData = {
```
mike182uk, line 64: Removed this as it wasn't being used

allouis: :D This is actually a bug - we should be passing `dataToStore` on line 76 🙈

allouis (commented):
Looks good - but we wanna keep `dataToStore` (and actually use it!)

src/user.ts
mike182uk: Updated in https://github.com/TryGhost/ActivityPub/pull/6/commits/a32d9356488ea158df674275a553982ab60ea2ff

allouis (approved):

PR #8, opened 2024-07-29
Fixed tests not running as expected for `yarn test` command

mike182uk:
When executing tests via `yarn test` they are executed inside of a container. The tests that are being executed are the ones added when the container was first built. Subsequent changes to the tests are not causing docker-compose to rebuild the container meaning the same set of tests were always being executed. This change mounts the `src` directory into the container so that the tests are always up to date.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -39,12 +39,12 @@ jobs:
         run: |
           echo "
           services:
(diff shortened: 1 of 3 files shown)


allouis:
https://github.com/TryGhost/ActivityPub/pull/11 Superceded by this

PR #11, opened 2024-07-30
Used exec to run tests in existing container

allouis:
The idea here is to simplify and streamline the dev stack. So rather than a separate docker-compose file for testing, and spinning up new containers we can run the tests in the existing activitypub container we use, which also has the local src files mounted.

`yarn dev` will start the stack as usual, but now it does it as a daemon, which means it runs in the background. We also pass the --no-recreate flag which ensures that running it multiple times doesn't destroy the container and is essentially idempotent.

`yarn logs` will tail the activitypub logs for the container

`yarn test` uses the `exec` docker compose command to run the tests, and also calls `yarn dev` first - which makes sure the stack is running if it wasn't already.


package.json
```diff
@@ -7,13 +7,15 @@
   "main": "src/app.ts",
   "type": "module",
   "scripts": {
-    "dev": "docker compose up",
+    "dev": "docker compose up -d --no-recreate",
+    "dev:record": "docker compose --env-file record.env up",
```
mike182uk, line 11: Is this an unrelated change?

package.json
```diff
@@ -7,13 +7,15 @@
   "main": "src/app.ts",
   "type": "module",
   "scripts": {
-    "dev": "docker compose up",
+    "dev": "docker compose up -d --no-recreate",
```
mike182uk, line 10: Should we add a script to tear down the env also? (can always do `docker compose down` i guess?)

mike182uk (approved):
Tested locally, works as expected 👍

package.json
allouis: Yessir - good 👁️

package.json
allouis: Yeah good thinking - I'll add a `yarn stop` and `yarn fix`

PR #13, opened 2024-07-30
Added announce handling

mike182uk:
Added announce handling

refs [MOM-280](https://linear.app/tryghost/issue/MOM-280/handle-incoming-announces)

### DIFF
--- a/src/app.ts
+++ b/src/app.ts
@@ -38,6 +38,8 @@ import {
     followersCounter,
     followingDispatcher,
     followingCounter,
+    inboxDispatcher,
+    inboxCounter,
     outboxDispatcher,
     outboxCounter,
     articleDispatcher,
@@ -129,6 +131,13 @@ fedify
         followingDispatcher,
     )
     .setCounter(followingCounter);
+
+fedify
+    .setInboxDispatcher(
+        '/.ghost/activitypub/inbox/{handle}',
+        inboxDispatcher,
+    )
+    .setCounter(inboxCounter);
 
 fedify
     .setOutboxDispatcher(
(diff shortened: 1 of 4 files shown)


src/app.ts
```diff
@@ -105,6 +111,13 @@ fedify
     )
     .setCounter(followingCounter);
 
+fedify
```
allouis, line 114: If we're using this - then I think we need to remove the custom inbox handler on line 220, as well as the backing code behind it

src/dispatchers.ts
```diff
@@ -205,6 +272,49 @@ export async function followingCounter(
...
+    console.log('Inbox Dispatcher');
+    const results = (await ctx.data.db.get<string[]>(['inbox'])) || [];
+    console.log(results);
+    let items: Activity[] = [];
+    for (const result of results) {
+        try {
+            const thing = await ctx.data.globaldb.get<StoredThing>([result]);
+
+            // If the object is a string, it's a URI, so we should to look it up
+            // in the globalDb. If it's not in the globalDb, we should just
```
allouis, line 292: Long term we probably want to fetch it from the network and store it - but this is fine for now

allouis (approved):

src/app.ts
mike182uk: 👍 Didn't realise we already had an inbox handler 🤦‍♂️ I'll move the logic to the existing handler and create an issue to refactor the existing handler to utilise `setInboxDispatcher` in the future

src/dispatchers.ts
mike182uk: Not sure I'm following - In the case of `Announce` we retrieve the announced object and store it at the point of handling the `Announce` activity. So at this point, we already have the object in the db?

(obviously this is only valid for `Announce` activities that have been handled by the handler)

PR #21, opened 2024-08-06
Added expanded followers data to the db

allouis:
We're running into some performance issues when fetching large followers lists. Ultimately we need to move away from using MySQL as a key value store - but for now we can store the expanded list which should help with performance.

### DIFF
--- a/src/dispatchers.ts
+++ b/src/dispatchers.ts
@@ -11,6 +11,10 @@ import {
     Update,
     Context,
     Announce,
+    isActor,
+    Service,
+    Actor,
+    Object as APObject,
 } from '@fedify/fedify';
 import { v4 as uuidv4 } from 'uuid';
 import { addToList } from './kv-helpers';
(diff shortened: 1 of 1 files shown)


src/dispatchers.ts
```diff
@@ -235,18 +250,25 @@ export async function followersDispatcher(
...
+    const fullResults = await ctx.data.db.get<any[]>(['followers', 'expanded']);
+    if (fullResults) {
+        items = (await Promise.all(
+            fullResults.map((result): Promise<Actor | null> => {
+                if (result.type === 'Person') {
+                    return Person.fromJsonLd(result)
+                }
+                if (result.type === 'Service') {
+                    return Service.fromJsonLd(result);
+                }
```
dahlia, line 263: To make it more robust:

```suggestion
                return Object.fromJsonLd(result).then(object => isActor(object) ? object : null);
```

(You need to import `Object` from `@fedify/fedify` in the above.)

allouis: Thank you :)

src/dispatchers.ts
```diff
@@ -235,18 +252,31 @@ export async function followersDispatcher(
...
-        }
+    let items: Recipient[] = [];
+    const fullResults = await ctx.data.db.get<any[]>(['followers', 'expanded']);
+    if (fullResults) {
+        items = (await Promise.all(
+            fullResults.map((result, i): Promise<Recipient> => {
+                return {
+                    ...result,
+                    id: new URL(result.id),
+                    inboxId: new URL(result.inbox)
```
dahlia, line 263: For efficient delivery of activities:

```suggestion
                    inboxId: new URL(result.inbox),
                    endpoints: result.endpoints?.sharedInbox != null
                        ? { sharedInbox: result.endpoints.sharedInbox }
                        : null,
```

PR #27, opened 2024-08-14
Increased connection pool size

mike182uk:
Increased connection pool size to accommodate traffic spikes when an update is published.


mike182uk:
Closing this for now as we have scaled our prod instances

PR #33, opened 2024-08-27
Added handler for incoming likes

mike182uk:
refs [AP-283](https://linear.app/tryghost/issue/AP-283/handle-incoming-likes)

Added handler for incoming likes that adds the like activity to the actors inbox and persists any relevant data to the database.

### DIFF
--- a/src/dispatchers.ts
+++ b/src/dispatchers.ts
@@ -15,6 +15,7 @@ import {
     Actor,
     Object as APObject,
     Recipient,
+    Like,
 } from '@fedify/fedify';
 import { v4 as uuidv4 } from 'uuid';
 import { addToList } from './kv-helpers';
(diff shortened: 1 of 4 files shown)


src/dispatchers.ts
```diff
@@ -229,6 +230,69 @@ export async function handleAnnounce(
...
+    }
+
+    // Lookup liked object - If not found in globalDb, perform network lookup
+    let object = null;
+    let existing = await ctx.data.globaldb.get([like.objectId.href]) ?? null;
+
+    if (!existing) {
+        console.log('Object not found in globalDb, performing network lookup');
+
+        object = await lookupObject(like.objectId);
```
dahlia, line 265: This can be simplified:

```suggestion
        object = await like.getObject();
```

mike182uk: Thanks @dahlia thats updated in https://github.com/TryGhost/ActivityPub/pull/33/commits/81816ea00bb5251a11eebd22218e955839443524

PR #35, opened 2024-09-05
[WIP] Sending & Un-sending Likes

allouis:
🍦 

We got a few things happening here!

We're adding a liked collection to actors, this is so that we can render a list of the liked content in the Admin. The liked collection functions similar to how we're doing the inbox and outbox, we're storing an array of the ids belonging to the collection.

We've got some basic object dispatchers wired up for `Like` and `Undo` which is necessary so that we can generate id's for those objects.

The inbox handler has been updating to attached a `liked` property to the `object` of the activity, this allows us to render the item correctly in the frontend with a filled in heart. I didn't store the `liked` property on the object, because of concerns of losing the data - e.g. if we start handling `Update` activities and we overwrite the object in the DB - we might lose the `liked` status. There's also the issue of what happens when we support multiple users, we can't store it on the object itself.

Adding the `liked` property is not efficient though - and we really need to rethink storage solutions around this.

And of course we have two actions "like" and "unlike" - they are fairly simple - we first check if the object we're trying to like/unlike exists, and 404 if it doesn't. Then we check if we've already performed that action, and 409 if we have. After that we construct the relevant activity, store it, and send it to our followers - we also need to send it to the original author of the object using `attributedTo`

### DIFF
--- a/src/handlers.ts
+++ b/src/handlers.ts
@@ -1,6 +1,8 @@
 import {
     Article,
     Follow,
+    Like,
+    Undo,
     RequestContext,
     lookupObject,
     isActor,
@@ -12,13 +14,14 @@ import {
 import { Context, Next } from 'hono';
 import sanitizeHtml from 'sanitize-html';
 import { v4 as uuidv4 } from 'uuid';
-import { addToList } from './kv-helpers';
+import { addToList, removeFromList } from './kv-helpers';
 import { toURL } from './toURL';
 import { ContextData, HonoContextVariables, fedify } from './app';
 import { getSiteSettings } from './ghost';
 import type { PersonData } from './user';
 import { ACTOR_DEFAULT_HANDLE } from './constants';
 import { Temporal } from '@js-temporal/polyfill';
+import { createHash } from 'node:crypto';
 
 type StoredThing = {
     object: string | {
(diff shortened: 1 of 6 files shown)


src/handlers.ts
```diff
@@ -67,6 +76,134 @@ async function postToArticle(ctx: RequestContext<ContextData>, post: Post) {
...
+        id: likeId,
+        actor: actor,
+        object: objectToLike,
+        to: PUBLIC_COLLECTION,
+        cc: apCtx.getFollowersUri(ACTOR_DEFAULT_HANDLE),
+    });
+    const likeJson = await like.toJsonLd();
+    await ctx.get('globaldb').set([like.id!.href], likeJson);
+
+    await addToList(ctx.get('db'), ['liked'], like.id!.href);
```
mike182uk, line 184: 🤔 Should this be the ID of the liked object rather than the like activity itself? Then the liked dispatcher returns the objects liked and not the like activities themselves?

Spec says:

> This is a list of every object from all of the actor's Like activities, added as a [side effect]

allouis: I did initially think that, but then in the liked dispatcher, fedify would error because it expects a return type of `Like[]` 😕 

I think we need to get fedify updated (and fixed) to handle returning objects from the liked dispatcher - I'll sort that next week

PR #37, opened 2024-09-10
Fixed followers dispatcher and api endpoint

allouis:
There was a bug in fedify with serialising Actors which mean we had to return incompleter followers data, which broke our frontend, this has now been fixed so we can update the followers dispatcher to return full Actor objects again


mike182uk:
I think we can close this @allouis These changes were made in https://github.com/TryGhost/ActivityPub/pull/75

PR #40, opened 2024-09-17
Added dedicated endpoint for retrieving activities for a given actor

mike182uk:
refs [AP-377](https://linear.app/tryghost/issue/AP-377/inbox-returning-33mb-of-data)

Added a new endpoint for retrieving activities for a given actor:

```
GET /.ghost/activitypub/activities/:handle?limit=<number>&cursor=<string>
```

At the moment this returns the same data as the inbox handler but in the future we are likely to want to augment this with additional data (i.e including the actor's own activities). This endpoint is designed to be used by the activity pub UI in Ghost Admin and not other activity pub compliant servers / clients.

The returned data is paginated using cursor based pagination. The `cursor` query parameter is optional and is used to fetch the next page of activities. The `limit` query parameter is also optional and is used to limit the number of activities returned in a single request.

### DIFF
--- a/src/handlers.ts
+++ b/src/handlers.ts
@@ -1,6 +1,8 @@
 import {
     Article,
+    Context as APContext,
     Follow,
+    KvStore,
     Like,
     Undo,
     RequestContext,
@@ -9,7 +11,7 @@ import {
     Note,
     Update,
     Actor,
-    PUBLIC_COLLECTION
+    PUBLIC_COLLECTION,
 } from '@fedify/fedify';
 import { Context, Next } from 'hono';
 import sanitizeHtml from 'sanitize-html';
@@ -24,7 +26,7 @@ import { Temporal } from '@js-temporal/polyfill';
 import { createHash } from 'node:crypto';
 import { lookupActor } from 'lookup-helpers';
 
-type StoredThing = {
+type InboxItem = {
     id: string;
     object: string | {
         id: string;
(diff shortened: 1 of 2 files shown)


src/handlers.ts
```diff
@@ -375,71 +377,102 @@ export async function siteChangedWebhook(
     });
 }
 
+async function buildInboxItem(
```
mike182uk, line 380: This was the guts of the `inboxHandler`. For now we want to use the same data structures in the newly added activities endpoint so have extracted this logic here and tidied up a little 🧹

src/handlers.ts
```diff
@@ -375,71 +377,102 @@ export async function siteChangedWebhook(
...
+                item.object.liked = true;
+            }
+        }
+    }
+
+    // Return the built item
+    return item;
+}
+
 export async function inboxHandler(
```
mike182uk, line 437: The diff for this is a little hard to read - The changes make this essentially the same as the implementation for `getActivities` just without any pagination

src/handlers.ts
```diff
@@ -450,3 +483,62 @@ export async function inboxHandler(
...
+    // This is used to add a "liked" property to the item if the user has liked it
+    const liked = (await db.get<string[]>(['liked'])) || [];
+
+    // Fetch the inbox from the database:
+    //   - Data is structured as an array of strings
+    //   - Each string is a URI to an object in the database
+    const inbox = (await db.get<string[]>(['inbox'])) || [];
+
+    // Slice the results array based on the cursor and limit
+    const paginatedInbox = inbox.slice(cursor, cursor + limit);
```
allouis, line 512: I think we want to slice from the end of the array, so that we get the most recent items first

allouis (approved):
Looks good! 

The pagination seems to be "offset" rather than "cursor" pagination tho - which is fine for now, but it would be nice to switch to cursor eventually - with the offset (and the switch to reading most recent items) it will be possible for us to send duplicate activities - but can be considered an edge case for now

src/handlers.ts
mike182uk: Yeh, previously we was sorting in the client. I've updated this now so the newest items are returned first and removed the manual sorting in the client 👍

mike182uk:
> The pagination seems to be "offset" rather than "cursor" pagination tho - which is fine for now, but it would be nice to switch to cursor eventually - with the offset (and the switch to reading most recent items) it will be possible for us to send duplicate activities - but can be considered an edge case for now

Nice one, I've updated the cursor implementation to use the ID of the item instead of the index of the item

PR #41, opened 2024-09-18
Dev environment

allouis:
Added a script to allow populating the local db with 5k followers

This also sets the groundwork for adding more scripts to run against our dev environment

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -104,3 +104,9 @@ services:
       - ./wiremock/mappings:/home/wiremock/mappings
       - ./wiremock/__files:/home/wiremock/__files
     entrypoint: ["/docker-entrypoint.sh", "--global-response-templating", "--disable-gzip", "--verbose"]
+
+  fake-mastodon:
+    image: wiremock/wiremock:latest
+    ports:
+      - "8084:8080"
+    entrypoint: ["/docker-entrypoint.sh", "--global-response-templating", "--disable-gzip", "--verbose"]
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -18,6 +18,12 @@ services:
       mysql:
         condition: service_healthy
 
+  scripts:
+    build: scripts
+    depends_on:
+      - activitypub
+      - fake-mastodon
+
   nginx:
     build: nginx
     ports:
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -10,6 +10,9 @@ services:
       - MYSQL_HOST=mysql
       - MYSQL_PORT=3306
       - MYSQL_DATABASE=activitypub
+      - NODE_ENV=testing
+      - ALLOW_PRIVATE_ADDRESS=true
+      - SKIP_SIGNATURE_VERIFICATION=true
     command: node --import tsx --watch src/app.ts
     depends_on:
       mysql:
--- a/nginx/server.conf
+++ b/nginx/server.conf
@@ -1,28 +1,36 @@
-server {
-    location /.ghost/activitypub {
-        proxy_set_header Host $host;
-        proxy_set_header X-Real-IP $remote_addr;
-        proxy_set_header X-Forwarded-For $proxy_add_x_forwarded_for;
-        proxy_set_header X-Forwarded-Proto https;
-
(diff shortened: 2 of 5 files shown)


nginx/server.conf
```diff
@@ -1,27 +1,47 @@
 server {
     location /.ghost/activitypub {
+        if ($http_x_forwarded_proto = "https") {
+            set $real_scheme https;
+        }
+        if ($http_x_forwarded_proto = "") {
+            set $real_scheme http;
+        }
```
mike182uk, line 8: Not a blocker - I think we could use a [map directive](https://nginx.org/en/docs/http/ngx_http_map_module.html#map) and dry this up:

```
map $http_x_forwarded_proto $real_scheme {
    default http;
    https https;
}
 ```
 
 I think it might have to go at the `http` level though 🤔
 
 ```
 http {
    map $http_x_forwarded_proto $real_scheme {
        default http;
        https https;
    }

    server {
        location /.ghost/activitypub {
            proxy_set_header Host $host;
            proxy_set_header X-Real-IP $remote_addr;
            proxy_set_header X-Forwarded-For $proxy_add_x_forwarded_for;
            proxy_set_header X-Forwarded-Proto $real_scheme;

            proxy_pass http://activitypub:8080;
        }

        
        ...
    }
}
 ```

File: docker-compose.yml [hunk 2]
```diff
@@ -10,11 +10,20 @@ services:
       - MYSQL_HOST=mysql
       - MYSQL_PORT=3306
       - MYSQL_DATABASE=activitypub
+      - NODE_ENV=testing
+      - ALLOW_PRIVATE_ADDRESS=true
+      - SKIP_SIGNATURE_VERIFICATION=true
```
mike182uk, line 15: Should we make a note here (comment) or / and track (in linear task) that we need to come back to this once the signature stuff has been sorted (or maybe make a note on the task for that if we have one?)

scripts/bin/populate-activitypub-db
```diff
@@ -0,0 +1,177 @@
...
+# Construct the Authorization header
+AUTH_HEADER="keyId=\"$KEY_ID\",algorithm=\"rsa-sha256\",headers=\"(request-target) host date digest\",signature=\"$SIGNATURE\""
+
+# Make the request
+curl -v -X POST http://activitypub:8080/.ghost/activitypub/inbox/index \
+  -H "Host: $HOST" \
+  -H "Date: $DATE" \
+  -H "Digest: SHA-256=$DIGEST" \
+  -H "Signature: $AUTH_HEADER" \
+  -d "$BODY"
```
mike182uk, line 177: This a left over? (occurs after the script has exited)

scripts/bin/populate-activitypub-db
```diff
@@ -0,0 +1,177 @@
+#!/usr/bin/env bash
```
mike182uk, line 1: Not a blocker - Kudos on persevering with raw bash here, the script is easy / clear to follow and grok 👌

Thoughts on using https://github.com/google/zx for stuff like this (not suggesting you change anything here 👍)?

scripts/bin/populate-activitypub-db
```diff
@@ -0,0 +1,177 @@
...
+            "Content-Type": "application/activity+json"
+        }
+    }
+}'
+
+# Define variables
+PRIVATE_KEY="private.pem"
+
+# Loop 5000 times
+for i in $(seq 1 5000); do
```
mike182uk, line 97: Should we make it so the `5000` can be passed into the script?

mike182uk (approved):

scripts/bin/populate-activitypub-db
allouis: Oooh I remember seeing this before, still not tried it! I'm happy to give it a go!

scripts/bin/populate-activitypub-db
allouis: Yessss

nginx/server.conf
allouis: Yeah this is nice - I'll try it and update

PR #47, opened 2024-09-24
Added endpoint for searching for Fediverse users

mike182uk:
refs [AP-352](https://linear.app/tryghost/issue/AP-352/search-for-mastodon-usernames-in-ghost-admin)

Added a new endpoint to enable searching for Fediverse users by their handle or URL

### DIFF
--- a/src/api/action/search.ts
+++ b/src/api/action/search.ts
@@ -1,6 +1,5 @@
 import { isActor } from '@fedify/fedify';
 import { Context } from 'hono';
-import ky from 'ky';
 import sanitizeHtml from 'sanitize-html';
 
 import {
@@ -35,6 +34,7 @@ export async function searchAction(
     });
 
     // Parse "query" from query parameters
+    // ?query=<string>
     const query = ctx.req.query('query') || '';
 
     // Init search results - At the moment we only support searching for an actor (ui calls them profiles)
@@ -94,11 +94,8 @@ export async function searchAction(
             result.handle = `@${actor.preferredUsername}@${actor.id!.host}`;
 
             // Retrieve follower count for the actor
-            const followers = await ky
-                .get(actor.followersId!.href)
-                .json<{ totalItems: number }>();
-
-            result.followerCount = followers.totalItems;
+            result.followerCount = (await actor.getFollowers() || { totalItems: 0 })
+                .totalItems || 0;
 
             // Determine if the current user is following the actor
             const following = (await db.get<string[]>(['following'])) || [];
--- a/src/api/action/search.ts
+++ b/src/api/action/search.ts
@@ -1,10 +1,19 @@
+import { type Context } from 'hono';
 import { isActor } from '@fedify/fedify';
-import { Context } from 'hono';
-import sanitizeHtml from 'sanitize-html';
 
 import {
+    getAttachments,
+    getFollowerCount,
+    getHandle,
+    getRecentActivities,
+    isFollowing,
+    isHandle,
+} from '../../helpers/activitypub/actor';
+import { sanitizeHtml } from '../../helpers/sanitize';
+import { isUri } from '../../helpers/uri';
+import {
     type HonoContextVariables,
-    fedify
+    fedify,
 } from '../../app';
 
 interface ProfileSearchResult {
@@ -19,21 +28,6 @@ interface SearchResults {
     profiles: ProfileSearchResult[];
 }
 
-// @<username>@<domain>.<tld>
-const HANDLE_REGEX = /^@([\w-]+)@([\w-]+\.[\w.-]+)$/;
-
-// http(s)://...
-const URI_REGEX = /^https?:\/\/[^\s/$.?#].[^\s]*$/;
-
-// Config for sanitizing HTML
-const SANITIZE_HTML_CONFIG = {
-    allowedTags: ['a', 'p', 'img', 'br', 'strong', 'em', 'span'],
-    allowedAttributes: {
-        a: ['href'],
-        img: ['src'],
-    }
-};
-
 export async function searchAction(
     ctx: Context<{ Variables: HonoContextVariables }>,
 ) {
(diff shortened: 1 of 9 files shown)


src/api/action/search.ts
```diff
@@ -0,0 +1,122 @@
...
+            headers: {
+                'Content-Type': 'application/json',
+            },
+            status: 200,
+        });
+    }
+
+    // Lookup actor by handle or url
+    try {
+        const actor = await apCtx.lookupObject(query);
```
mike182uk, line 59: @allouis I was noticing that if `lookupObject` errors out (i.e a malformed url is passed), it doesn't get caught in this try / catch, which I would of expected?

src/app.ts
```diff
@@ -56,6 +56,8 @@ import {
     handleLike
 } from './dispatchers';
 
+import { searchAction } from './api/action/search';
```
mike182uk, line 59: @allouis I took a slightly different approach on this action and placed it here. It feels like the handlers file is starting to get a bit unwieldily. Thoughts?

src/api/action/search.ts
allouis: I guess it returns `null` instead of throwing? Or is there something else here?

allouis (approved):
We might wanna store the actor in our global DB eventually - not sure! Also would be good to eventually search our DB before the network - but I think for now this is fine!

src/api/action/search.ts
dahlia: Yeah, it returns `null` instead of throwing!

mike182uk: I was actually wanting to stop verbose output in the logs 😅 So at the moment if I passed this `@foo@bar.baz` I get 

```
at node:internal/deps/undici/undici:13178:13
     at process.processTicksAndRejections (node:internal/process/task_queues:95:5)
     at async lookupWebFinger (file:///opt/activitypub/node_modules/@fedify/fedify/esm/webfinger/lookup.js:32:24)
     at async lookupObject (file:///opt/activitypub/node_modules/@fedify/fedify/esm/vocab/lookup.js:58:21)
     at async searchAction (/opt/activitypub/src/api/action/search.ts:59:23)
     at async dispatch (file:///opt/activitypub/node_modules/hono/dist/compose.js:29:17)
     at async <anonymous> (/opt/activitypub/src/app.ts:279:5)
     at async dispatch (file:///opt/activitypub/node_modules/hono/dist/compose.js:29:17)
     at async <anonymous> (/opt/activitypub/src/app.ts:259:5)
     at async dispatch (file:///opt/activitypub/node_modules/hono/dist/compose.js:29:17) {
   [stack]: 'TypeError: fetch failed\n' +
     '    at node:internal/deps/undici/undici:13178:13\n' +
     '    at process.processTicksAndRejections (node:internal/process/task_queues:95:5)\n' +
     '    at async lookupWebFinger (file:///opt/activitypub/node_modules/@fedify/fedify/esm/webfinger/lookup.js:32:24)\n' +
     '    at async lookupObject (file:///opt/activitypub/node_modules/@fedify/fedify/esm/vocab/lookup.js:58:21)\n' +
     '    at async searchAction (/opt/activitypub/src/api/action/search.ts:59:23)\n' +
     '    at async dispatch (file:///opt/activitypub/node_modules/hono/dist/compose.js:29:17)\n' +
     '    at async <anonymous> (/opt/activitypub/src/app.ts:279:5)\n' +
     '    at async dispatch (file:///opt/activitypub/node_modules/hono/dist/compose.js:29:17)\n' +
     '    at async <anonymous> (/opt/activitypub/src/app.ts:259:5)\n' +
     '    at async dispatch (file:///opt/activitypub/node_modules/hono/dist/compose.js:29:17)',
   [message]: 'fetch failed',
   [cause]: Error: getaddrinfo ENOTFOUND bar.baz
       at GetAddrInfoReqWrap.onlookupall [as oncomplete] (node:dns:120:26) {
     [stack]: 'Error: getaddrinfo ENOTFOUND bar.baz\n' +
       '    at GetAddrInfoReqWrap.onlookupall [as oncomplete] (node:dns:120:26)',
     [message]: 'getaddrinfo ENOTFOUND bar.baz',
     errno: -3008,
     code: 'ENOTFOUND',
     syscall: 'getaddrinfo',
     hostname: 'bar.baz'
   }
 }
 ```

Comment (mike182uk) [reply src/api/action/search.ts:59]: Ah, its because Fedify logs it as part of `DBG` - ` 06:47:26.790 DBG fedify·webfinger·lookup Failed to fetch WebFinger resource descriptor: TypeError: fetch failed`

Ignore me 😄

**Comment by mike182uk** [comment]
@allouis Yeh i wasn't sure on whether to store this as it might not be data we want to keep around forever (i.e only 1 user ever searches for this specific user). Also we would have no idea if the data was up to date or not 🤔 Could be a good candidate for whatever caching mechanism we end up implementing?

File: src/api/action/search.ts [hunk 3]
```diff
@@ -0,0 +1,122 @@
...
+
+            // Compute the full handle for the actor
+            result.handle = `@${actor.preferredUsername}@${actor.id!.host}`;
+
+            // Retrieve follower count for the actor
+            const followers = await ky
+                .get(actor.followersId!.href)
+                .json<{ totalItems: number }>();
+
+            result.followerCount = followers.totalItems;
```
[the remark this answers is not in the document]
dahlia: If you want to get a collection of followers, you can easily do so with `await actor.getFollowers()`:

```suggestion
            result.followerCount = (await actor.getFollowers()).totalItems;
```

I guess you're maintaining the cache for ActivityStreams objects, it would be more efficient if you implement a `DocumentLoader` so that it looks for the cache first before actually fetching an ActivityStreams object through the network. 🤔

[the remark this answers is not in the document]
mike182uk: Aha, yes this is what I was looking for, thanks!

> I guess you're maintaining the cache for ActivityStreams objects, it would be more efficient if you implement a DocumentLoader so that it looks for the cache first before actually fetching an ActivityStreams object through the network. 🤔

Yep, we already do this elsewhere / have this functionality 👍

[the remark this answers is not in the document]
mike182uk: Was thinking about this a little more @dahlia 

Do you think it would be useful to have this functionality within Fedify itself, so you could just do:

```js
await actor.getFollowersCount()
```

PR #48, opened 2024-09-25
Ensure object replies is always an array

mike182uk:
Ensure object `replies` is always an array when querying for activities so that the client does not have to check for the type of `replies` before iterating over them. If we didn't do this and an object had a `replies` property, but a `repliesMap` was not present, the `replies` property could be an activitypub collection, which the client is not expecting

This also simplifies the handling of the `inbox` seeming though we do not need to do as much processing on the results now that the client is utilising the `activities` endpoint and not the `inbox` endpoint directly


mike182uk:
Closing this for now as we are going to be changing how replies work under the hood

PR #51, opened 2024-09-29
Miscellaneous improvements

allouis:
This is some cleanup that happened during the auth work

### DIFF
--- a/src/app.ts
+++ b/src/app.ts
@@ -74,7 +74,7 @@ if (process.env.SENTRY_DSN) {
 await configure({
     sinks: { console: getConsoleSink() },
     filters: {},
-    loggers: [{ category: 'fedify', sinks: ['console'], level: 'debug' }],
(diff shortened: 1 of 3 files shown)


src/app.ts
```diff
@@ -74,7 +74,7 @@ if (process.env.SENTRY_DSN) {
 await configure({
     sinks: { console: getConsoleSink() },
     filters: {},
-    loggers: [{ category: 'fedify', sinks: ['console'], level: 'debug' }],
+    loggers: [{ category: 'fedify', sinks: ['console'], level: null }],
```
mike182uk, line 77:
```suggestion
    loggers: [{ category: 'fedify', sinks: ['console'], level: 'error' }],
```

maybe?

mike182uk (approved):

src/app.ts
dahlia: `'warning'` might be better than `'error'` as Fedify logs deprecation messages as warnings.

PR #62, opened 2024-10-04
Added e2e tests for API endpoints

mike182uk:
no refs

Added e2e tests for API endpoints


src/app.ts
```diff
@@ -83,7 +83,11 @@ if (process.env.SENTRY_DSN) {
 await configure({
     sinks: { console: getConsoleSink() },
     filters: {},
-    loggers: [{ category: 'fedify', sinks: ['console'], level: 'warning' }],
+    loggers: [{
+        category: 'fedify',
+        sinks: ['console'],
+        level: process.env.NODE_ENV === 'testing' ? 'debug' : 'warning'
```
mike182uk, line 89: `debug` was useful to figure out why my test wasn't working 😅

docker-compose.yml
```diff
@@ -112,28 +113,29 @@ services:
...
-
   mysql-testing:
     image: mysql:lts
     environment:
       - MYSQL_ROOT_PASSWORD=root
       - MYSQL_USER=ghost
       - MYSQL_PASSWORD=password
       - MYSQL_DATABASE=activitypub
+    ports:
+      - "3308:3306"
```
mike182uk, line 124: Useful for debugging test db

docker-compose.yml
```diff
@@ -81,6 +81,7 @@ services:
       - NODE_ENV=testing
       - SKIP_SIGNATURE_VERIFICATION=true
       - ALLOW_PRIVATE_ADDRESS=true
+      - NODE_TLS_REJECT_UNAUTHORIZED=0
```
mike182uk, line 84: @allouis I needed to add this otherwise the webfinger lookup on the `fake-external-activitypub` would not work.

When Fedify does the lookup based on just being provided a handle (`@mike@mikes.world`) it defaults to using `https` so i had to expose the `--https-port` on the wiremock containers. This uses a self signed certificate, so had to enable `NODE_TLS_REJECT_UNAUTHORIZED` so that node didn't complain about the certificate.

 Not sure if i've not set something up correct somewher thats meaning i need to do this?

src/app.ts
```diff
@@ -103,7 +103,7 @@ await configure({
     filters: {},
     loggers: [
         { category: 'activitypub', sinks: ['console'], level: 'info' },
-        { category: 'fedify', sinks: ['console'], level: 'warning' }
+        { category: 'fedify', sinks: ['console'], level: process.env.NODE_ENV === 'testing' ? 'debug' : 'warning' }
```
mike182uk, line 106: debug was useful to figure out why my test wasn't working 😅

mike182uk:
This is way out of date, closing for now

PR #63, opened 2024-10-08
Add filter for filtering out activities that are replies to other activities

mike182uk:
refs [TryGhost/Ghost#21254](https://github.com/TryGhost/Ghost/pull/21254)

### DIFF
--- a/src/api/action/getActivities.ts
+++ b/src/api/action/getActivities.ts
@@ -141,6 +141,13 @@ export async function getActivitiesAction(
                     return siteHost === replyHost;
                 }
 
+                 // ?filter={type: ['<activityType>:<objectType>:notReply']}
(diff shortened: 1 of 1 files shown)


src/api/action/getActivities.ts
```diff
@@ -141,6 +141,13 @@ export async function getActivitiesAction(
                     return siteHost === replyHost;
                 }
 
+                 // ?filter={type: ['<activityType>:<objectType>:notReply']}
+                 if (filter.criteria && filter.criteria.startsWith('notReply')) {
+                    if (activity.reply_object_url) {
```
mike182uk, line 146: slightly confusing diff context -`reply_object_url` is not a property an activitypub object, is a property of an activity object returned from `activityMeta.get`

PR #64, opened 2024-10-22
Added functionality to filter out non-follower activity

mike182uk:
refs [AP-477](https://linear.app/ghost/issue/AP-477/incorrect-posts-in-home-feed)

Added functionality to filter out non-follower activity when the `excludeNonFollower` query param is set to true

### DIFF
--- a/src/api/action/getActivities.ts
+++ b/src/api/action/getActivities.ts
@@ -54,12 +54,18 @@ export async function getActivitiesAction(
         }
     });
 
+    // Parse "excludeNonFollowers" from query parameters
+    // This is used to exclude activities from non-followers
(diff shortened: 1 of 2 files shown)


src/api/action/getActivities.ts
```diff
@@ -109,16 +115,16 @@ export async function getActivitiesAction(
     // Filter the activity refs by any provided type filters
     if (typeFilters.length > 0) {
         activityRefs = activityRefs.filter(ref => {
-            const activity = activityMeta.get(ref)!;
+            const meta = activityMeta.get(ref)!;
```
mike182uk, line 118: Renamed this `meta` to avoid confusion with thinking the thing returned from `activityMeta` is the `activity`

PR #75, opened 2024-10-24
Added pagination to ActivityPub collections

mike182uk:
refs [AP-526](https://linear.app/ghost/issue/AP-526/implement-pagination-for-fedify-collections)

Added pagination to `followers`, `following` and `outbox` collections. Also ensured that the `followers` collection items get expanded.


dahlia (approved):

src/dispatchers.ts
```diff
@@ -300,45 +306,66 @@ export async function inboxErrorHandler(
...
+            ? (offset + FOLLOWERS_PAGE_SIZE).toString()
+            : null;
+
+        const slicedResults = results.slice(offset, offset + FOLLOWERS_PAGE_SIZE);
+
+        const actors = (
+            await Promise.all(
+                slicedResults.map((result) => lookupActor(ctx, result))
+            )
+        ).filter((item): item is Actor => isActor(item))
```
allouis, line 351: We do a filter here _after_ slicing - which means we may get a smaller page than expected, I think it's fine for now - but do you mind adding a TODO comment to the code - we really should be able to remove this check

allouis (approved):

src/dispatchers.ts
mike182uk: 🤔 I think this was to get around some typing issue, maybe there is a better approach

mike182uk: Ah, i see what you mean, this was in the original actor resolution block. Will add a note 👍

mike182uk: Note added and tracking [here](https://linear.app/ghost/issue/AP-549/following-dispatcher-may-return-less-results-than-defined-page-size)

PR #87, opened 2024-10-29
Added `nodeinfo` routing

mike182uk:
refs [AP-485](https://linear.app/ghost/issue/AP-485/add-nodeinfo-endpoint)

Added routing for the `nodeinfo` endpoint

### DIFF
--- a/src/app.ts
+++ b/src/app.ts
@@ -54,6 +54,7 @@ import {
     likeDispatcher,
     undoDispatcher,
     articleDispatcher,
+    nodeInfoDispatcher,
     noteDispatcher,
     followDispatcher,
     acceptDispatcher,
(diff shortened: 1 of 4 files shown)


src/app.ts
```diff
@@ -242,6 +243,11 @@ fedify.setObjectDispatcher(
     undoDispatcher,
 );
 
+fedify.setNodeInfoDispatcher(
+    '/.ghost/activitypub/nodeinfo/2.1',
```
allouis, line 247: I don't fully understand this - so does `fedify` serve a document on `/.well-known/nodeinfo` that links to this URL?

allouis (approved):

src/app.ts
dahlia: Yes.

mike182uk: When you register a `nodeInfoDispatcher` it enables `/.well-known/nodeinfo` routing (automagic) which when visited returns:

```
{
    "links":
    [
        {
            "rel": "http://nodeinfo.diaspora.software/ns/schema/2.1",
            "href": "https://mikebook-pro.tail5da2a.ts.net/.ghost/activitypub/nodeinfo/2.1",
            "type": "application/json; profile=\"http://nodeinfo.diaspora.software/ns/schema/2.1#\""
        }
    ]
}
```

and `href` is the link to our `nodeinfo`

PR #92, opened 2024-10-30
Restricted inbox to following

allouis:
closes https://linear.app/ghost/issue/AP-545

Previously we were adding all incoming messages in our inbox to our inbox, now
we restrict it to only people we're following. We do still store the object in
the global db - but we maybe be able to optimise this in future.

We're only restricting Create and Announce activities at the moment. We expect
to receive Likes from Actors we don't follow, as well as Follow activities too!

### DIFF
--- a/src/dispatchers.ts
+++ b/src/dispatchers.ts
@@ -27,6 +27,7 @@ import {
     LIKED_PAGE_SIZE,
     OUTBOX_PAGE_SIZE,
 } from './constants';
+import { isFollowing } from './helpers/activitypub/actor';
 import { getUserData, getUserKeypair } from './helpers/user';
 import { addToList } from './kv-helpers';
 import { lookupActor, lookupObject } from './lookup-helpers';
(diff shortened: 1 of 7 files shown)


src/dispatchers.ts
```diff
@@ -161,10 +161,12 @@ export async function handleCreate(
         return;
     }
 
-    // TODO Check Sender is in our following
     const createJson = await create.toJsonLd();
     ctx.data.globaldb.set([create.id.href], createJson);
-    await addToList(ctx.data.db, ['inbox'], create.id.href);
+    const following = await ctx.data.db.get<string[]>(['following']);
+    if (following?.includes(sender.id.href)) {
```
mike182uk, line 167: Not a blocker - We have a [helper](https://github.com/TryGhost/ActivityPub/blob/main/src/helpers/activitypub/actor.ts#L78) for this

mike182uk (approved):

PR #97, opened 2024-10-31
Looked up remote objects with authenticated fetch

allouis:
refs https://linear.app/ghost/issue/AP-533

Threads requires authenticated fetch when interacting with their API, which is something we should default to, for now we've added a helper, but this should eventually be opt-out rather than opt-in

### DIFF
--- a/src/api/action/profile/get.ts
+++ b/src/api/action/profile/get.ts
@@ -1,7 +1,6 @@
 import { isActor } from '@fedify/fedify';
 import type { Context } from 'hono';
 
-import { lookupObject } from 'lookup-helpers';
 import {
     type HonoContextVariables,
     fedify,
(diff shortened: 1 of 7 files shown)


src/api/action/profile/get.ts
```diff
@@ -1,6 +1,7 @@
 import { isActor } from '@fedify/fedify';
 import type { Context } from 'hono';
 
+import { lookupObject } from 'lookup-helpers';
```
mike182uk, line 4: Should we use relative path to resolve this? I hit a weird issue the other week where we ended up utilising a package and helper with the same name and the wrong one was being imported and it wasn't totally clear what the issue was 😅

allouis: Yeah good shout - my typescript lang server keeps using absolute tho!

PR #111, opened 2024-11-05
Added `pubsub` service to development environment

mike182uk:
refs [AP-564](https://linear.app/ghost/issue/AP-564/setup-a-local-pubsub-environment)

Added a new "pubsub" service to the development environment that emulates [Google Cloud Pub/Sub](https://cloud.google.com/pubsub/docs/overview) which is what we are using in production

This service used to facilitate queueing functionality when running the application locally

You can test everything is working correctly by:

1. Running the `pubsub` service:

```bash
docker compose up pubsub
```

2. Creating a script (i.e `test.js`) with the following code:

```js
process.env.PUBSUB_EMULATOR_HOST = '127.0.0.1:8085';

import { PubSub } from '@google-cloud/pubsub';

const pubsub = new PubSub({
    projectId: 'activitypub'
});

const [topic] = await pubsub.createTopic('t_foo');

console.log(`Topic ${topic.name} created`);

const [subscription] = await topic.createSubscription('s_foo');

subscription.on('message', message => {
    console.log('Received message:', message.data.toString());

    process.exit(0);
});

subscription.on('error', error => {
    console.error('Received error:', error);

    process.exit(1);
});

topic.publishMessage({
    data: Buffer.from('Hello world!')
});
```

3. Executing the script and checking the output:

```bash
node test.js
```

You should see the following output:

```
Topic projects/activitypub/topics/t_foo created
Received message: Hello world!
```

Reference documentation:
- https://cloud.google.com/pubsub/docs/overview
- https://cloud.google.com/sdk/docs/downloads-docker
- https://cloud.google.com/pubsub/docs/emulator
- https://cloud.google.com/sdk/gcloud/reference/beta/emulators/pubsub/start
- https://github.com/googleapis/nodejs-pubsub


docker-compose.yml
```diff
@@ -28,16 +30,6 @@ services:
       mysql:
         condition: service_healthy
 
-  migrate-testing:
```
mike182uk, line 31: Moved this down to be grouped with the other "testing" services

allouis (approved):
Nice and simple!

Do we need a separate service for our tests? Or can we reuse the same one with a different topic?

daniellockyer (approved):

mike182uk:
> Do we need a separate service for our tests? Or can we reuse the same one with a different topic?

Ah i hadn't thought about. I've added a separate service for testing to keep consistent with how we initialise test versions of the other services 👍

PR #143, opened 2024-11-07
Added GCloud Pub/Sub message queue implementation

mike182uk:
refs [AP-566](https://linear.app/ghost/issue/AP-566/implement-a-pubsub-backed-queue-for-fedify)

Added and configured a GCloud Pub/Sub message queue implementation for Fedify


src/fedify/mq/gcloud-pubsub-mq.ts
```diff
@@ -0,0 +1,105 @@
...
+        message: any,
+        options?: MessageQueueEnqueueOptions,
+    ): Promise<void> {
+        const delay = options?.delay?.total('millisecond') ?? 0;
+
+        this.logger.info(
+            `Enqueuing message [FedifyID: ${message.id}] with delay: ${delay}ms`,
+        );
+
+        if (delay > 0) {
```
allouis, line 38: Do we know when/why we need to support a delay?

src/fedify/mq/gcloud-pubsub-mq.ts
```diff
@@ -0,0 +1,105 @@
...
+                message.nack();
+
+                this.logger.error(
+                    `Failed to handle message [FedifyID: ${fedifyId}, PubSubID: ${message.id}]: ${error}`,
+                );
+            }
+        });
+
+        return await new Promise((resolve) => {
+            options.signal?.addEventListener('abort', () => {
```
allouis, line 97: If signal is not defined, this promise will never resolve - is that expected?

src/app.ts
```diff
@@ -153,8 +162,56 @@ export type ContextData = {
 
 const fedifyKv = await KnexKvStore.create(client, 'key_value');
 
+let messageQueue: MessageQueue | undefined;
+
+if (process.env.USE_MQ === 'true') {
+    logging.info('Message queue is enabled');
+
+    try {
```
allouis, line 170: Everything in this block feels like "factory" code to me - I think it would be a lot nicer to have a static create factory method on the class, which can handle the assertions that the topic exists etc...

```
class GCloudPubSubMessageQueue {  
  ...
  async static create() {
    // all the init checks etc...
    return new GCloudPubSubMessageQueue(args);
  }
}
```

You can then mark the `constructor` as `private` and be sure that any instance of the class has had all of the requirements checked and met

src/fedify/mq/gcloud-pubsub-mq.ts
mike182uk: Going off the docs → https://fedify.dev/manual/mq#implement-messagequeue-enqueue-method

I think this is an internal Fedify thing for the retries (exponential backoff)

src/fedify/mq/gcloud-pubsub-mq.ts
```diff
@@ -0,0 +1,105 @@
...
+                    attributes: {
+                        fedifyId: message.id,
+                    },
+                });
+
+            this.logger.info(
+                `Message [FedifyID: ${message.id}] was enqueued [PubSubID: ${messageId}]`,
+            );
+        } catch (error) {
+            this.logger.error(
```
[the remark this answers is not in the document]
mike182uk: The most likely culprit for failure would be network issues, or something borked on GCP's end (topic missing, refusal to publish message). I would if this actually gets retried and we should just let the error be thrown 🤔

src/fedify/mq/gcloud-pubsub-mq.ts
mike182uk: I think so yeh, from the docs:

> A Promise object it returns should never resolve unless the given [signal](https://jsr.io/@fedify/fedify@1.3.0-dev.492+450a6939/doc/~/MessageQueueListenOptions.signal) is triggered.

mike182uk: Oh my bad, you meant if `signal` is not provided at all. I wonder why that would be the case 🤔 

@dahlia In what cases would `signal` not be provided?

dahlia: The `signal` is usually not provided, unless you're [manually starting the worker](https://fedify.dev/manual/mq#separating-message-processing-from-the-main-process). And yes, this promise is expected to be never resolved unless any abort signal is invoked.

src/app.ts
mike182uk: Yeh, i did actually start down a similar path to this, but outside of the class itself (felt better, made easier to test). See what you think → https://github.com/TryGhost/ActivityPub/pull/143/commits/ea15558c1e29fc888198c391a5f855156ed2ed29

src/fedify/mq/gcloud-pubsub-mq.ts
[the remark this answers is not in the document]
mike182uk: Have updated this to `throw` the error (so a retry happens) and capture the exception to in Sentry

src/fedify/mq/gcloud-pubsub-mq.ts
allouis: What about if the underlying listener is aborted unexpectedly - should the promise resolve then?

allouis (approved):
Questions surrounding the returned promise from `listen` but looks good to get this in a tested

PR #180, opened 2024-11-16
Refactored GCloud Pub/Sub message queue implementation

mike182uk:
refs [AP-566](https://linear.app/ghost/issue/AP-566/implement-a-pubsub-backed-queue-for-fedify), [TryGhost/ActivityPub#143](https://github.com/TryGhost/ActivityPub/pull/143)

Refactorings include:
- Renamed `gcloud-pubsub-mq` to `gcloud-pubsub-push-mq`
- Colocated all message queue related logic in `src/mq/gcloud-pubsub-push` for better reasonability / maintainability
- Reworked how the Hono middleware interfaces with the message queue (removed app-wide event bus)
- Reworked message queue error handling to not use Sentry directly
- Improved strcutured logging
- Improved test coverage
- Added inline documentation


src/app.ts
```diff
@@ -157,34 +156,21 @@ export type ContextData = {
...
     try {
-        const host = String(process.env.MQ_PUBSUB_HOST) || undefined;
-        const emulatorMode = process.env.NODE_ENV !== 'production';
-        const projectId = String(process.env.MQ_PUBSUB_PROJECT_ID) || undefined;
-        const topicName = String(process.env.MQ_PUBSUB_TOPIC_NAME);
-        const subscriptionName = String(
-            process.env.MQ_PUBSUB_SUBSCRIPTION_NAME,
-        );
+        queue = await createMessageQueue(logging, {
+            pubSubHost: String(process.env.MQ_PUBSUB_HOST) || undefined,
```
allouis, line 166: Why are we using `String` here? `process.env` should be a `Dict<string>` which means the properties will always be a `string` or `undefined`

src/mq/gcloud-pubsub-push/mq.ts
```diff
@@ -0,0 +1,433 @@
...
+    /**
+     * Handles a message
+     *
+     * Ensure that the message queue is listening before calling this method
+     *
+     * @param message {Message} Message to handle
+     */
+    handleMessage(message: Message): void {
+        if (this.isListening === false) {
+            this.logger.warn(
```
allouis, line 244: This warns, but still emits the event - I think it should probably throw here as it's not a valid interaction?

src/mq/gcloud-pubsub-push/mq.ts
```diff
@@ -0,0 +1,433 @@
...
+ * const queue = await createMessageQueue(...);
+ *
+ * app.post('/mq', handlePushMessage(queue));
+ * ```
+ */
+export function handlePushMessage(
+    mq: GCloudPubSubPushMessageQueue,
+): (ctx: Context) => Promise<Response> {
+    return async (ctx: Context) => {
+        const json = await ctx.req.json<IncomingPushMessageJson>();
```
allouis, line 306: I don't think we can declare the `IncomingPushMessageJson` type here - as we've no idea what shape the data is? I think this should be `unknown` and either have manual type checks, or use `zod` to validate the type like we do for webhooks

src/app.ts
mike182uk: Yep no need for this, thanks 👍

Fixed in https://github.com/TryGhost/ActivityPub/pull/180/commits/db60dc3e5a7d0e345c52068a6d8d1f31a1f998f2

src/mq/gcloud-pubsub-push/mq.ts
```diff
@@ -0,0 +1,433 @@
...
+            return;
+        }
+
+        this.logger.info(`Enqueuing message [FedifyID: ${message.id}]`, {
+            fedifyId: message.id,
+        });
+
+        try {
+            const messageId = await this.pubSubClient
+                .topic(getFullTopic(this.pubSubClient.projectId, this.topic))
```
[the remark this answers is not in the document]
mike182uk: Good shout, updated in https://github.com/TryGhost/ActivityPub/pull/180/commits/00bca501190ee7408110121a8da2684d84f733bd

src/mq/gcloud-pubsub-push/mq.ts
mike182uk: 👍 Refactored how this works in https://github.com/TryGhost/ActivityPub/pull/180/commits/e076990c4502436bb9d27a00ec9568d3986944b4

src/mq/gcloud-pubsub-push/mq.ts
```diff
@@ -0,0 +1,362 @@
...
+     */
+    async listen(
+        handler: (message: any) => Promise<void> | void,
+        options: MessageQueueListenOptions = {},
+    ): Promise<void> {
+        this.messageHandler = handler;
+
+        return await new Promise((resolve) => {
+            options.signal?.addEventListener('abort', () => {
+                resolve();
```
allouis, line 147: I think we need to unassign `this.messageHandler` here? Or more specifically have some way of making `this.isListening` evaluate to false

src/mq/gcloud-pubsub-push/mq.ts
```diff
@@ -0,0 +1,362 @@
...
+        });
+    }
+
+    /**
+     * Handles a message
+     *
+     * @param message {Message} Message to handle
+     */
+    async handleMessage(message: Message): Promise<void> {
+        if (this.messageHandler === undefined) {
```
allouis, line 158: Would be good to use `this.isListening` here

allouis (approved):
Two small comments but otherwise looks great!

src/mq/gcloud-pubsub-push/mq.ts
mike182uk: 👍 Updated in https://github.com/TryGhost/ActivityPub/pull/180/commits/12ea3f6c2a592156b7a771b7be3c629b09612dbe

src/mq/gcloud-pubsub-push/mq.ts
mike182uk: Yeh I did try this, but further down where we use `this.messageHandler` it was giving me a type error about `this.messageHandler` being potentially undefined (which it wouldn't be as the `this.isListening` check would fail in that case). Any ideas how to get around this?

PR #207, opened 2024-11-25
Added support for updating the site actor

allouis:
ref https://linear.app/ghost/issue/AP-592

When we fetch the site data we'll attempt to update the site actor too, this
will ensure that when Ghost boots any changes are propagated to us.

We're handling the sending of the Update activity ourself, and there's room to
remove duplication here. I'd like to think about what pattern we use for this.
Ideally I think a repository would be responsible for this or at least abstract
it.

### DIFF
--- a/src/handlers.ts
+++ b/src/handlers.ts
@@ -25,7 +25,7 @@ import {
 } from './helpers/activitypub/activity';
 import { getSiteSettings } from './helpers/ghost';
 import { toURL } from './helpers/uri';
-import type { PersonData } from './helpers/user';
+import { type PersonData, getUserData } from './helpers/user';
 import { addToList, removeFromList } from './kv-helpers';
 import { lookupActor, lookupObject } from './lookup-helpers';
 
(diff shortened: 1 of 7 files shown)


src/handlers.ts
```diff
@@ -543,30 +546,103 @@ export async function postPublishedWebhook(
...
+        });
+
+        await ctx
+            .get('globaldb')
+            .set([update.id!.href], await update.toJsonLd());
+
+        await apCtx.sendActivity({ handle }, 'followers', update, {
+            preferSharedInbox: true,
+        });
+    }
```
mike182uk, line 593: Should this also be apart of `updateSiteActor` and reused across the 2 handlers?

src/handlers.ts
```diff
@@ -543,30 +546,103 @@ export async function postPublishedWebhook(
...
+
+    const current = await db.get<PersonData>(['handle', handle]);
+
+    if (
+        current &&
+        current.icon === settings.site.icon &&
+        current.name === settings.site.title &&
+        current.summary === settings.site.description
+    ) {
+        logger.info('No site settings changed, nothing to do');
```
mike182uk, line 615: Maybe we should update this now it is utilised in multiple places - `No site settings changed, not updating actor` or something? Or maybe move out altogether and let the handler log as it sees fit?

src/handlers.ts
```diff
@@ -543,30 +546,103 @@ export async function postPublishedWebhook(
...
+
+    return new Response(JSON.stringify(site), {
+        status: 200,
+        headers: {
+            'Content-Type': 'application/json',
+        },
+    });
+}
+
+async function updateSiteActor(db: KvStore, logger: Logger, host: string) {
```
mike182uk, line 603: Could move this to the actor helpers and add a test?

src/helpers/activitypub/actor.unit.test.ts
```diff
@@ -226,3 +235,123 @@ describe('isHandle', () => {
...
+        const host = 'example.com';
+
+        const apCtx = mockApContext(db, globaldb);
+
+        const result = await updateSiteActor(apCtx, getSiteSettings, host);
+
+        assert(result === false);
+    });
+
+    it('Should update the site actor if the site settings have changed', async () => {
```
mike182uk, line 318:
```suggestion
    it('should update the site actor if the site settings have changed', async () => {
```

src/helpers/activitypub/actor.unit.test.ts
```diff
@@ -226,3 +235,123 @@ describe('isHandle', () => {
...
+            },
+        });
+
+        const host = 'example.com';
+
+        const apCtx = mockApContext(db, globaldb);
+
+        const result = await updateSiteActor(apCtx, getSiteSettings, host);
+
+        assert(result === false);
```
mike182uk, line 315:
```suggestion
        expect(result).toBe(false);
```

For consistency with other tests in this file

src/helpers/activitypub/actor.unit.test.ts
```diff
@@ -226,3 +235,123 @@ describe('isHandle', () => {
...
+            },
+        });
+
+        const host = 'example.com';
+
+        const apCtx = mockApContext(db, globaldb);
+
+        const result = await updateSiteActor(apCtx, getSiteSettings, host);
+
+        assert(result === true);
```
mike182uk, line 355: Should we make some assertions that the actual update happened with the values we expect + a message was published?

PR #210, opened 2024-11-26
Added handling for `Group` announced `Create` activities

mike182uk:
refs [AP-608](https://linear.app/ghost/issue/AP-608/ghost-does-not-work-with-lemmy-or-wordpress)

Added handling for `Group` announced `Create` activities so that these activities are processed by the service. Activities are usually only announced as part of Group federation

See https://codeberg.org/fediverse/fep/src/branch/main/fep/1b12/fep-1b12.md for more details

### DIFF
--- a/features/handle-group-announce.feature
+++ b/features/handle-group-announce.feature
@@ -1,0 +1,14 @@
+Feature: Handling activities announced by a Group
+  Background:
+    Given a Person "Alice"
+    And a Group "Wonderland"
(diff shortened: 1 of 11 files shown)


features/handle-group-announce.feature
```diff
@@ -0,0 +1,14 @@
+Feature: Handling activities announced by a Group
+  Background:
```
allouis, line 2: Nice

allouis (approved):

PR #233, opened 2024-12-17
Ensured activities are delivered to all followers

mike182uk:
refs [AP-638](https://linear.app/ghost/issue/AP-638/ghost-is-not-delivering-activities-to-all-followers)

Added logic to ensure that activities are delivered to all followers and not just the first page of followers

### DIFF
--- a/features/step_definitions/stepdefs.js
+++ b/features/step_definitions/stepdefs.js
@@ -534,8 +534,9 @@ Given('we are followed by:', async function (actors) {
         const object = this.actors.Us;
         const activity = await createActivity('Follow', object, actor);
 
-        this.activities[name] = activity;
-        this.objects[name] = object;
+        const key = `Follow(Us)_${name}`;
+        this.activities[key] = activity;
+        this.objects[key] = object;
 
         // Send the follow activity to the inbox
         this.response = await fetchActivityPub(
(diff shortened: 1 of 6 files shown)


features/step_definitions/stepdefs.js
```diff
@@ -524,6 +524,50 @@ Given('we follow {string}', async function (name) {
...
+    for (const { name, type } of actors.hashes()) {
+        // Create the actor
+        this.actors[name] = await createActor(name, { type });
+
+        // Create the follow activity
+        const actor = this.actors[name];
+        const object = this.actors.Us;
+        const activity = await createActivity('Follow', object, actor);
+
+        this.activities[name] = activity;
```
allouis, line 537: Not a big deal, but this is a bit of a vague name - not sure what a better one would be, we can leave it for now I reckon, until we actually need to use them!

allouis (approved):

features/step_definitions/stepdefs.js
mike182uk: Ah, I missed this, copy pasta 🍝

Have updated in [4459752](https://github.com/TryGhost/ActivityPub/pull/233/commits/4459752c342a44ac22f7ddc88f0c8d504bc5cef0) 👍

PR #244, opened 2025-01-06
Ensure only public posts are published to the Fediverse

mike182uk:
refs [AP-461](https://linear.app/ghost/issue/AP-641/only-public-posts-should-be-sent-to-the-fediverse)

Added logic to ensure only public posts are published to the Fediverse


mike182uk:
The requirements for this are incorrect so going to rework and create a new PR

PR #275, opened 2025-01-22
Created new Account for Site on initialisation

allouis:
ref https://linear.app/ghost/issue/AP-658

This is required as going forward actors will be based on Accounts

### DIFF
--- a/src/site/site.service.integration.test.ts
+++ b/src/site/site.service.integration.test.ts
@@ -1,12 +1,15 @@
-import { beforeEach, describe, expect, it } from 'vitest';
+import { beforeEach, describe, expect, it, vi } from 'vitest';
 
 import { TABLE_SITES } from '../constants';
 import { client as db } from '../db';
 
+import { AccountService } from '../account/account.service';
+import type { Account } from '../account/types';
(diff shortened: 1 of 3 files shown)


src/site/site.service.integration.test.ts
```diff
@@ -15,21 +18,28 @@ describe('SiteService', () => {
         await db(TABLE_SITES).truncate();
         await db.raw('SET FOREIGN_KEY_CHECKS = 1');
 
+        accountService = Object.create(AccountService.prototype);
```
mike182uk, line 21: Nice!

src/site/site.service.ts
```diff
@@ -65,6 +69,11 @@ export class SiteService {
             throw new Error(`Site initialisation failed for ${host}`);
         }
 
+        const internalAccount = await this.accountService.createInternalAccount(
+            newSite,
+            'index',
```
mike182uk, line 74: Use `ACTOR_DEFAULT_HANDLE` from consts maybe?

allouis: Will do this with a cleanup of site service to use consts for tablenames too

PR #287, opened 2025-01-27
Updated application to use `isFollowedByDefaultSiteAccount`

mike182uk:
no refs

Updated application to use `isFollowedByDefaultSiteAccount` to determine if an actor is followed by the default site account. This should replace the old `isFollowing` helper function that reads from the old KV store.

Affected areas:

- Search results to indicate if a search result account is followed by the
  default site account
- Profile result to indicate if a profile is followed by the default
  site account
- Follows result to indicate if a profile follow result is followed by the
  default site account
- `Create` handler when checking if an item should be added to the inbox
- `Announce` handler when checking if an item should be added to the inbox
- `AnnouncedCreate` handler when checking if an item should processed

This change also ensures that `site` on the `AppContext` is typed correctly so that request handlers do not have to manually resolve the site via the site service

### DIFF
--- a/src/app.ts
+++ b/src/app.ts
@@ -46,6 +46,8 @@ import {
     actorDispatcher,
     articleDispatcher,
     createAcceptHandler,
+    createAnnounceHandler,
+    createCreateHandler,
     createDispatcher,
     createFollowHandler,
     createFollowersCounter,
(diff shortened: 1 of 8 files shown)


src/app.ts
```diff
@@ -380,10 +390,7 @@ export type HonoContextVariables = {
     globaldb: KvStore;
     logger: Logger;
     role: GhostRole;
-    site: {
-        host: string;
-        webhook_secret: string;
-    };
+    site: Site;
```
mike182uk, line 393: Updated this to use the `Site` type as this is what gets set in the middleware

PR #288, opened 2025-01-27
Removed duplicate method for fetching default account

allouis:
Whilst we were building out the infrastructure for accounts and sites we ended up with some duplicate code, this consolidates it into the account service, slightly updating the signature to never resolve with null, throwing if an account is not found instead. We also clean up all the call sites and tests.

### DIFF
--- a/src/account/account.service.ts
+++ b/src/account/account.service.ts
@@ -138,7 +138,7 @@ export class AccountService {
      *
      * @param site Site
      */
-    async getDefaultAccountForSite(site: Site): Promise<Account | null> {
+    async getDefaultAccountForSite(site: Site): Promise<Account> {
         const users = await this.db(TABLE_USERS).where('site_id', site.id);
 
         if (users.length === 0) {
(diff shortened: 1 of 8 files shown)


src/account/account.service.ts
```diff
@@ -153,9 +153,15 @@ export class AccountService {
 
         // We can safely assume that there is an account for the user due to
         // the foreign key constraint on the users table
-        return await this.db(TABLE_ACCOUNTS)
+        const account = await this.db(TABLE_ACCOUNTS)
             .where('id', user.account_id)
             .first();
+
+        if (!account) {
```
mike182uk, line 160: ooh nice, i didn't think about putting this here, but it actually makes a ton of sense seeming though its unexpected that a site doesn't have a default account

mike182uk: I think there might be a couple of places where we do this check outside of the service, could we clean those up as part of this PR?

mike182uk: also could we add a test case for this change in functionality?

PR #290, opened 2025-01-30
Unfollow api

allouis:


src/account/account.service.ts
```diff
@@ -120,6 +120,24 @@ export class AccountService {
...
 
+    /**
+     * Record an account follow
+     *
+     * @param follower The follower account
+     * @param following This account being followed
+     */
+    async removeAccountFollow(
+        follower: Account,
+        following: Account,
```
mike182uk, line 131: I think we should switch these around to match `recordAccountFollow` which first takes the account to follow, and then the account that is doing the following.

Also i wonder if `followed` and `follower` could be nicer terms here - i.e "Heres the `followed` account, remove this `follower`" (i suppose `following` also works here)

src/dispatchers.ts
```diff
@@ -463,6 +463,53 @@ export async function handleAnnoucedCreate(
     await addToList(ctx.data.db, ['inbox'], create.id.href);
 }
 
+export const createHandleUndo = (accountService: AccountService) =>
```
mike182uk, line 466: Can we call this `createUndoHandler` to keep consistency with the other handlers that are created in this manner? i.e `createFollowHandler`

src/dispatchers.ts
```diff
@@ -463,6 +463,53 @@ export async function handleAnnoucedCreate(
...
+            follow.objectId.href,
+        );
+        if (!unfollowing) {
+            ctx.data.logger.info('Could not find unfollowing');
+            return;
+        }
+
+        await ctx.data.globaldb.set([undo.id.href], await undo.toJsonLd());
+
+        await removeFromList(ctx.data.db, ['following'], follow.objectId.href);
```
mike182uk, line 506: We shouldn't need to do this right? All the followers / following logic should be using the `follows` table now (i think 😅)

src/handlers.ts
```diff
@@ -304,6 +304,101 @@ export async function replyAction(
...
+            ctx.get('site'),
+        );
+
+        if (actorToUnfollow.id!.href === account.ap_id) {
+            return new Response(null, {
+                status: 400,
+            });
+        }
+
+        let unfolloweeAccount = await accountService.getAccountByApId(
```
mike182uk, line 336: Naming is hard 🙃

mike182uk: How about `followedAccount` or `accountToUnfollow`?

src/handlers.ts
```diff
@@ -304,6 +304,101 @@ export async function replyAction(
...
+
+        const account = await accountService.getDefaultAccountForSite(
+            ctx.get('site'),
+        );
+
+        if (actorToUnfollow.id!.href === account.ap_id) {
+            return new Response(null, {
+                status: 400,
+            });
+        }
```
mike182uk, line 334: Do we need to account for this? We already have logic that prevents us from following ourself right?

src/handlers.ts
```diff
@@ -304,6 +304,101 @@ export async function replyAction(
...
+        let unfolloweeAccount = await accountService.getAccountByApId(
+            actorToUnfollow.id!.href,
+        );
+
+        // TODO I think we can exit early here - there is obviously no follow relation if there is no account
+        if (!unfolloweeAccount) {
+            unfolloweeAccount = await accountService.createExternalAccount(
+                await mapActorToExternalAccountData(actorToUnfollow),
+            );
+        }
```
mike182uk, line 345: Yeh i think we can bail early here - If we can't find an account for the actor being unfollowed, then there isn't a record of them following us so no need to carry on

src/handlers.ts
```diff
@@ -304,6 +304,101 @@ export async function replyAction(
...
+            });
+        }
+
+        // Need to get the follow
+        const unfollowId = apCtx.getObjectUri(Undo, {
+            id: uuidv4(),
+        });
+
+        const follow = new Follow({
+            id: null,
```
mike182uk, line 364: 🤔 This is what you was talking about with it being tricky to grab the original follow activity?

I assume this is safe to do, i guess other implementations are just going to grab the object anyways rather than try match IDs of activities up (as that would mean they also have to store the original activity)

src/handlers.ts
```diff
@@ -304,6 +304,101 @@ export async function replyAction(
...
+
+        const unfollowJson = await unfollow.toJsonLd();
+
+        await ctx.get('globaldb').set([unfollow.id!.href], unfollowJson);
+
+        await removeFromList(
+            ctx.get('db'),
+            ['following'],
+            actorToUnfollow.id!.href,
+        );
```
mike182uk, line 383: We shouldn't need to do this right? All the followers / following logic should be using the follows table now (i think 😅)

src/handlers.ts
```diff
@@ -304,6 +304,101 @@ export async function replyAction(
...
+        await apCtx.sendActivity(
+            { handle: ACTOR_DEFAULT_HANDLE },
+            actorToUnfollow,
+            unfollow,
+        );
+
+        await accountService.removeAccountFollow(account, unfolloweeAccount);
+
+        // We return the actor because the serialisation of the object property is not working as expected
+        return new Response(JSON.stringify(unfollowJson), {
```
mike182uk, line 394: The comment doesn't match what is happening here 😅 We are returning the activity JSON and not the actor

Should we keep the output here consistent with the `followActon`? (return the `actor`)

features/follow-account.feature
```diff
@@ -13,3 +13,16 @@ Feature: Follow accounts from their handle
...
+    Given we follow "Alice"
+    Then the request is accepted
+    Given a "Accept(Follow(Alice))" Activity "A" by "Alice"
+    And "Alice" sends "A" to the Inbox
+    And "A" is in our Inbox
+    Then the object "Alice" should be in the "following" collection
+    Given we unfollow "Alice"
+    Then the request is accepted
+    Then the object "Alice" should not be in the "following" collection
+    And Activity "Unfollow(Alice)" is sent to "Alice"
```
mike182uk, line 28: I feel the scenarios read better when we follow the `Given`, `When`, `Then` order:

- `Given` the system is in some state (can be background also)
- `When` an action occurs
- `Then` we expected `x` to happen

What do you think to changing this around slightly:

```
Scenario: We can unfollow an account
	Given an Actor "Person(Alice)"
	And we follow "Alice"
	And the request is accepted
	And a "Accept(Follow(Alice))" Activity "A" by "Alice"
	And "Alice" sends "A" to the Inbox
	And "A" is in our Inbox
	And the object "Alice" is in the "following" collection
	When we unfollow "Alice"
	Then the request is accepted
	And the object "Alice" is not in the "following" collection
	And Activity "Unfollow(Alice)" is sent to "Alice"
```

I know there are alot of `And`s which make it tedious, this is just a symptom of us having lots of granular steps (which i don't think is necessarily a bad thing), but i really like its clear that `When` `x` happens `Then` `y` should happen

features/follow-account.feature
```diff
@@ -13,3 +13,16 @@ Feature: Follow accounts from their handle
   Scenario: We cannot follow ourselves
     Given we follow "Us"
     Then the request is rejected with a 400
+
```
mike182uk, line 16: Should we add a scenario for being unfollowed by somebody else?

src/account/account.service.ts
```diff
@@ -120,6 +120,24 @@ export class AccountService {
             .ignore();
     }
 
+    /**
+     * Record an account follow
+     *
+     * @param follower The follower account
+     * @param following This account being followed
+     */
+    async removeAccountFollow(
```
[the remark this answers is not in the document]
mike182uk: Could we add a smol integration test for this also? I know technically it would be covered by the e2e test, but it would be good to get a little coverage at this level, and complements the testing that we have for all the other methods in the service

src/handlers.ts
allouis: Yess, I was thinking the same!

src/handlers.ts
allouis: Yeah this works for our implementation, as well as mastodons, and we can address full compatibility in future!

src/handlers.ts
allouis: I like returning the activity because it makes testing a lot easier - I think follow should probs return the activity too, but not part of this PR

src/account/account.service.ts
allouis: I've switched argument order but kept naming for now for consistency with the db

src/handlers.ts
allouis: Removed comment

features/follow-account.feature
allouis: Dun!

mike182uk (approved):

PR #302, opened 2025-02-05
Added API endpoint to repost a post

sagzy:
ref https://linear.app/ghost/issue/AP-702

- added `POST /actions/reposts/:id endpoint` to create a Repost
- a Repost is the equivalent of the "Announce" activity in the ActivityPub vocabulary, or "Boost" in the Mastodon vocabulary

### DIFF
--- a/features/repost-activity.feature
+++ b/features/repost-activity.feature
@@ -1,4 +1,3 @@
-@only
 Feature: Reposting a post
   In order to share content with my followers
   As a user
--- a/features/repost-activity.feature
+++ b/features/repost-activity.feature
@@ -1,30 +1,31 @@
-Feature: Reposting an object
+@only
+Feature: Reposting a post
+  In order to share content with my followers
   As a user
-  I want to repost an object in my feed
-  So that I can share the content with my followers
+  I want to be able to repost a post in my feed
 
-  Scenario: Reposting an object that has not been reposted before
+  Scenario: Reposting a post that has not been reposted before
     Given an Actor "Person(Alice)"
-    Given we follow "Alice"
-    Then the request is accepted
-    Given a "Accept(Follow(Alice))" Activity "Accept" by "Alice"
+    And we follow "Alice"
+    And the request is accepted
+    And a "Accept(Follow(Alice))" Activity "Accept" by "Alice"
     And "Alice" sends "Accept" to the Inbox
     And "Accept" is in our Inbox
-    Given a "Create(Note)" Activity "Note" by "Alice"
-    When "Alice" sends "Note" to the Inbox
+    And a "Create(Note)" Activity "Note" by "Alice"
+    And "Alice" sends "Note" to the Inbox
     And "Note" is in our Inbox
-    And we repost the object "Note"
+    When we repost the object "Note"
     Then a "Announce(Note)" activity is sent to "Alice"
 
-  Scenario: Reposting an object that has been reposted before
+  Scenario: Reposting an post that has been reposted before
     Given an Actor "Person(Alice)"
-    Given we follow "Alice"
-    Then the request is accepted
-    Given a "Accept(Follow(Alice))" Activity "Accept" by "Alice"
+    And we follow "Alice"
+    And the request is accepted
+    And a "Accept(Follow(Alice))" Activity "Accept" by "Alice"
     And "Alice" sends "Accept" to the Inbox
     And "Accept" is in our Inbox
-    Given a "Create(Note)" Activity "Note" by "Alice"
-    When "Alice" sends "Note" to the Inbox
+    And a "Create(Note)" Activity "Note" by "Alice"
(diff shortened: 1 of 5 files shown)


src/handlers.ts
```diff
@@ -573,3 +574,80 @@ export async function inboxHandler(
         },
     );
 }
+
+export async function repostAction(
```
sagzy, line 578: @allouis @mike182uk I haven't used the newer pattern introduced in `createFollowActionHandler` — would it bring any value here for Repost?

mike182uk: Nice one, i'd say use this pattern here as also we are very likely going to need to pass in some kind of service to persist the post once we have migrated away from the kv store 👍

src/handlers.ts
```diff
@@ -573,3 +574,80 @@ export async function inboxHandler(
...
+        actor: actor,
+        object: objectToRepost,
+        to: PUBLIC_COLLECTION,
+        cc: apCtx.getFollowersUri(ACTOR_DEFAULT_HANDLE),
+    });
+
+    const announceJson = await announce.toJsonLd();
+
+    await ctx.get('globaldb').set([announce.id!.href], announceJson);
+    await addToList(ctx.get('db'), ['reposted'], announce.id!.href);
```
mike182uk, line 618: @sagzy this is where we will also want to add to the outbox - https://github.com/TryGhost/ActivityPub/blob/main/src/handlers.ts#L273

The `outbox` contains posts created / reposted by the current user

features/repost-activity.feature
```diff
@@ -0,0 +1,32 @@
+Feature: Reposting an object
```
mike182uk, line 1: Maybe we use the terminology `post` instead of `object` in this?

features/repost-activity.feature
```diff
@@ -0,0 +1,32 @@
...
+    Given we follow "Alice"
+    Then the request is accepted
+    Given a "Accept(Follow(Alice))" Activity "Accept" by "Alice"
+    And "Alice" sends "Accept" to the Inbox
+    And "Accept" is in our Inbox
+    Given a "Create(Note)" Activity "Note" by "Alice"
+    When "Alice" sends "Note" to the Inbox
+    And "Note" is in our Inbox
+    And we repost the object "Note"
+    Then a "Announce(Note)" activity is sent to "Alice"
```
mike182uk, line 17: How about:

```
  Scenario: Reposting a post that has not been reposted before
    Given an Actor "Person(Alice)"
    And we follow "Alice"
    And the request is accepted
    And a "Accept(Follow(Alice))" Activity "Accept" by "Alice"
    And "Alice" sends "Accept" to the Inbox
    And "Accept" is in our Inbox
    And a "Create(Note)" Activity "Note" by "Alice"
    And "Alice" sends "Note" to the Inbox
    And "Note" is in our Inbox
    When we repost "Note"
    Then a "Announce(Note)" activity is sent to "Alice"
```

This reads better imo, and follows the pattern of `Given`, `When`, `Then` (I know we don't always follow this 😅)

features/repost-activity.feature
```diff
@@ -0,0 +1,32 @@
...
+
+  Scenario: Reposting an object that has been reposted before
```
mike182uk, line 19: Based on the feedback above, see if you can clean this up 👍

src/app.ts
```diff
@@ -380,7 +382,11 @@ fedify.setObjectDispatcher(
     '/.ghost/activitypub/undo/{id}',
     spanWrapper(undoDispatcher),
 );
-
+fedify.setObjectDispatcher(
```
mike182uk, line 385: Nice one, wasn't expecting to do this, but definitely something we will need 👌

features/repost-activity.feature
```diff
@@ -0,0 +1,32 @@
+Feature: Reposting an object
+  As a user
+  I want to repost an object in my feed
+  So that I can share the content with my followers
```
mike182uk, line 4: Maybe a slight twist on this:

```
In order to share content with my followers
As a user
I want to be repost a post in my feed
```

src/handlers.ts
sagzy: Added in a987a82930267c20beedf7c3030e4223e3ead97d

src/handlers.ts
sagzy: Added in a987a82930267c20beedf7c3030e4223e3ead97d

features/repost-activity.feature
sagzy: Updated in 9faffb4b1a509c7027d792aba0b493008496e9c9

features/repost-activity.feature
sagzy: Updated in 9faffb4b1a509c7027d792aba0b493008496e9c9

PR #311, opened 2025-02-11
Update account data using site info while creating internal account

vershwal:
Ref https://linear.app/ghost/issue/AP-737/

- While creating an account we used the default name, icon and bio
- This change will use the site settings to fetch the name, icon and bio while creating the AP account and we'll use the default ones only if the info is missing in site setting.

### DIFF
--- a/src/account/account.service.integration.test.ts
+++ b/src/account/account.service.integration.test.ts
@@ -94,7 +94,7 @@ describe('AccountService', () => {
             const account = await service.createInternalAccount(
                 site,
                 username,
-                siteSettings
+                siteSettings,
             );
 
             // Assert the created account was returned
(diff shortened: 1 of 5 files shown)


allouis (commented):
Nice catch 🎣 

Change overall looks good, but I think we want to keep the account service focused on accounts, and the site service focused on sites - So something like passing `name`, `bio` and `avatar_url` to the `createInternalAccount` and doing the `settings?.site?.<thing>` in the SiteService

I reckon adding extra params get's a bit too much:
```ts
    async createInternalAccount(
        site: Site,
        username: string,
        name?: string,
        bio?: string,
        avatar_url?: string,
    ): Promise<Account> {
```

So maybe bundle it into an object like
```ts
    async createInternalAccount(
        site: Site,
        accountData: CreateInternalAccountDataOrSomething,
    ): Promise<Account> {
```

It's a bit of an annoying change but I do think it's useful to keep all the site related stuff in one place

allouis (approved):

PR #315, opened 2025-02-13
Clean up old kv store based code

vershwal:
Refs https://linear.app/ghost/issue/AP-671


features/step_definitions/stepdefs.js
```diff
@@ -894,7 +895,7 @@ Then(
 );
 
 Then(
-    'the object {string} should not be in the {string} collection',
+    'the object {string} should not be in the {string} collection', // Test failing for accept-follows.feature
```
vershwal, line 898: These two tests are failing after removing the kv-store based code. My assumption was that these should still work end to end after removing the old code.

src/dispatchers.ts
```diff
@@ -132,33 +130,7 @@ export function createFollowHandler(accountService: AccountService) {
...
-        const currentFollowers =
-            (await ctx.data.db.get<string[]>(['followers'])) ?? [];
-        const shouldRecordFollower =
-            currentFollowers.includes(sender.id.href) === false;
-
-        // Add follow activity to inbox
-        const followJson = await follow.toJsonLd();
-
-        ctx.data.globaldb.set([follow.id.href], followJson);
-        await addToList(ctx.data.db, ['inbox'], follow.id.href);
```
allouis, line 144: We want to keep the `inbox` and the follow storage for now because we haven't migrated away from the key_value for these - only for the followings and followers!

PR #316, opened 2025-02-13
Added a new note to posts table

vershwal:
Refs https://linear.app/ghost/issue/AP-717

### DIFF
--- a/src/http/api/note.ts
+++ b/src/http/api/note.ts
@@ -42,22 +42,13 @@ export async function handleCreateNote(
 
     // Save to posts table when a note is created
     const account = await accountRepository.getBySite(ctx.get('site'));
-    const post = new Post(
-        null,
-        null,
-        account,
-        PostType.Note,
-        Audience.Public,
-        null,
-        null,
-        data.content,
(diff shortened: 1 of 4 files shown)


src/http/api/note.ts
```diff
@@ -33,6 +40,25 @@ export async function handleCreateNote(ctx: AppContext) {
...
+    const post = new Post(
+        null,
+        null,
+        account,
+        PostType.Note,
+        Audience.Public,
+        null,
+        null,
+        data.content,
+        new URL(account?.avatarUrl || ''),
```
vershwal, line 54: I added a random url to make it work, What should be the url of a note? The url field is 'not null' as per the schema but the [doc](https://www.notion.so/ghost/Posts-Migration-18851439c03080eaa24bf6abd7241e01?pvs=4) doesn't specify what the url should be in case of notes?

allouis: Good question, nice catch! Our notes don't really have a canonical URL - so for now we should use the same value as `ap_id` - that will require a change to the `Post` model to be able to handle

PR #317, opened 2025-02-13
Updated Like activity handler to write to `posts` & `likes` table

allouis:


src/post/post.repository.knex.ts
```diff
@@ -1,38 +1,151 @@
...
+                    ap_id: post.apId.href,
+                });
+
+                if (potentiallyNewLikes.length > 0) {
+                    const likesToInsert = potentiallyNewLikes.map(
+                        (accountId) => ({
+                            account_id: accountId,
+                            post_id: id,
+                        }),
+                    );
```
vershwal, line 109:
```
if (potentiallyNewLikes.length > 0) {
                    const likesToInsert = potentiallyNewLikes.map(
                        (accountId) => ({
                            account_id: accountId,
                            post_id: id,
                        }),
                    );
```

This part of the code is repeated in both if and else block, maybe we can have `likesToInsert` outside the if block along with `const potentiallyNewLikes = post.getPotentiallyNewLikes();`.

allouis: They're slightly different because in one branch of the block we don't have access to the `id` until we write to the database, so we can't pull this out

src/account/account.repository.knex.ts
```diff
@@ -42,4 +42,48 @@ export class KnexAccountRepository {
             site,
         );
     }
+
+    async getByApId(apId: URL): Promise<Account | null> {
```
mike182uk, line 46: 🤔 would this be more appropriate at the service level? Feels weird creating an object literal of a `Site` within the method (and leaks `site` implementation details into this repository)

mike182uk: Hm, i see we already have this method at this level. Maybe we pass the site in?

src/account/account.service.ts
```diff
@@ -37,8 +46,44 @@ export class AccountService {
...
+        if (account) {
+            return account;
+        }
+
+        const context = this.fedifyContextFactory.getFedifyContext();
+
+        const documentLoader = await context.getDocumentLoader({
+            handle: 'index',
+        });
+        const potentialActor = await lookupObject(id, { documentLoader });
```
mike182uk, line 69: 🤔 We have `FedifyActorResolver` thats used by the publishing service. Could we utilise the same thing here?

src/account/account.service.ts
```diff
@@ -37,8 +46,44 @@ export class AccountService {
...
+
+        if (!isActor(potentialActor)) {
+            throw new URL('Account not found');
+        }
+
+        const data = await mapActorToExternalAccountData(potentialActor);
+
+        await this.createExternalAccount(data);
+
+        const newlyCreatedAccount = await this.accountRepository.getByApId(id);
```
mike182uk, line 79: How come we can't use the result of `createExternalAccount`?

src/account/account.service.ts
```diff
@@ -11,23 +17,26 @@ import {
     TABLE_FOLLOWS,
     TABLE_USERS,
 } from '../constants';
+import type { Account } from './account.entity';
+import type { KnexAccountRepository } from './account.repository.knex';
 import type {
-    Account,
+    Account as AccountType,
```
mike182uk, line 23: This feels awkward, I think i'd have preferred aliasing the entity to to `AccountEntity`

Is this likely to be a temporary thing? Should `AccountType` and `Account` entity be the same thing?

src/post/post.repository.knex.ts
```diff
@@ -1,38 +1,155 @@
...
+
+                    await transaction('likes')
+                        .insert(likesToInsert)
+                        .onConflict()
+                        .ignore();
+
+                    const [[{ insertedRows }]] = await transaction.raw(
+                        'SELECT ROW_COUNT() as insertedRows',
+                    );
```
mike182uk, line 135: 🤔 Is this because you don't know if the like was inserted or not from the above query?

features/feed.feature
```diff
@@ -36,18 +36,18 @@ Feature: Feed
     And the feed does not contain "Article1"
 
   Scenario: Feed only includes posts
-    Given a "Create(Note)" Activity "Note1" by "Alice"
-    And "Alice" sends "Note1" to the Inbox
-    And "Note1" is in our Inbox
+    Given a "Create(Note)" Activity "CreateNote1(Note1)" by "Alice"
```
mike182uk, line 39: Im not grokking the verbose naming of the Activity?

src/account/account.service.ts
allouis: `createExternalAccount` is still using the old `Account` type which is just data and no functionality

src/account/account.service.ts
allouis: Yeah this is temporary, eventually `AccountType` will be completely replaced with `Account` 😬

src/post/post.repository.knex.ts
allouis: Yeah exactly - we use `ON CONFLICT IGNORE` so we don't know how many rows are inserted

features/feed.feature
allouis: So the problem here is that we weren't actually creating a `Like(Note)` we were creating a `Like(Create(Note))` - which allowed the tests to pass, but wasn't actually correct behaviour - so the verbose naming here is to allow us to refer to `Note1` as the object inside the activity `CreateNote1`

src/account/account.repository.knex.ts
allouis: It does leak Site a bit - but I think `Site` is intrinsically tied to `Account`, because an `Account` is internal or external based on a relation with a `Site` through the `users` table. If anything I would move `Site` closer into account here 🤔

src/account/account.service.ts
allouis: Yeah so this is annoying - the `lookupActor` method which the `FedifyActorResolver` uses is only available on a `FedifyRequestContext` - but we only have access to a `Fedify` context because the inbox handlers aren't run in the context of a request - they're run in the context of the queue

PR #321, opened 2025-02-17
Added to posts table when reply API action is executed

vershwal:
Refs https://linear.app/ghost/issue/AP-716/add-to-posts-table-when-reply-api-action-is-executed

### DIFF
--- a/src/handlers.ts
+++ b/src/handlers.ts
@@ -28,6 +28,7 @@ import { escapeHtml } from './helpers/html';
 import { getUserData } from './helpers/user';
 import { addToList, removeFromList } from './kv-helpers';
 import { lookupActor, lookupObject } from './lookup-helpers';
+import { Audience, Post, PostType } from './post/post.entity';
 import type { KnexPostRepository } from './post/post.repository.knex';
 import type { PostService } from './post/post.service';
 import type { SiteService } from './site/site.service';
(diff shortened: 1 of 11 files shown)


src/post/post.entity.ts
```diff
@@ -33,6 +33,13 @@ interface PostData {
     url?: URL | null;
     imageUrl?: URL | null;
     publishedAt?: Date;
+    likeCount?: number;
+    repostCount?: number;
+    replyCount?: number;
+    inReplyTo?: Post | null;
+    threadRoot?: Post | null;
+    readingTimeMinutes?: number | null;
+    apId?: URL | null;
```
vershwal, line 42: Updated this to reuse for replies.

src/post/post.repository.knex.ts
```diff
@@ -152,4 +154,92 @@ export class KnexPostRepository {
             throw err;
         }
     }
+
+    async updateReplyCount(postId: number, increment = true): Promise<void> {
```
vershwal, line 158: Added this param increment so that we can reuse this function fore delete later.

src/post/post.repository.knex.ts
```diff
@@ -152,4 +154,92 @@ export class KnexPostRepository {
...
+            row.author_id,
+            row.username,
+            row.name,
+            row.bio,
+            parseURL(row.avatar_url),
+            parseURL(row.banner_image_url),
+            null,
+        );
+
+        const post = new Post(
```
vershwal, line 222: I don't like doing this. Maybe there is a better way to map the row with post entity. 
We might not want to create a new Post object every time we write a function to fetch a row from posts table. This will be even more messy when we'll have to get multiple rows.

src/publishing/service.ts
```diff
@@ -192,7 +192,9 @@ export class FedifyPublishingService implements PublishingService {
         ];
 
         const fedifyNote = new FedifyNote({
-            id: this.uriBuilder.buildObjectUri(FedifyNote, uuidv4()),
+            id:
+                note.apId ||
+                this.uriBuilder.buildObjectUri(FedifyNote, uuidv4()),
```
vershwal, line 197: This change is to have the same apId for the note published and the note in our posts table.

src/handlers.ts
```diff
@@ -181,128 +184,159 @@ const ReplyActionSchema = z.object({
...
+        const parentPost = await postService.getByApId(objectToReplyTo.id!);
 
-    await addToList(ctx.get('db'), ['outbox'], create.id!.href);
+        const replyData = {
+            content: data.content,
+            type: PostType.Note,
+            audience: Audience.Public,
+            apId: replyId,
+            inReplyTo: parentPost,
+            threadRoot:
```
vershwal, line 298: The way I have implemented this is:-
1. ThreadRoot for a root post is NULL.
2. ThreadRoot for any reply is the ThreadRoot of it's parent. If the ThreadRoot of parent is NULL, parent is the root.

allouis: Nice, I like it

src/handlers.ts
```diff
@@ -181,128 +184,159 @@ const ReplyActionSchema = z.object({
...
-        id: createId,
-        actor: actor,
-        object: reply,
-        to: to,
-        ccs: cc,
-    });
+        await addToList(ctx.get('db'), ['outbox'], create.id!.href);
 
-    const activityJson = await create.toJsonLd();
+        const account = await accountService.getByApId(actor?.id!);
```
allouis, line 288: Let's avoid using `?` and `!` here and instead check that these properties exist before hand - if they don't then we probably don't want to handle this.

allouis: Although I just realised that `actor` here is referring to the actor of the user doing the reply - in this case we can use the `getDefaultAccount` method instead and forget about the `actor` completely

src/handlers.ts
```diff
@@ -181,128 +184,159 @@ const ReplyActionSchema = z.object({
...
-        { handle: ACTOR_DEFAULT_HANDLE },
-        attributionActor,
-        create,
-        {
-            preferSharedInbox: true,
-        },
-    );
+        const newReply = Post.createFromData(account, replyData);
+        await postRepository.save(newReply);
+        await postRepository.updateReplyCount(parentPost.id!, true);
```
allouis, line 307: This is putting the logic of updating the reply count in the service - which we don't want! It could be possible that somewhere else in the app we create/save replies and forget to call `updateReplyCount`

There are a few other options we can do to make sure we don't run into this problem:

We could have something in the post repository implementation check for an `inReplyTo` and if there is one, it will handle updating the count for the row - this makes sure that we can't forget to update it - but it also means that the business logic is in the repository rather than the domain (maybe fine for something like this)

We could also model it in the domain layer with something like:
```ts
parentPost.addReply(newReply);

await postRepository.save(parentPost);
```

Or even
```ts
const newReply = parentPost.createReply(account, replyData);

await postRepository.save(parentPost, newReply);
```

Both of these means that the logic happens in the domain - and so should be unit testable too!

They're just ideas, there's a million ways to do it - we can catch up in slack if you wanna go through it more

src/post/post.repository.knex.ts
```diff
@@ -361,6 +431,74 @@ export class KnexPostRepository {
...
+            published_at: reply.publishedAt,
+            in_reply_to: reply.inReplyTo?.id,
+            thread_root: reply.threadRoot?.id,
+            ap_id: reply.apId.href,
+        }));
+
+        await transaction(TABLE_POSTS)
+            .insert(repliesToInsert)
+            .onConflict()
+            .ignore();
```
vershwal, line 493: We have a unique check on ap_id_hash so this will reject the duplicate replies

src/post/post.repository.knex.ts
```diff
@@ -19,6 +20,39 @@ export class KnexPostRepository {
...
+            row.title,
+            row.excerpt,
+            row.content,
+            new URL(row.url),
+            parseURL(row.image_url),
+            new Date(row.published_at),
+            row.like_count,
+            row.repost_count,
+            row.reply_count,
+            row.in_reply_to,
```
allouis, line 49: This is required to be of type `Post | null` - and same for the `threadRoot` param, but here it will be a number.

I'm not sure if we want to eagerly load the parent posts, because we could end up with a lot of data (and queries) when fetching posts which are deep in a chain of replies 🤔 

I think it would be better to have some methods either on the repository or the service which can fetch the parent, root or entire chain for posts 🤔 

This is a fairly big change so I'm gonna take a look at them as you're off today

vershwal: Thanks, I think it will be good to have those methods. I needed that to set the threadRoot [here](https://github.com/TryGhost/ActivityPub/pull/321/files#diff-c77b89ae7ba2627f28f4b8fe20ecbb5c169809193c3a3cfd404cd93f12750b83R307) but I ended up making two queries for it.

src/post/post.repository.knex.ts
```diff
@@ -430,99 +389,4 @@ export class KnexPostRepository {
...
-
-    /**
-     * Increment a count stored on a post by a given amount
-     *
-     * @param post Post to increment the count of
-     * @param countType Type of count to increment
-     * @param incrementBy Amount to increment the count by
-     * @param transaction Database transaction to use
-     */
-    private async incrementCount(
```
allouis, line 372: This was only used in one place for each update type - I felt it is much easier to read by just inlining the call

mike182uk: You can remove `PostCountType` as it should be unused now

PR #325, opened 2025-02-19
Added reposts to db when repost API action is executed

vershwal:
Refs https://linear.app/ghost/issue/AP-726/add-to-reposts-table-when-repost-api-action-is-executed

### DIFF
--- a/src/handlers.ts
+++ b/src/handlers.ts
@@ -16,6 +16,7 @@ import type { Context } from 'hono';
 import { v4 as uuidv4 } from 'uuid';
 import z from 'zod';
 
+import type { KnexAccountRepository } from './account/account.repository.knex';
 import type { AccountService } from './account/account.service';
 import { mapActorToExternalAccountData } from './account/utils';
 import { type HonoContextVariables, fedify } from './app';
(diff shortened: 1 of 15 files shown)


src/handlers.ts
```diff
@@ -611,6 +617,23 @@ export function createRepostActionHandler(accountService: AccountService) {
...
+            ctx.get('logger').info('Invalid Repost - no actor id');
+            return;
+        }
+
+        if (!post.id) {
+            ctx.get('logger').info('Invalid Repost - no post id');
+            return;
+        }
+
+        const account = await accountService.getByApId(actor.id);
```
allouis, line 630: I think we should use `accountService.getDefaultAccountForSite` (or however it's called) here, rather than depending on the `actor` - because eventually the `actor` won't be present in this method

allouis (approved):
Very small comment - looks great, nice and small!

src/account/utils.ts
```diff
@@ -77,3 +79,21 @@ export async function mapActorToExternalAccountData(
 export function getAccountHandle(host?: string, username?: string) {
     return `@${username || 'unknown'}@${host || 'unknown'}`;
 }
+
+/**
+ * Converts Account interface to Account entity
+ */
+export function mapToAccountEntity(
```
vershwal, line 86: I'll rename this to `mapAccountTypeToAccountEntity` for better clarity.

src/handlers.ts
vershwal: Done

PR #333, opened 2025-02-20
Checked yarn.lock integrity during CI

allouis:
We want to make sure that we don't add changes to the package.json which are not reflected correctly in the yarn.lock file.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -21,6 +21,21 @@ jobs:
       - name: Run Biome
         run: biome ci .
 
+  check-yarn-lock:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v4
+
+      - name: Set up Node.js
+        uses: actions/setup-node@v4
+        with:
+          node-version: 'lts/*'
+          cache: 'yarn'
+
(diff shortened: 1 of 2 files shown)


.github/workflows/build.yml
```diff
@@ -21,6 +21,25 @@ jobs:
       - name: Run Biome
         run: biome ci .
 
+  check-yarn-lock:
```
allouis, line 24: @daniellockyer I thought you might be interested in this, and wondered if you have any feedback on the approach

.github/workflows/build.yml
```diff
@@ -21,6 +21,25 @@ jobs:
...
+        with:
+          node-version: 'lts/*'
+          cache: 'yarn'
+
+      - name: Install dependencies
+        run: yarn
+
+      - name: Check for yarn.lock changes
+        run: |
+          git diff --exit-code yarn.lock || (echo "yarn.lock has changed. Commit the updated file." && exit 1)
```
daniellockyer, line 41: This is probably OK but there's also https://classic.yarnpkg.com/en/docs/cli/install#toc-yarn-install-frozen-lockfile, which is the built-in way of doing it. Could be worth using that

PR #336, opened 2025-02-24
 Wired up unlike action to the new db tables

allouis:
Before we wire anything up to the unlike action, we need to addthe capability
of unliking to our model. Now that we have likes and unlikes we need to make
sure that they stay in sync, so we remove from the opposite list when
performing an action, to cancel out any opposing actions.

With this approach we can no longer have a one-sided get method, because it
would then be possible for things to fall out of sync, so we move this to a
getChangedLikes method, which will read from and clear both lists. It is still
up to the repository to make sure it actually deals with these correctly ofc.

We can safely ignore removing likes in the case of a new post because it's
impossible for there to already be rows. Knex will return the deleted number
from a call to `del` so we're able to so a single update for the count, taking
into account the inserted and deleted rows.

For outgoing unlikes, we want to remove the row from our `likes` table
immediately, this is so that we can track the removal of likes of posts outside
of our system, as well as not relying on the incoming activity to remove likes
for posts inside our system.


src/post/post.entity.ts
```diff
@@ -106,9 +107,29 @@ export class Post extends BaseEntity {
...
+        }
+        this.potentiallyNewLikes.delete(account.id);
+        this.likesToRemove.add(account.id);
+    }
+
+    getChangedLikes() {
+        const likesToRemove = [...this.likesToRemove.values()];
+        this.likesToRemove.clear();
+        const likesToAdd = [...this.potentiallyNewLikes.values()];
+        this.potentiallyNewLikes.clear();
```
vershwal, line 126: We can rename `potentiallyNewLikes` to `likesToAdd` as we did in [post.repository.knex.ts](https://github.com/TryGhost/ActivityPub/pull/336/files?diff=split&w=1#diff-80d34d2729ea70828740b1b3a8b31097755c1c8c0d4443102b68ce637814ec6bR103), to maintain consistent naming.

src/post/post.entity.ts
```diff
@@ -106,9 +107,29 @@ export class Post extends BaseEntity {
...
+        return {
+            likesToRemove,
+            likesToAdd,
+        };
+    }
+
     getPotentiallyNewLikes() {
```
vershwal, line 133: Looks like this function is not used anymore. Let's remove this?

src/post/post.entity.ts
allouis: Yeah I agree, I'll do this in a cleanup commit

src/post/post.entity.ts
allouis: Will do this with the cleanup commit!

mike182uk (approved):

PR #338, opened 2025-02-24
Wired up derepost action to the new db tables

vershwal:
Ref https://linear.app/ghost/issue/AP-733/remove-from-reposts-table-when-derepost-api-action-is-executed

### DIFF
--- a/src/handlers.ts
+++ b/src/handlers.ts
@@ -789,7 +789,11 @@ export function createRepostActionHandler(
     };
 }
 
-export function createDerepostActionHandler(accountService: AccountService) {
+export function createDerepostActionHandler(
+    accountRepository: KnexAccountRepository,
+    postService: PostService,
+    postRepository: KnexPostRepository,
+) {
     return async function derepostAction(
         ctx: Context<{ Variables: HonoContextVariables }>,
     ) {
@@ -828,6 +832,21 @@ export function createDerepostActionHandler(accountService: AccountService) {
         const announceToUndo = await Announce.fromJsonLd(announceToUndoJson);
 
         const actor = await apCtx.getActor(ACTOR_DEFAULT_HANDLE); // TODO This should be the actor making the request
+
+        if (!post.id) {
+            ctx.get('logger').info('Invalid Derepost - no post id');
+            return;
+        }
+
+        const account = await accountRepository.getBySite(ctx.get('site'));
+        if (account !== null) {
+            const originalPost = await postService.getByApId(post.id);
+
+            if (originalPost) {
+                originalPost.removeRepost(account);
+                await postRepository.save(originalPost);
+            }
+        }
 
         const undo = new Undo({
             id: undoId,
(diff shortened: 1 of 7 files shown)


src/post/post.repository.knex.ts
```diff
@@ -155,21 +151,34 @@ export class KnexPostRepository {
...
+                    const removedRepostsCount =
+                        repostsToRemove.length > 0
+                            ? await this.removeReposts(
+                                  post,
+                                  repostsToRemove,
+                                  transaction,
+                              )
+                            : 0;
+
+                    repostAccountIds = accountIdsInserted.filter(
```
vershwal, line 173: We use these repostAccountIds to emit `PostRepostedEvent`, do we also need something like `PostDeRepostedEvent`?

mike182uk: Yeah i think so, although I think that falls under the umbrella of deleting content 🤔 (as we would need to remove the `post` from the `feed` on receipt of this event). Not sure if we want to do this now or later?

vershwal: Sorry I deleted my earlier comment(before I could see your reply😬) as I have already added the PostDerepostedEvent here: https://github.com/TryGhost/ActivityPub/pull/338/commits/6c604e4ad1f7b293605004c60e00d6c22126a207. 
I think there should be no harm in emitting the event and we can use it later while implementing delete.

allouis: I think we do want to remove the repost from feeds as part of this work - @mike182uk how much overlap is that with your stuff?

src/handlers.ts
```diff
@@ -764,6 +768,21 @@ export function createDerepostActionHandler(accountService: AccountService) {
 
         const actor = await apCtx.getActor(ACTOR_DEFAULT_HANDLE); // TODO This should be the actor making the request
 
+        if (!post.id) {
+            ctx.get('logger').info('Invalid Derepost - no post id');
+            return;
+        }
+
+        const account = await accountRepository.getBySite(ctx.get('site'));
+        if (account !== null) {
```
allouis, line 777: This will always be true because `accountRepository.getBySite` returns `Promise<Account>` not `Promise<Account | null>`

src/handlers.ts
```diff
@@ -764,6 +768,21 @@ export function createDerepostActionHandler(accountService: AccountService) {
 
         const actor = await apCtx.getActor(ACTOR_DEFAULT_HANDLE); // TODO This should be the actor making the request
 
+        if (!post.id) {
```
allouis, line 771: I think it's slightly better to use the `id` parsed in from params here, because when we eventually pull the fedify stuff out of this handler, there will be no `post` instance to use the `id` of!

It's not a big deal, can always fix it then

src/post/post.entity.ts
```diff
@@ -119,13 +120,27 @@ export class Post extends BaseEntity {
...
+        this.repostsToAdd.add(account.id);
     }
 
-    getPotentiallyNewReposts() {
-        const reposts = [...this.potentiallyNewReposts.values()];
-        this.potentiallyNewReposts.clear();
-        return reposts;
+    removeRepost(account: Account) {
+        if (!account.id) {
+            throw new Error('Cannot add repost for account with no id');
```
allouis, line 129: Wording needs updating here

allouis (commented):
A few minor comments, but I think we do need to wire this up to the FeedService so that we remove from feeds.

PR #340, opened 2025-02-25
Added changes to save and retrieve attachments of a post

vershwal:
Refs https://linear.app/ghost/issue/AP-781/store-attachments-for-posts


src/post/post.service.ts
```diff
@@ -11,6 +11,38 @@ export class PostService {
...
 
+    /**
+     * Get the attachments for a post
+     *
+     * @param attachments
+     */
+    private async getPostAttachments(
+        foundObject: Note | Article,
+    ): Promise<PostAttachment[] | null> {
+        const attachments = foundObject.getAttachments();
```
allouis, line 22: This is a different approach to how we currently get attachments here https://github.com/TryGhost/ActivityPub/blob/b62bcefd84459fe775f911ad514738c382ae6cfd/src/http/api/helpers/post.ts#L167-L186

I'm not sure which one is correct, but I'm interested to know what you think

allouis (commented):
Left a couple of comments

src/post/post.service.ts
vershwal: Yes, I had to change the approach a bit. 

In [ActivityPub/src/http/api/helpers/post.ts](https://github.com/TryGhost/ActivityPub/blob/b62bcefd84459fe775f911ad514738c382ae6cfd/src/http/api/helpers/post.ts#L167-L186), we had activity.object of type [ActivityObject](https://github.com/TryGhost/ActivityPub/blob/b62bcefd84459fe775f911ad514738c382ae6cfd/src/helpers/activitypub/activity.ts#L16), where attachments are directly accessible. So we just did `activity.object.attachment` [to get the attachments](https://github.com/TryGhost/ActivityPub/blob/b62bcefd84459fe775f911ad514738c382ae6cfd/src/http/api/helpers/post.ts#L173). 

Here we have foundObject of type Note/Article, where attachments are accessible through `getAttachments()` function(present in @fedify/fedify/types/vocab/vocab.d.ts)

```
/** Identifies a resource attached or related to an object that potentially
     * requires special handling.  The intent is to provide a model that is at
     * least semantically similar to attachments in email.
     */
    getAttachments(options?: {
        documentLoader?: DocumentLoader;
        contextLoader?: DocumentLoader;
        suppressError?: boolean;
        tracerProvider?: TracerProvider;
    }): AsyncIterable<Object | Link | PropertyValue>;
```

I wasn't sure how I can map the Note/Article to Activity, or if it's the right thing to do?

I have also seen that the approach I have used here is also used to get attachments for actor. [Here](https://github.com/TryGhost/ActivityPub/blob/b62bcefd84459fe775f911ad514738c382ae6cfd/src/helpers/activitypub/actor.ts#L29) and [here](https://github.com/TryGhost/ActivityPub/blob/b62bcefd84459fe775f911ad514738c382ae6cfd/src/account/utils.ts#L20).

allouis: Okay that makes sense! @mike182uk Do you have any opinions on this? I just got a bit confused with the different types

mike182uk: Ah, so in my version (the one you've linked), the activity is coming from the database (which means we have previously serialised it for storage, not sure to if `toJsonLd` implicitly calls `getAttachments`)

In this version here, the activity has come straight from Fedify and is not serialised, so i'm guessing `object.attachments` isn't available, which means `getAttachments` is probably the correct way to go

src/post/post.entity.ts
```diff
@@ -36,6 +43,7 @@ export interface PostData {
     publishedAt?: Date;
     inReplyTo?: Post | null;
     apId?: URL | null;
+    attachments?: PostAttachment[] | null;
```
mike182uk, line 46: 🤔 Do we need `null` for this? Like could it not always just be `PostAttachment[]` and if there are no attachments its an empty array? Then when we need to check if there are attachments we just do `attachments.length > 0`? Would mean we don't have to explicitly check if the value is `null` before trying to check `length`

vershwal: Good point. I'll update it!

src/post/post.entity.ts
```diff
@@ -81,6 +90,7 @@ export class Post extends BaseEntity {
         public readonly inReplyTo: number | null = null,
         public readonly threadRoot: number | null = null,
         private readonly readingTimeMinutes: number | null = null,
+        attachments: PostAttachment[] | null = null,
```
allouis, line 93: If instead we do `public readonly attachments: PostAttachment[]` 

Then I think we save a bunch of code, we don't need the public member on the class above on line 73, and we don't need the assignment on line 112. The repository is already doing the mapping anyway, so we don't need to do it here!

That also means that `getPostAttachments` can return `Promise<PostAttachment[]>` and simplifies the implementation there, too!

allouis (approved):
I left one comment, but it's not a big deal so up to you if you address it ☺️

PR #349, opened 2025-02-26
Reading attachments from post table

vershwal:


src/http/api/types.ts
```diff
@@ -119,24 +119,7 @@ export interface PostDTO {
...
-        /**
-         * Name of the attachment
-         */
-        name: string;
-        /**
-         * URL of the attachment
-         */
-        url: string;
-    }[];
+    attachments: PostAttachment[];
```
mike182uk, line 122: I think we should leave this as it was so that the API response is not dependent on the Entity representation (this means we can be flexible about whats returned where)

src/http/api/feed.ts
```diff
@@ -71,7 +71,9 @@ export function createGetFeedHandler(
                 likedByMe: result.post_liked_by_user === 1,
                 replyCount: result.post_reply_count,
                 readingTimeMinutes: result.post_reading_time_minutes,
-                attachments: [], // TODO
+                attachments: result.post_attachments
+                    ? result.post_attachments
+                    : [],
```
mike182uk, line 76: Rather than trusting that whats in the field is correct, i think we should explicitly map out the values so we can guarantee we are returning the correct thing in the correct structure

vershwal:
Hey @mike182uk , I have updated this PR, could you please give it a quick look again?

src/http/api/feed.ts
```diff
@@ -71,7 +71,14 @@ export function createGetFeedHandler(
                 likedByMe: result.post_liked_by_user === 1,
                 replyCount: result.post_reply_count,
                 readingTimeMinutes: result.post_reading_time_minutes,
-                attachments: [], // TODO
+                attachments: result.post_attachments
+                    ? result.post_attachments.map((attachment) => ({
+                          type: attachment.type ?? '',
+                          mediaType: attachment.mediaType ?? '',
+                          name: attachment.name ?? '',
+                          url: attachment.url,
```
mike182uk, line 79: do we want to `?? ''` on this aswell?

mike182uk (approved):

src/http/api/feed.ts
allouis: Does an attachment without a URL make sense? Should the entity not even allow those?

vershwal: Yeah, attachment.url is a non null field. And we don't need `?? ''` as an attachment without a URL will not make sense.

PR #358, opened 2025-03-03
Implemented removing deleted posts from feeds

mike182uk:
refs [AP-802](https://linear.app/ghost/issue/AP-802/remove-deleted-posts-from-feeds)

Implemented removing deleted posts from feeds


allouis (commented):
Do we wanna to run the `select` and `delete` in a transaction? Otherwise looks 👍

mike182uk:
> Do we wanna to run the `select` and `delete` in a transaction? Otherwise looks 👍

I thought about it, but I don't think there is a need - If the delete query fails, an error will be thrown before the return so `updatedFeedUserIds` won't be returned

PR #363, opened 2025-03-04
Support persistence of deleted posts

allouis:
ref https://linear.app/ghost/issue/AP-812

We need to mark deleted posts as deleted in the DB, as well as emit a PostDeletedEvent so that other systems can react to it, for example removing the post from the feed after a deletion.

Newly created posts that are immediately deleted are not a use-case we have and it's not clear on how that should work, this implementation simply does not persist them for now, but we can change that in future if we need to support this.

When a post is deleted, the only column we write to is the deleted_at column, this ensures that the content is still available for audit, debug, undoing etc...

We also have to add a new check that the Post author Account has an id property, I would like to move this constraint to the entity layer, but that is outside the scope of this PR. Right now this check still happens, but it is implicit at the DB level with a foreign key constraint on the column - this just makes it explicit so that 1) we are all aware of it and 2) TypeScript doesn't complain about referencing author.id when instantiating the PostDeletedEvent.

### DIFF
--- a/src/post/post.repository.knex.ts
+++ b/src/post/post.repository.knex.ts
@@ -6,6 +6,7 @@ import { Account } from '../account/account.entity';
 import { TABLE_LIKES, TABLE_POSTS, TABLE_REPOSTS } from '../constants';
 import { parseURL } from '../core/url';
 import { PostCreatedEvent } from './post-created.event';
+import { PostDeletedEvent } from './post-deleted.event';
 import { PostDerepostedEvent } from './post-dereposted.event';
 import { PostRepostedEvent } from './post-reposted.event';
 import { Post } from './post.entity';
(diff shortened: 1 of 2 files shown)


src/post/post.repository.knex.ts
```diff
@@ -379,6 +392,12 @@ export class KnexPostRepository {
                         (accountId) => accountId,
                     );
                 }
+            } else if (isDeletedPost) {
+                await transaction('posts')
```
mike182uk, line 396: We discussed this on the call (i think), but we also want to decrement the reply count on `inReplyTo` if it has one

src/post/post.repository.knex.ts
```diff
@@ -379,6 +393,31 @@ export class KnexPostRepository {
...
+                        })
+                        .where({ id: post.id });
+
+                    if (post.inReplyTo) {
+                        await transaction('posts')
+                            .update({
+                                reply_count: this.db.raw('reply_count - 1'),
+                            })
+                            .where({ id: post.inReplyTo });
+                    }
```
mike182uk, line 417: Think we need to cover this in the existing test or a new test

src/post/post.repository.knex.ts
```diff
@@ -379,6 +393,31 @@ export class KnexPostRepository {
...
+                    .select('deleted_at')
+                    .where({
+                        id: post.id,
+                    })
+                    .first();
+
+                if (existingRow && existingRow.deleted_at === null) {
+                    await transaction('posts')
+                        .update({
+                            deleted_at: transaction.raw('CURRENT_TIMESTAMP'),
```
mike182uk, line 407: Based on what was discussed in Slack, do we want to use `UTC_TIMESTAMP` here now, or do we want to do a big bang change later?

mike182uk (approved):

src/post/post.repository.knex.ts
allouis: Yeah so I wanted to do a big bang change so that we can do any migrations if necessary and all the current data is consistent

PR #365, opened 2025-03-05
Improve tests

allouis:
### [Made all test:* scripts runnable from the host](https://github.com/TryGhost/ActivityPub/commit/c1287513cf4aae95dbb0f97899dccce61515252d) 

Some of our test scripts required to be run inside of docker which
makes working with them locally really annoying.

This standardises all of our test scripts using the `_test` prefix for
the command which executes the test and the `test` prefix for the
command which can run the test from the host.

This means we can use `yarn test:unit` or `yarn test:integration` and
they will run in the docker environment with access to everything they need.

We are now running the unit tests inside of docker just to be
consistent, I haven't seen a noticable impact on the speed here - but
it can always be optimised in future if desired.

### [Supported running the test suite in parallel](https://github.com/TryGhost/ActivityPub/commit/9559ac60ec1677d28ed0d324b47785c0e1a268de) 

Our integration tests were all using the same database, which meant
that they would interfere with one another and cause random failures,
we initially fixed this by forcing the tests to run in series, but as
our suite has grown this has caused it to slow down considerably.

This introduces a new test helper `createTestDb` which will create a
brand new database with the same schema as the activitypub one, so
that test suites can each create their own isolated environment.

We've had to use the `root` user for MySQL in our testing container so
that it has permission to create new databases.

### [Switched to using the dot reporter for tests](https://github.com/TryGhost/ActivityPub/commit/5d06829b583ac65a3b15fe5bf8344a4c99e652b0) 

The default reporter is quite noisy and we're using the dot report for
cucumber so this felt like a natural change to make.


allouis:
This halves the time to run the vitest tests on CI

### [Before:](https://github.com/TryGhost/ActivityPub/actions/runs/13654864657/job/38171463841)
```
 Test Files  30 passed (30)
      Tests  169 passed (169)
   Start at  13:34:27
   Duration  60.40s (transform 610ms, setup 0ms, collect 10.31s, tests 42.12s, environment 16ms, prepare 2.35s)
```
### [After:](https://github.com/TryGhost/ActivityPub/actions/runs/13668748633/job/38214899194?pr=365)
```
 Test Files  30 passed (30)
      Tests  169 passed (169)
   Start at  04:52:17
   Duration  30.08s (transform 1.07s, setup 0ms, collect 15.57s, tests 42.22s, environment 24ms, prepare 3.42s)
```

PR #367, opened 2025-03-06
Updated key value stores when post is deleted

vershwal:


allouis (approved):

src/http/api/post.ts
```diff
@@ -43,9 +46,25 @@ export function createDeletePostHandler(
...
 
         try {
+            // Delete the post from the database
             post.delete(account);
             await postRepository.save(post);
 
+            // Find all activities that reference this post and remove them from the kv-store
+            const relatedActivities = await getRelatedActivities(idAsUrl.href);
+
+            const activities = (await relatedActivities) as any[];
```
allouis, line 56: We should try not to use `any` here if we can, I think you can return `Promise<{id: string}[]>` from the `getRelatedActivities` function

PR #369, opened 2025-03-06
Handling incoming delete acticities

vershwal:
Ref https://linear.app/ghost/issue/AP-787/[feature]-when-someone-i-follow-deletes-a-post-it-gets-removed-from-my

- When someone I follow deletes a post, my server receives a Delete Activity and deletes it from the database
- The post should then not appear in my Inbox, Feed, or when visiting their Profile


features/handle-delete.feature
```diff
@@ -0,0 +1,36 @@
...
+  We want to handle Delete(Note) activities in the Inbox
+
+  Background:
+    Given an Actor "Person(Bob)"
+    And we follow "Bob"
+    And the request is accepted
+    And a "Accept(Follow(Bob))" Activity "Accept" by "Bob"
+    And "Bob" sends "Accept" to the Inbox
+    And "Accept" is in our Inbox
+    Given an Actor "Person(Alice)"
```
mike182uk, line 11: Remember we ideally only want 1 `Given`, `When` and `Then` then use `And` to link together the steps. I know that then means theres a lot of `Ands` here

src/activity-handlers/delete.handler.ts
```diff
@@ -0,0 +1,53 @@
...
+
+        if (post === null) {
+            ctx.data.logger.info('Post not found, exit early');
+            return;
+        }
+
+        post.delete(senderAccount);
+        await this.postRepository.save(post);
+
+        return;
```
mike182uk, line 51: no need for return here

src/activity-handlers/delete.handler.ts
```diff
@@ -0,0 +1,53 @@
...
+
+        const post = await this.postService.getByApId(deleteActivity.objectId);
+
+        if (post === null) {
+            ctx.data.logger.info('Post not found, exit early');
+            return;
+        }
+
+        post.delete(senderAccount);
+        await this.postRepository.save(post);
```
mike182uk, line 49: Do we also need to do a similar thing that was done with an outgoing delete with regards to removing from `likes`, `reposts` etc?

mike182uk: This would be for the user receiving the activity

And we would also need to take into account the `likes` and `reposts` db tables?

mike182uk (approved):

PR #370, opened 2025-03-06
Send Delete activity to followers when we delete a post

allouis:
Had to make some tweaks to the Account entity and how we load it
in the first two commits - making sure we have access to the 
apFollowers URL as well as the isInternal flag - so we can send the
activities correctly.

I've used the same pattern as the existing object dispatchers here
to keep things simple and read from the global DB, but I have moved
the dispatcher into its own file

Finally we wire up the event to the fediverse bridge, which handles the
sending of the activities


features/step_definitions/stepdefs.js
```diff
@@ -1518,12 +1518,18 @@ Then(
...
-                    (object
-                        ? body.object.id === object.id
-                        : body.object.type === objectNameOrType)
-                );
+                if (body.type !== activityType) {
+                    return false;
+                }
+
+                if (object) {
+                    if (typeof body.object === 'string') {
```
allouis, line 1526: This is because `Delete` activities just use an id for the `object` rather than an expanded object (because it's been deleted)

features/step_definitions/stepdefs.js
```diff
@@ -1579,6 +1585,7 @@ When(
             const activity = await this.response.clone().json();
 
             this.activities[noteName] = activity;
+            this.objects[noteName] = activity.object;
```
allouis, line 1588: We only had access to the `Create(Note)` activity but we needed the underlying object for checking the `Delete(Note)` activity.

Maybe we wanna fix these up eventually so they have different names? Not sure

mike182uk: Yeh i ended up doing this same thing in some other step

src/account/account.entity.ts
```diff
@@ -70,6 +78,17 @@ export class Account extends BaseEntity {
         );
     }
 
+    getApFollowers() {
```
mike182uk, line 81: 🤔 should we be more explicit with the name? `getApFollowersUrl`?

mike182uk (approved):

src/account/account.entity.ts
allouis: IMO The `URL` type communicates that it is a url

PR #375, opened 2025-03-10
Handling Post/Account not found error for incoming delete

vershwal:


src/activity-handlers/delete.handler.ts
```diff
@@ -33,14 +35,28 @@ export class DeleteHandler {
             return;
         }
 
-        const senderAccount = await this.accountService.getByApId(sender.id);
+        let senderAccount: Account | null = null;
+
+        try {
+            senderAccount = await this.accountService.getByApId(sender.id);
+        } catch (error) {
+            ctx.data.logger.info('Error fetching sender account', { error });
```
vershwal, line 43: Not sure if this is required, but account service throws an error [here](https://github.com/TryGhost/ActivityPub/blob/main/src/account/account.service.ts#L75) so maybe good to have?

allouis (approved):
Looks good, just one comment about the logging

PR #378, opened 2025-03-10
Moved content preparation to the Post entity

allouis:
ref https://linear.app/ghost/issue/AP-869

Our replies were not going through the publishing service which meant that content preparation was not being applied to them. We have instead moved this logic into the entity. And created two new static methods for creating Notes and Replies for our internal accounts. This allows us to contain all of the business logic in the enitity and ensure that it is applied consistently

We have also updated the reply action handler to use the `content` and `apId` from the Post entity, in the same way that the note action does. This delegates the business logic to our entity layer, keeping everything in one place.


features/create-note.feature
```diff
@@ -24,6 +24,15 @@ Feature: Creating a note
     Then "Note" is in our Outbox
     And "Note" has the content "<p>Hello<br />World</p>"
 
+  Scenario: Created note contains newlines
```
mike182uk, line 27: Isn't this the same as the scenario above?

mike182uk (approved):

features/create-note.feature
allouis: Yes - I added this to the `create-reply.feature` tests instead

PR #380, opened 2025-03-11
Ensured the sanitisation of content

allouis:
ref https://linear.app/ghost/issue/AP-870

We want to be sure that we're sanitising content when we read/write to the db through the entity layer. But in the same pattern as the activities endpoint we are also adding sanitisation to the feed service to ensure that any issue with the storing of html content is still protected when reading data for the feed.


features/feed.feature
```diff
@@ -120,3 +120,12 @@ Feature: Feed
     And post "1" in the feed is "Article2"
     And post "2" in the feed is "Article1"
     And the feed does not contain "Note1"
+
+  Scenario: Feed is sanitised
```
mike182uk, line 124: Do we also want to add a test case for a reply thread? (although functionality should be covered by the post entity)

mike182uk (approved):

PR #391, opened 2025-03-13
Enabled ActivityPub to be run as just a MessageQueue handler

allouis:
We move the `/mq` endpoint a lot earlier in the HTTP stack so we skip unnecessary middleware like auth.

We also add a new `MANUALLY_START_QUEUE` option for the env, when this is set we will force Fedify to start the queue when the application boots.

The `startQueue` method requires that we pass a valid `ContextData` to it, and we've had to pass a KV store scoped to `"UNUSED_HOST"` in this case. However this store is never used, due to the `ensureCorrectContext` function which wraps all `fedify` handlers. 

This function will override the `db` property on the context data, scoping the store to the `host`. This `host` is captured in the message on the queue and will be set by Fedify on the context

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -126,7 +126,15 @@ jobs:
           job: migrations
           flags: '--wait --execute-now --set-cloudsql-instances=ghost-activitypub:europe-west4:activitypub-db'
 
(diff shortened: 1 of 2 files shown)


rmgpinto (approved):
Tested right now, working as intended. Thank you @allouis!

PR #396, opened 2025-03-18
Removed pagination support from followers

allouis:
ref https://linear.app/ghost/issue/AP-871

There seems to be a problem with how Fedify handles Followers Sync for paginated followers dispatchers which don't support filtering. We have seen some loss of followers and think it might be linked. We're going to disable pagination for now to see if it fixes the problem.

### DIFF
--- a/src/activitypub/followers.service.ts
+++ b/src/activitypub/followers.service.ts
@@ -1,0 +1,47 @@
+import type { Recipient } from '@fedify/fedify';
+import type { Account } from 'account/account.entity';
+import { parseURL } from 'core/url';
+import type { Knex } from 'knex';
+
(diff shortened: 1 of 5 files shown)


src/activitypub/followers.service.ts
```diff
@@ -0,0 +1,47 @@
+import type { Recipient } from '@fedify/fedify';
+import type { Account } from 'account/account.entity';
+import { parseURL } from 'core/url';
+import type { Knex } from 'knex';
+
+export class FollowersService {
```
vershwal, line 6: We can add some tests for this new file.

vershwal (approved):
LGTM!

PR #400, opened 2025-03-18
Writing to new `published_at` column in `feeds` table

vershwal:


src/feed/feed.service.ts
```diff
@@ -272,6 +280,8 @@ export class FeedService {
 
         const feedEntries = userIds.map((userId) => ({
             post_type: post.type,
+            published_at:
+                repostedBy && repost ? repost.created_at : post.publishedAt,
```
allouis, line 284: Do we need to check the type of `repost.created_at` here?

allouis (commented):
I think this could do with some integration tests added

allouis (approved):

PR #405, opened 2025-03-18
Added notifications query perf testing

mike182uk:
refs https://linear.app/ghost/issue/AP-897

Added data generation scripts for notifications and a query to read notifications


mike182uk:
This is in an unfinished state and we don't feel its worth pursing further, closing for now

PR #410, opened 2025-03-19
Replaced Post object with postId in PostCreatedEvent

allouis:
ref https://linear.app/ghost/issue/AP-880

Changed PostCreatedEvent to use postId instead of the full post object to reduce event payload size. This optimization prepares for future PubSub integration where message size constraints are important.


allouis:
Closing this in favour of serialising the events at a higher level

PR #419, opened 2025-03-20
Added getById method to KnexPostRepository

allouis:
ref https://linear.app/ghost/issue/AP-880

When sending messages on PubSub we want to keep size to a minimum, to achieve this we'll use id's instead of values and then fetch the values from the DB.

### DIFF
--- a/src/post/post.repository.knex.ts
+++ b/src/post/post.repository.knex.ts
@@ -23,6 +23,118 @@ export class KnexPostRepository {
         private readonly events: AsyncEvents,
     ) {}
 
+    async getById(id: Post['id']): Promise<Post | null> {
+        const row = await this.db(TABLE_POSTS)
+            .join('accounts', 'accounts.id', 'posts.author_id')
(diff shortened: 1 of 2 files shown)


src/post/post.repository.knex.ts
```diff
@@ -23,6 +23,118 @@ export class KnexPostRepository {
         private readonly events: AsyncEvents,
     ) {}
 
+    async getById(id: Post['id']): Promise<Post | null> {
```
vershwal, line 26: This function looks pretty similar to `getByApId`, and there’s quite a bit of duplicated code. Might be a good idea to refactor some of the shared parts into common utility methods. We an also reuse them in other get queries later on.

vershwal (approved):
Looks good. Just a minor comment about refactoring.

src/post/post.repository.knex.ts
allouis: That's a good point, thanks for pointing it out - I'll clean it up before merging!

PR #430, opened 2025-03-24
Handling non paginated followers in profile endpoint

vershwal:
Ref https://linear.app/ghost/issue/AP-956, https://linear.app/ghost/issue/AP-959

- Changed profile API endpoint to handle non-paginated followers list

### DIFF
--- a/src/http/api/profile.ts
+++ b/src/http/api/profile.ts
@@ -3,8 +3,7 @@ import {
     Activity,
     type Actor,
     Announce,
-    type Collection,
-    type CollectionPage,
+    CollectionPage,
     Like,
     isActor,
 } from '@fedify/fedify';
@@ -201,9 +200,8 @@ export function createGetProfilePostsHandler(accountService: AccountService) {
                     next,
                 )) as CollectionPage | null;
 
-                // Explicitly check that we have a valid page seeming though we
-                // can't be type safe due to lookupObject returning a generic object
-                if (!page?.itemIds) {
+                // Check that we have a valid page
+                if (!(page instanceof CollectionPage) || !page?.itemIds) {
                     page = null;
                 }
             } else {
@@ -375,7 +373,6 @@ export function createGetProfileFollowersHandler(
         };
 
         let page: CollectionPage | null = null;
-        let followers: Collection | null = null;
 
         try {
             if (next !== '') {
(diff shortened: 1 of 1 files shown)


src/http/api/profile.ts
```diff
@@ -410,6 +412,57 @@ export function createGetProfileFollowersHandler(
...
+                if (!actor) {
+                    continue;
+                }
+
+                result.followers.push({
+                    actor: await actor.toJsonLd({
+                        format: 'compact',
+                    }),
+                    isFollowing: await isFollowedByDefaultSiteAccount(
+                        actor as Actor,
```
allouis, line 444: We're casting to `Actor` here but we don't know it is one! We can instead call `isActor` where we do the `if (!actor) continue` check

src/http/api/profile.ts
```diff
@@ -394,6 +396,57 @@ export function createGetProfileFollowersHandler(
...
+                headers: {
+                    'Content-Type': 'application/json',
+                },
+                status: 200,
+            });
+        }
+
         if (!page) {
             return new Response(null, { status: 404 });
         }
```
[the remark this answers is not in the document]
allouis: I think this is valid - we can remove this

src/http/api/profile.ts
```diff
@@ -400,7 +402,7 @@ export function createGetProfileFollowersHandler(
                     page = null;
```
allouis, line 402: I know you didn't change this code - but I'm seeing the comment about not being type-safe, I think we can do an `instanceof CollectionPage` check instead, which is much safer

src/http/api/profile.ts
```diff
@@ -374,6 +375,7 @@ export function createGetProfileFollowersHandler(
         };
 
         let page: CollectionPage | null = null;
+        let followers: Collection | null = null;
```
allouis, line 378: We don't _need_ to have a shared `followers` variable here, we can just `await actor.getFollowers()` again in the second block - fedify won't do two lookups.

It's not a big deal, just letting you know ☺️

allouis (commented):
Main comment is about the `as Actor` cast not being safe, everything else is small - looks good

src/http/api/profile.ts
vershwal: Done ✅

PR #431, opened 2025-03-24
Fixed author for reposts when `attributedTo` is a string

vershwal:
Ref https://linear.app/ghost/issue/AP-962/

### DIFF
--- a/src/http/api/profile.ts
+++ b/src/http/api/profile.ts
@@ -273,6 +273,19 @@ export function createGetProfilePostsHandler(accountService: AccountService) {
                     });
                 }
 
+                if (typeof activity.object.attributedTo === 'string') {
+                    const attributedTo = await lookupObject(
+                        apCtx,
+                        activity.object.attributedTo,
+                    );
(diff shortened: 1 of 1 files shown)


src/http/api/profile.ts
```diff
@@ -273,6 +273,19 @@ export function createGetProfilePostsHandler(accountService: AccountService) {
                     });
                 }
 
+                if (typeof activity.object.attributedTo === 'string') {
+                    const attributedTo = await lookupObject(
+                        apCtx,
+                        activity.object.attributedTo,
+                    );
+                    if (isActor(attributedTo)) {
```
mike182uk, line 281: What if this returns falsy? We will still have the same issue right? We could just adding the post to `result.posts` in this case maybe?

mike182uk (approved):

src/http/api/profile.ts
vershwal: I agree. It's worth not adding it to `result.posts` in that case

PR #441, opened 2025-03-26
Added notification for an account being followed

mike182uk:
ref https://linear.app/ghost/issue/AP-901

Update the system to emit a `account.followed` event when an account is followed. This event is then picked up by the `NotificationEventService` which will insert a notification into the database for the account being followed.

### DIFF
--- a/src/account/account-followed.event.ts
+++ b/src/account/account-followed.event.ts
@@ -1,0 +1,20 @@
+import type { Account } from 'account/types';
+
+export class AccountFollowedEvent {
+    constructor(
+        private readonly account: Account,
+        private readonly follower: Account,
+    ) {}
+
+    getAccount(): Account {
+        return this.account;
+    }
+
+    getFollower(): Account {
+        return this.follower;
+    }
+
+    static getName(): string {
+        return 'account.followed';
+    }
+}
--- a/src/notification/notification-event.service.ts
+++ b/src/notification/notification-event.service.ts
@@ -1,0 +1,24 @@
+import type { EventEmitter } from 'node:events';
+import { AccountFollowedEvent } from 'account/account-followed.event';
+import type { NotificationService } from 'notification/notification.service';
+
+export class NotificationEventService {
+    constructor(
+        private readonly events: EventEmitter,
+        private readonly notificationService: NotificationService,
+    ) {}
+
+    init() {
+        this.events.on(
+            AccountFollowedEvent.getName(),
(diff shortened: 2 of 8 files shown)


src/account/account-followed.event.ts
```diff
@@ -0,0 +1,20 @@
+import type { Account } from 'account/types';
```
mike182uk, line 1: I know we wanted to get away from using the `Account` type, but it felt like too big a jump to do as part of this work

To use the `Account` entity `AccountService.recordAccountFollow` would need to be updated to accept `Account` entity instead of `Account` type

`AccountService.recordAccountFollow` is used in the `Follow` & `Accept` handlers to resolve the follower / followee. If one of these does not exist we call `AccountService.createExternalAccount` which returns `Account` type and we then pass this to `AccountService.recordAccountFollow`. This means we would need to update `AccountService.createExternalAccount` to return an `Account` entity, which is used in the `Follow` & `Accept` handlers as well as `followAction` and `unfollowAction` (and some tests).

I feel this migration would be more appropriate outside of the scope of this PR (i know this technically adds more work to migration) to keep the changeset small / focused on what it is actually doing.

Thoughts?

vershwal: It’s always nicer to clean things up when we’re in the area and context loaded, but I had a look at the `recordAccountFollow`, `createExternalAccount` and related test, and yeah, this one’s definitely a bit too much to squeeze into this PR. Would’ve made it way harder to review/understand the main logic of this PR.

Makes sense to leave it out for now. 
Just wondering — are we thinking of doing that migration all at once later, or more incrementally? Like maybe you plan to do a quick follow-up PR after this to clean it up?

src/notification/notification-event.service.ts
```diff
@@ -0,0 +1,24 @@
+import type { EventEmitter } from 'node:events';
+import { AccountFollowedEvent } from 'account/account-followed.event';
+import type { NotificationService } from 'notification/notification.service';
+
+export class NotificationEventService {
```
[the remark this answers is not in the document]
vershwal: Honestly, I don’t have a better name in mind right now either 😅 but happy to revisit if something clicks later.

vershwal (approved):
LGTM

src/account/account-followed.event.ts
mike182uk: Yeh i think it will probably be incremental. Once I've got this in, I'll make the changes noted above in a separate PR and ping you 👍

PR #446, opened 2025-03-27
Updated `account.followed` handler to better handle non-existent users

mike182uk:
ref https://linear.app/ghost/issue/AP-901

Updated `account.followed` handler to return early instead of throw an error if a user associated with the account can not be found. This is because the handler will be executed for follows of "external accounts" (i.e a Ghost site follows a Mastodon account) and in this case it is valid that the user does not exist.


vershwal (approved):
Looks good!

PR #448, opened 2025-03-27
Set timezone to UTC in mysql connection config 

vershwal:
Ref https://linear.app/ghost/issue/AP-886/
Ref https://linear.app/ghost/issue/AP-992/


sagzy (approved):
Nice one! LGTM 👍

PR #450, opened 2025-03-27
Updated precision of timestamp columns in accounts table

vershwal:
Ref https://linear.app/ghost/issue/AP-947/
Ref https://linear.app/ghost/issue/AP-993/


sagzy (approved):
LGTM

Curious that we have allowed `NULL` for `created_at` historically; I don't think we should change that in the same migration to avoid trouble, but I don't know whether this was an intentional decision

PR #461, opened 2025-03-28
Update `AccountService.recordAccountFollow` to use `Account` entity

mike182uk:
refs https://github.com/TryGhost/ActivityPub/pull/441/files#diff-fb5a965e319e4cc068e247181f63852dc961379121ca4cad8a64a70d025866a5

- Updated `Account` type to have `uuid` as this was missing
- Introduced `asAccountEntity` helper that maps `Account` type to `Account` entity
- Updated `AccountService.recordAccountFollow` to accept `Account` entity instead of `Account` type

### DIFF
--- a/src/account/utils.ts
+++ b/src/account/utils.ts
@@ -1,5 +1,6 @@
 import { type Actor, PropertyValue } from '@fedify/fedify';
-import type { ExternalAccountData } from './types';
+import { Account } from './account.entity';
+import type { Account as AccountType, ExternalAccountData } from './types';
 
 interface PublicKey {
     id: string;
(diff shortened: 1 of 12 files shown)


src/account/utils.ts
```diff
@@ -77,3 +78,24 @@ export async function mapActorToExternalAccountData(
...
+export function asAccountEntity(account: AccountType): Account {
+    return new Account(
+        account.id,
+        account.uuid,
+        account.username,
+        account.name,
+        account.bio,
+        account.avatar_url ? new URL(account.avatar_url) : null,
+        account.banner_image_url ? new URL(account.banner_image_url) : null,
+        null, // site
```
mike182uk, line 96: `AccountType` doesn't have the site data, so we should in theory be safe to leave this as null as anywhere that is using `AccountType` currently won't be accessing the site. We could pass the site as another argument to the fn if needed?

mike182uk:
Closing this in favour of modelling the follow / unfollow functionality within the account entity

PR #468, opened 2025-03-31
Improved `handleGetProfilePosts` perf

mike182uk:
ref https://linear.app/ghost/issue/AP-1018

Changed `handleGetProfilePosts` to retrieve repost & reply counts for a post in the outbox from the `posts` table instead of computing the counts based on the data in the `key_value` table. This should make retrieving posts for a remote account more peformant.


vershwal (approved):
looks good!

PR #470, opened 2025-04-01
Added local lookup for Ghost sites in the profile view

vershwal:
Ref https://linear.app/ghost/issue/AP-1019/

- Removed `GET '/.ghost/activitypub/profile/:handle'`. 
- Using `GET /.ghost/activitypub/account/@username@domain'` to get profile for both local ghost users and remote lookup


features/account.feature
```diff
@@ -0,0 +1,32 @@
...
+    And we are followed by "Alice"
+    And we are followed by "Bob"
+
+  Scenario: Get default account
+    When an authenticated "get" request is made to "/.ghost/activitypub/account"
+    Then the request is accepted with a 200
+    And the response contains our account details
+
+  Scenario: Get account by handle
+    When an authenticated "get" request is made to "/.ghost/activitypub/account?handle=@johnonolan@mastodon.xyz"
```
vershwal, line 18: Not sure if this is the best thing to do but this API does a actual remote lookup so can't test it with a fake handle

src/http/api/types.ts
```diff
@@ -61,6 +61,19 @@ export interface AccountDTO {
      * Whether the account of the current user is following this account
      */
     followedByMe: boolean;
+    /**
+     * Attachments of the account
+     */
+    attachment: {
+        /**
```
vershwal, line 68: Local ghost profiles don't have  attachment but remote profile might have them.

features/account.feature
allouis: We shouldn't add an actual external network lookup in our tests - we have a `fake-external-activitypub` service which we can use here, it's a wiremock instance and we can have it return whatever we like

src/http/api/helpers/account.ts
```diff
@@ -0,0 +1,121 @@
...
+import type { AccountDTO } from '../types';
+
+/**
+ * Checks if an account is internal
+ * @param account - The account to check
+ *
+ * @returns boolean indicating if the account is internal
+ */
+export function isInternalAccount(account: AccountType): boolean {
+    return account.ap_id.toString().includes(AP_BASE_PATH);
```
allouis, line 25: This will match other instances of Ghost AP - so I don't think it's the correct check.

The `AccountType` is deprecated and shouldn't be used, instead you can use the `Account` entity, which has an `isInternal` check built in

src/http/api/account.ts
```diff
@@ -42,57 +49,49 @@ export function createGetAccountHandler(accountService: AccountService) {
...
+            globaldb: ctx.get('globaldb'),
+            logger,
+        });
 
-        if (handle === '') {
-            return new Response(null, { status: 400 });
+        const defaultAccount =
+            await accountService.getDefaultAccountForSite(site);
+
+        const handle = ctx.req.query('handle');
```
allouis, line 64: Curious - why did we move to a query param here? I think the existing pattern is route params, not a big deal I think but I'd like to understand why

src/http/api/account.ts
```diff
@@ -42,57 +49,49 @@ export function createGetAccountHandler(accountService: AccountService) {
...
-                    await accountService.getFollowerAccountsCount(account),
-                /**
-                 * At the moment we only expect to be returning the account for
-                 * the current user, so we can hardcode these values to false as
-                 * the account cannot follow, or be followed by itself
-                 */
-                followsMe: false,
-                followedByMe: false,
-            };
+            if (account && isInternalAccount(account)) {
```
allouis, line 81: I don't fully understand these if/else blocks 🤔 

If there is `!account` then we should 404 straight away I think - because the remote lookup is already handled by `getAccountByApId`, that means we don't need to do a remote lookup in the `getAccountDTOByHandle` unless I'm missing something here?

So then I'm unclear on the difference between internal and external accounts - I think it's just the counts for followers/posts/etc...

If we look at the pattern for the helpers for the post API - it takes just the entity & some metadata, doesn't do any remote lookups or anything - just converts it to an object ready to send down the API.

I think we can use the same approach here where we have an if/else block to load the counts differently, and then once we have all the information - we convert it to JSON and respond with it

allouis (commented):
Few comments on structure, but the approach looks good - this should speed things up!

src/http/api/helpers/account.ts
vershwal: Great, I wanted something more reliable than checking for AP_BASE_PATH. Thanks!!

I wanted to use Account entity but some of the functions that I needed here used AccountType because I din't want to end up making this PR too big. Let me check if I can use Account entity without too many changes to the existing stuff.

src/http/api/account.ts
vershwal: Earlier in the account API the handle was always passed as 'index' and it assumed that it needs to return the profile of the current user. 

I changed it to :-

Getting profile for myself -> `GET /.ghost/activitypub/account`
Getting profile for others -> `GET /.ghost/activitypub/account/?handle=@username@domain`

My thinking is that the handle is optional so it should be in a query param. 

Other approach could be to keep sending 'index' for fetching "My profile" and have a check in backend.

src/http/api/account.ts
vershwal: Oh, I am actually using [getAccountByApId](https://github.com/TryGhost/ActivityPub/blob/main/src/account/account.service.ts#L217-L223) from the `account.service.ts` that doesn't do a remote lookup but only check our own db.

 But I really like the approach you suggested. I think it will make things much cleaner. Gonna try this!

src/http/api/helpers/account.ts
vershwal: I have changed this to using `Account` entity ✅

src/http/api/account.ts
vershwal: Changed it to route params to make consistent with other endpoints

src/http/api/account.ts
allouis: Oooh I see - sorry, I though it was implemented the same as `getByApId` for the post service - that's my bad!
But yeah I think following the same pattern as the PostService will make this a lot nicer!

src/helpers/activitypub/actor.ts
```diff
@@ -83,12 +83,17 @@ export async function isFollowedByDefaultSiteAccount(
...
+        siteDefaultAccount.id,
+        followeeAccount.id,
     );
 }
 
 export function isHandle(handle: string): boolean {
+    // In test environments, we have handles that don't match the regex. Ex: @Alice@fake-external-activitypub
+    if (process.env.NODE_ENV === 'testing') {
+        return true;
+    }
```
mike182uk, line 95: How difficult would it be to make a change that meant we didn't have to do this? (i.e update test environment to use valid handles)

src/http/api/account.ts
```diff
@@ -42,57 +52,52 @@ export function createGetAccountHandler(accountService: AccountService) {
     return async function handleGetAccount(ctx: AppContext) {
         const logger = ctx.get('logger');
         const site = ctx.get('site');
+        let account: Account | null = null;
+        const db = ctx.get('db');
 
-        // Validate input
-        const handle = ctx.req.param('handle') || '';
+        const apCtx = fedify.createContext(ctx.req.raw as Request, {
```
mike182uk, line 58: We could inject `fedify` instead of importing it in?

src/http/api/account.ts
```diff
@@ -1,8 +1,15 @@
...
 import { getAccountHandle } from 'account/utils';
-import type { AppContext } from 'app';
-import { sanitizeHtml } from 'helpers/html';
+import { type AppContext, fedify } from 'app';
+import { isHandle } from 'helpers/activitypub/actor';
+import { lookupAPIdByHandle } from 'lookup-helpers';
 import type { PostService } from 'post/post.service';
+import {
+    getAccountDTOByHandle,
+    getAccountDtoFromAccount,
```
mike182uk, line 11: Any reason to use different casing on `DTO`?

src/http/api/account.ts
```diff
@@ -42,57 +52,52 @@ export function createGetAccountHandler(accountService: AccountService) {
...
+        });
 
-        if (handle === '') {
-            return new Response(null, { status: 400 });
+        const defaultAccount = await accountRepository.getBySite(
+            ctx.get('site'),
+        );
+
+        const handle = ctx.req.param('handle');
+        if (handle === 'index') {
```
mike182uk, line 69: Hm, if would be nice if we didn't have to hardcode this. What about assuming if the param is empty that the request is for the current user?

src/http/api/account.ts
```diff
@@ -42,57 +52,52 @@ export function createGetAccountHandler(accountService: AccountService) {
...
+            if (account?.isInternal) {
+                accountDto = await getAccountDtoFromAccount(
+                    account,
+                    defaultAccount,
+                    accountService,
+                );
+            } else {
+                //Otherwise, do a remote lookup to fetch the updated data
+                accountDto = await getAccountDTOByHandle(
+                    handle || '',
```
mike182uk, line 95: Why do we need to default to `''`? If the handle is empty we shouldn't even get to this right?

src/http/api/helpers/account.ts
```diff
@@ -0,0 +1,115 @@
...
+    getHandle,
+    isFollowedByDefaultSiteAccount,
+} from 'helpers/activitypub/actor';
+import { sanitizeHtml } from 'helpers/html';
+import { lookupObject } from 'lookup-helpers';
+import type { AccountDTO } from '../types';
+
+/**
+ * Converts an Account to an AccountDTO
+ * @param account - The account to convert
```
mike182uk, line 20: Nitpick: Space between the desc and param, no need for the ` - ` between the param and desc

src/http/api/account.ts
vershwal: No reason. just a mistake

src/http/api/account.ts
```diff
@@ -42,57 +52,52 @@ export function createGetAccountHandler(accountService: AccountService) {
...
-        const handle = ctx.req.param('handle') || '';
+        const apCtx = fedify.createContext(ctx.req.raw as Request, {
+            db,
+            globaldb: ctx.get('globaldb'),
+            logger,
+        });
 
-        if (handle === '') {
-            return new Response(null, { status: 400 });
+        const defaultAccount = await accountRepository.getBySite(
```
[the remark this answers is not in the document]
vershwal: The reason for doing this is because I need the default account [here](https://github.com/TryGhost/ActivityPub/pull/470/files#diff-6f156cf4702dbf461441adc25316cacf44c4529812180b6455a315616be7365cR54) to find if the profile is followed by the defaultAccount. So I need it even if the handle is not index.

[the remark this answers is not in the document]
mike182uk: Ah i see, didn't spot that. All good 👍

PR #476, opened 2025-04-02
Converted URLs in Notes to links

allouis:
closes https://linear.app/ghost/issue/AP-923

When we create notes and replies we want any URL's to be clickable links for clients, both in Ghost and in third party services like mastodon. In order to ensure this we convert the URL to an anchor tag, which is then used both internally in Ghost & sent as the content over the ActivityPub network.

### DIFF
--- a/package.json
+++ b/package.json
@@ -73,6 +73,8 @@
     "jsonwebtoken": "9.0.2",
     "knex": "3.1.0",
     "ky": "1.7.3",
+    "linkify-html": "4.2.0",
+    "linkifyjs": "4.2.0",
     "mysql2": "3.11.5",
     "node-jose": "2.2.0",
     "sanitize-html": "2.14.0",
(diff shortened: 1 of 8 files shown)


package.json
```diff
@@ -73,6 +73,8 @@
     "jsonwebtoken": "9.0.2",
     "knex": "3.1.0",
     "ky": "1.7.3",
+    "linkify-html": "4.2.0",
+    "linkifyjs": "4.2.0",
```
allouis, line 77: This is a peer dependency of linkify-html

src/post/post.entity.ts
```diff
@@ -302,6 +302,7 @@ export class Post extends BaseEntity {
             escapeHtml: true,
             convertLineBreaks: true,
             wrapInParagraph: true,
+            extractLinks: true,
```
vershwal, line 305: We have to add it [here](https://github.com/TryGhost/ActivityPub/blob/main/src/post/post.entity.ts#L350) as well to make it work with replies. 

Optional but I think it will be good to have a small test for `createReply` too.

vershwal (commented):
Overall looks good, added one comment!

src/post/post.entity.ts
allouis: Great catch, thank you!

vershwal (approved):

PR #481, opened 2025-04-03
Remove duplicate `accounts` rows

allouis:
This is a pretty gnarly migration and has taken a few days to piece together.

We have duplicate (by ap_id) rows in the `accounts` table, which this migration
will clean up completely.

First we create a temporary table `account_masters` which serves as a mapping
between each ap_id and its "master" account, the account that we want to keep.

We populate this initially with internal accounts, using an inner join we only
get ap_ids which have a corresponding row in the users table.

After this we create a list of the remaining ap_ids, and use that to populate
the `account_masters` with the lowest id of each group of ap_ids.

From here we create a temporary table `accounts_to_migrate` which serves as a
mapping between a duplicate account id and its master account id.

Then we update all tables which reference `accounts.id` to update the link from
a duplicate id to the master. Note that for `likes` and `reposts` there is a
unique constraint covering the `account_id` columns - which means it's possible
we violate it when doing the UPDATE. We handle that by doing an UPDATE IGNORE
for these tables, which is safe because we also have an ON DELETE CASCADE
trigger set up for these columns which will be triggered by the next step.

At this point we still have all accounts in our system, however all duplicate
accounts are not referenced anywhere in the database (with the exception of
some rows in the `likes` and `reposts` table as mentioned above), any previous
references have been updated to point at the master account.

We can now delete all duplicate accounts, triggered any ON DELETE CASCADE which
cleans up remaining rows.

In order to prevent this from happening we want to use the same approach as the
`posts` table, so we add an `ap_id_hash` column to the `accounts` table as well
as a UNIQUE constraint over it.

### DIFF
--- a/migrate/migrations/000038_remove-duplicate-account-rows.down.sql
+++ b/migrate/migrations/000038_remove-duplicate-account-rows.down.sql
@@ -1,0 +1,5 @@
+ALTER TABLE accounts
+DROP INDEX idx_ap_id_hash;
+
+ALTER TABLE accounts
+DROP COLUMN ap_id_hash;
--- a/migrate/migrations/000038_remove-duplicate-account-rows.up.sql
+++ b/migrate/migrations/000038_remove-duplicate-account-rows.up.sql
@@ -1,0 +1,85 @@
+START TRANSACTION;
+
+CREATE TEMPORARY TABLE account_masters (
+    ap_id VARCHAR(255) NOT NULL,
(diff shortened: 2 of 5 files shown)


rmgpinto:
The `bin/create` is not on the migrations job, do we need to run it or it's just for local debugging?

allouis:
@rmgpinto That's just local tooling to create blank migration files - sorry for the confusion

mike182uk (approved):

PR #483, opened 2025-04-03
Improved pref for loading profile posts

vershwal:
Ref: https://linear.app/ghost/issue/AP-1024

Removed `GET '/.ghost/activitypub/profile/:handle/posts'`
Using `GET /.ghost/activitypub/posts/@username@domain'` to get profile posts for both local ghost users and remote lookup


src/app.ts
```diff
@@ -977,12 +971,12 @@ app.get(
...
 app.get(
-    '/.ghost/activitypub/posts',
+    '/.ghost/activitypub/posts/:handle',
     requireRole(GhostRole.Owner, GhostRole.Administrator),
-    spanWrapper(createGetAccountPostsHandler(accountService, postService)),
+    spanWrapper(createGetAccountPostsHandler(postService, accountRepository)),
 );
 app.get(
-    '/.ghost/activitypub/posts/liked',
+    '/.ghost/activitypub/posts/liked/:handle',
```
vershwal, line 979: Had to add the handle here as `posts/liked` was otherwise routing to `posts/:handle`

src/post/post.repository.knex.ts
```diff
@@ -874,7 +874,10 @@ export class KnexPostRepository {
      * @param accountId ID of the account to check
      * @returns True if the post is liked by the account, false otherwise
      */
-    async isLikedByAccount(postId: number, accountId: number) {
+    async isLikedByAccount(postId: number, accountId: number | null) {
+        if (accountId === null) {
```
vershwal, line 878: Added a null check because `accountId` can be null for `AccountEntity`

src/post/post.repository.knex.ts
```diff
@@ -892,7 +895,10 @@ export class KnexPostRepository {
      * @param accountId ID of the account to check
      * @returns True if the post is reposted by the account, false otherwise
      */
-    async isRepostedByAccount(postId: number, accountId: number) {
+    async isRepostedByAccount(postId: number, accountId: number | null) {
+        if (accountId === null) {
```
vershwal, line 899: Added a null check because accountId can be null for AccountEntity

src/post/post.service.ts
```diff
@@ -359,7 +371,7 @@ export class PostService {
                 this.on('likes.post_id', 'posts_with_source.id').andOnVal(
                     'likes.account_id',
                     '=',
-                    accountId.toString(),
+                    defaultAccountId.toString(),
```
vershwal, line 374: Earlier this query was used to get profile post of self account so we were checking if the post is liked/reposted by the account. Updated this to check if the post is liked/reposted by default account.

src/post/post.service.ts
```diff
@@ -492,6 +504,174 @@ export class PostService {
...
+     *
+     * @param handle - The handle to look up (e.g., "@username@domain")
+     *
+     */
+    async getPostsByRemoteLookUp(
+        defaultAccount: Account,
+        handle: string,
+        next: string,
+    ): Promise<GetProfileDataResult | Error> {
+        const context = this.fedifyContextFactory.getFedifyContext();
```
vershwal, line 518: This function has the exact same code deleted from `/profile.ts`

src/app.ts
```diff
@@ -962,27 +961,26 @@ app.get(
...
     '/.ghost/activitypub/account/:handle',
     requireRole(GhostRole.Owner, GhostRole.Administrator),
-    spanWrapper(createGetAccountHandler(accountService, accountRepository)),
+    spanWrapper(
+        createGetAccountHandler(accountService, accountRepository, fedify),
+    ),
 );
 app.get(
-    '/.ghost/activitypub/posts',
+    '/.ghost/activitypub/posts/:handle',
```
vershwal, line 976: This endpoint returns a list of Activity for remote profiles and a list of Posts for local profiles. I went with this approach because mapping activity to postDTO here in backend and then converting it back to activity in the FE just felt like extra work.
I did try mapping it though, but it wasn’t super smooth and caused a bunch of UI issues that needed a lot of manual testing. Doing it this way made testing and shipping a lot easier.

src/post/post.service.ts
vershwal: I have not used the `remoteRepository` here in this PR because by the time we discussed that I already did the BE changes. I'll do that refactoring incrementally after shipping these changes.

src/app.ts
mike182uk: Mentioned this on the FE PR, what about using `/posts/me/liked` instead?

src/app.ts
mike182uk: I know what you are saying, it does seem a little counter productive BUT we are eventually going to migrate the client to using `posts` and not activities so we are going to have to do this at some point. I think it would be worth while doing this now and not having to worry about the backend when we come to update the frontend. I think its also a bit of a smell that an endpoint could potentially return 2 different types of data and its not obvious why this is happening.

Could we not just use the same mapping fn thats used on the frontend?

src/post/post.repository.knex.ts
vershwal: Fixed it!

src/app.ts
vershwal: Added a mapping function [here](https://github.com/TryGhost/ActivityPub/pull/483/files#diff-eb3249a9a93286287064f8d43e8e503ae447978501b3af5b5f1c3d76eb1341a4R151).

src/post/post.service.ts
vershwal: This was a query change so I did it separately [here](https://github.com/TryGhost/ActivityPub/pull/488). 

> What is accountId now being used for

It is used to get the posts for that account. Where as the `defaultAccountId` is used to check if the post is liked/reposted by the default/current user.

PR #488, opened 2025-04-08
Updated `getPostsByAccount` query to check if a post id liked/reposted by the current user

vershwal:
Ref https://linear.app/ghost/issue/AP-1024/

Previously, this query (`getPostsByAccount`) was used to fetch only the posts of the current user for the Profile view, so `likedByMe` and `repostedByMe` indicated whether the posts were liked or reposted by that account. Now, this query will also fetch posts from other ghost accounts, so the `likedByMe` and `repostedByMe` fields should indicate whether the current user (the one viewing the profile) has liked or reposted the posts.

### DIFF
--- a/src/http/api/account.ts
+++ b/src/http/api/account.ts
@@ -257,6 +257,7 @@ export function createGetAccountPostsHandler(
         );
         const { results, nextCursor } = await postService.getPostsByAccount(
             account.id,
+            account.id,
             params.limit,
             params.cursor,
         );
--- a/src/post/post.service.ts
+++ b/src/post/post.service.ts
@@ -17,11 +17,11 @@ interface BaseGetProfileDataResultRow {
     post_image_url: string | null;
     post_published_at: Date;
     post_like_count: number;
-    post_liked_by_user: 0 | 1;
+    post_liked_by_current_user: 0 | 1;
     post_reply_count: number;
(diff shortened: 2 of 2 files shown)


src/http/api/account.ts
```diff
@@ -256,6 +256,7 @@ export function createGetAccountPostsHandler(
             ctx.get('site'),
         );
         const { results, nextCursor } = await postService.getPostsByAccount(
+            account.id,
```
vershwal, line 259: For now, we're passing `account.id` as `defaultAccount` because they are the same in this context, but this will change in my next PR.

allouis (approved):
I don't think we needed to rename all of the column/table aliases here 🤔 It would have been a 4 line diff which is a lot easier to review!

Do you have plans to move this code into the AccountRepository? Based on the stuff we spoke about yesterday we should have like Service -> Repository -> DB

When we move to the AccountRepository it would be good to have the return type as similar as possible to the `getThread` method - so that they can share a DTO serialiser

vershwal:
Ah, fair point! I just thought the alias might be confusing with two users, so I renamed it. But yeah, a small diff would’ve been easier. Thanks for the feedback! 
The refactoring chances - yup on my list - That would come in another PR

PR #496, opened 2025-04-08
Added various counts onto the `Account` entity

mike182uk:
no ref

Added the following counts onto the `Account` entity:

- `postCount`
- `repostCount`
- `likedPostCount`
- `followerCount`
- `followingCount`

Having these counts on the entity means we can move away from using "helper" methods in services / repositories to retrieve this information from the database. This is particulary useful when we retrieve a "remote" account and we can't utilise the database to retrieve these count values (as we don't have them). This will later come into play when we refactor the account service to fetch account data either locally or remotely and return a single `Account` entity that can be serialised by the controller action.

### DIFF
--- a/src/account/account.repository.knex.ts
+++ b/src/account/account.repository.knex.ts
@@ -26,6 +26,33 @@ export class KnexAccountRepository {
         // We can safely assume that there is an account for the user due to
         // the foreign key constraint on the users table
         const account = await this.db('accounts')
+            .select(
+                'accounts.id',
+                'accounts.uuid',
+                'accounts.username',
+                'accounts.name',
+                'accounts.bio',
+                'accounts.avatar_url',
+                'accounts.banner_image_url',
+                'accounts.ap_id',
+                'accounts.url',
+                'accounts.ap_followers_url',
+                this.db.raw(
+                    '(select count(*) from posts where posts.author_id = accounts.id) as post_count',
+                ),
+                this.db.raw(
+                    '(select count(*) from likes where likes.account_id = accounts.id) as like_count',
+                ),
+                this.db.raw(
+                    '(select count(*) from reposts where reposts.account_id = accounts.id) as repost_count',
+                ),
+                this.db.raw(
+                    '(select count(*) from follows where follows.follower_id = accounts.id) as following_count',
+                ),
+                this.db.raw(
+                    '(select count(*) from follows where follows.following_id = accounts.id) as follower_count',
+                ),
+            )
             .where('id', user.account_id)
             .first();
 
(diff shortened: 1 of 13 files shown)


src/account/account.repository.knex.ts
```diff
@@ -75,6 +107,21 @@ export class KnexAccountRepository {
...
+                ),
+                this.db.raw(
+                    '(select count(*) from reposts where reposts.account_id = accounts.id) as repost_count',
+                ),
+                this.db.raw(
+                    '(select count(*) from follows where follows.follower_id = accounts.id) as following_count',
+                ),
+                this.db.raw(
+                    '(select count(*) from follows where follows.following_id = accounts.id) as follower_count',
+                ),
```
vershwal, line 124: Just a thought — These methods - `getByApId` and `getBySite` are used in multiple places where the count fields (`follower_count`, `repost_count`, etc.) aren’t actually needed. Running those extra queries in those cases ends up being unnecessary DB calls.

Maybe we could make those count fields `number | null` and only fetch them when explicitly needed (with something like an `includeCounts` flag - which can be `false` by default)? Could help avoid some extra DB load in the cases where we don't need the count data.

mike182uk: Yeh I was a little concerned re: this, but this then creates a situation where you can't fully trust that the `Account` entity has all of the relevant data. If somewhere in the system I get passed an `Account` that doesn't have this data, how do I know the data is missing vs the data is correct (i.e the counts are zero). How would I then get this data? If I've been passed this `Account`, I have no idea where its come from. Is it safe to assume I can call some service and get an `Account` with the exact same representation, but this extra count data?

I think the ideal would maybe having these counts on the account record in the database, similar to what we do for `posts`, but this is feels like quiet a big jump for what I wanted to do here

vershwal: > how do I know the data is missing vs the data is correct
You have a fair point here but just to add, we can only trust this data in case of our internal accounts - this is something we'll have to be aware of while using this data for any other use case. 

Yeah, storing these numbers in db is def out of scope of what we want to do here. Maybe an improvement for later.

mike182uk: Ah yeh, but in this case we do have a way to differentiate between an internal & external account via `isInternal`, so in that case we would know why the counts are zero, but this does poke a hole in my logic.

Thinking out loud, an alternative could be have some way to decorate the `Account` with the data? Like having something that takes type` Account` and returns a new type `AccountWithCounts` that is derived from `Account`

```
function getCountsForAccount(account: Account): Promise<AccountWithCounts>
```

Then the consumer could check the type of `Account` (i think) to know if it definitely has counts?

```
if account is not AccountWithCounts
    someService.getGetCountsForAccount(account)
```

This might be just over-complicating things though 😅

mike182uk:
Closing this in favour of using the `View` pattern

PR #503, opened 2025-04-09
Refactored follows logic to AccountFollowsView

vershwal:
Ref: https://linear.app/ghost/issue/AP-1042
Ref: https://linear.app/ghost/issue/AP-1041

Refactored the follows lookup to AccountFollowsViewer, this does not have any additional logic, just the refactoring changes.

### DIFF
--- a/src/http/api/views/account.follows.viewer.integration.test.ts
+++ b/src/http/api/views/account.follows.viewer.integration.test.ts
@@ -1,0 +1,434 @@
+import { AsyncEvents } from 'core/events';
+import type { Knex } from 'knex';
+import { generateTestCryptoKeyPair } from 'test/crypto-key-pair';
+import { createTestDb } from 'test/db';
+import {
+    afterEach,
+    beforeAll,
+    beforeEach,
+    describe,
+    expect,
+    it,
+    vi,
+} from 'vitest';
+import { KnexAccountRepository } from '../../../account/account.repository.knex';
+import { AccountService } from '../../../account/account.service';
+import type { InternalAccountData, Site } from '../../../account/types';
+import { FedifyContextFactory } from '../../../activitypub/fedify-context.factory';
+import { AccountFollowsViewer } from './account.follows.viewer';
+
+vi.mock('@fedify/fedify', async () => {
+    const actual = await vi.importActual('@fedify/fedify');
+    return {
+        ...actual,
+    };
+});
+
+describe('AccountFollowsViewer', () => {
+    let viewer: AccountFollowsViewer;
+    let accountService: AccountService;
+    let events: AsyncEvents;
+    let site: Site;
+    let internalAccountData: InternalAccountData;
+    let db: Knex;
+
+    beforeAll(async () => {
+        db = await createTestDb();
+    });
+
+    afterEach(() => {
+        events.removeAllListeners();
+    });
+
+    beforeEach(async () => {
+        vi.useRealTimers();
+
+        // Clean up the database
+        await db.raw('SET FOREIGN_KEY_CHECKS = 0');
+        await db('follows').truncate();
+        await db('accounts').truncate();
+        await db('users').truncate();
+        await db('sites').truncate();
+        await db.raw('SET FOREIGN_KEY_CHECKS = 1');
+
+        // Insert a site
+        const siteData = {
+            host: 'www.example.com',
+            webhook_secret: 'secret',
+        };
+        const [id] = await db('sites').insert(siteData);
+
+        site = {
+            id,
+            ...siteData,
+        };
+
+        // Init reusable account data
+        internalAccountData = {
+            username: 'index',
+            name: 'Test Site Title',
+            bio: 'Test Site Description',
+            avatar_url: 'https://example.com/avatar.jpg',
+        };
+
+        // Init dependencies
+        events = new AsyncEvents();
+        const accountRepository = new KnexAccountRepository(db, events);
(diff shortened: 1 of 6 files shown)


src/http/api/views/account.follows.viewer.ts
```diff
@@ -0,0 +1,228 @@
+import type { Account } from 'account/account.entity';
+import { getAccountHandle } from 'account/utils';
+import type { Knex } from 'knex';
+import type { FedifyContextFactory } from '../../../activitypub/fedify-context.factory';
```
allouis, line 4: I think this should be a non relative import?

src/http/api/views/account.follows.viewer.ts
```diff
@@ -0,0 +1,228 @@
...
+interface AccountRow {
+    id: number;
+    ap_id: string;
+    name: string;
+    username: string;
+    avatar_url: string;
+}
+
+export class AccountFollowsViewer {
+    /**
```
allouis, line 34: This JSDoc doesn't have all the arguments - but I think we can just remove it, it doesn't really provide any extra info

src/http/api/views/account.follows.viewer.ts
```diff
@@ -0,0 +1,228 @@
...
+        };
+        return accountFollowsView;
+    }
+
+    /**
+     * Get the number of accounts that the provided account is following
+     *
+     * @param accountId id of the account
+     */
+    async getFollowingAccountsCount(accountId: number | null): Promise<number> {
```
allouis, line 107: Should this (and the following methods) be private?

src/http/api/views/account.follows.viewer.integration.test.ts
```diff
@@ -0,0 +1,434 @@
...
+import {
+    afterEach,
+    beforeAll,
+    beforeEach,
+    describe,
+    expect,
+    it,
+    vi,
+} from 'vitest';
+import { KnexAccountRepository } from '../../../account/account.repository.knex';
```
allouis, line 14: Should these be absolute imports too? 🤔

src/http/api/views/account.follows.viewer.integration.test.ts
```diff
@@ -0,0 +1,434 @@
...
+import { AccountService } from '../../../account/account.service';
+import type { InternalAccountData, Site } from '../../../account/types';
+import { FedifyContextFactory } from '../../../activitypub/fedify-context.factory';
+import { AccountFollowsViewer } from './account.follows.viewer';
+
+vi.mock('@fedify/fedify', async () => {
```
allouis, line 20: Why do we mock fedify here and then replace it with the actual import?

src/http/api/views/account.follows.viewer.integration.test.ts
```diff
@@ -0,0 +1,434 @@
...
+    beforeAll(async () => {
+        db = await createTestDb();
+    });
+
+    afterEach(() => {
+        events.removeAllListeners();
+    });
+
+    beforeEach(async () => {
+        vi.useRealTimers();
```
allouis, line 44: Why do we need this?

src/http/api/views/account.follows.viewer.integration.test.ts
```diff
@@ -0,0 +1,434 @@
...
+    let site: Site;
+    let internalAccountData: InternalAccountData;
+    let db: Knex;
+
+    beforeAll(async () => {
+        db = await createTestDb();
+    });
+
+    afterEach(() => {
+        events.removeAllListeners();
```
allouis, line 40: I don't think we need this - `events` will just fall out of scope and be cleaned up

src/http/api/views/account.follows.viewer.ts
```diff
@@ -0,0 +1,228 @@
...
+
+        return Number(result[0].count);
+    }
+
+    /**
+     * Get the number of accounts that are following the provided account
+     *
+     * @param accountId id of the account
+     */
+    async getFollowerAccountsCount(accountId: number | null): Promise<number> {
```
allouis, line 124: Do we need this param to be nullable? I think it's preferable to have it as `number` and do any checking in the method above (which afaict already happens)

This makes it easier to reason about and we don't have to deal with default return values (e.g. an account that doesn't exist doesn't have 0 followers, it has undefined/null followers - the question doesn't make sense)

src/http/api/views/account.follows.viewer.ts
```diff
@@ -0,0 +1,228 @@
...
+            // the same time)
+            // @TODO: Make this configurable via the options?
+            .orderBy('follows.created_at', 'desc')
+            .orderBy('accounts.id', 'desc');
+    }
+
+    /**
+     * Check if an account is following another account
+     *
+     * @param accountId id of the account to check
```
allouis, line 209: I think if we're going to add JSDoc comments we should be adding more context

These descriptions don't give me much more than the parameter name itself - maybe we can remove this, or clarify exactly what's happening

allouis (commented):
Overall structure looks good - left a few specific comments though!

src/http/api/views/account.follows.viewer.integration.test.ts
vershwal: Removed this. Not needed

src/http/api/views/account.follows.viewer.integration.test.ts
vershwal: Not really! 🤔

PR #505, opened 2025-04-09
Refactored account retrieval to use `AccountView`

mike182uk:
no ref

Refactored account retrieval to use `AccountView`


src/http/api/account.ts
```diff
@@ -39,75 +43,37 @@ type FollowAccount = Pick<
 /**
  * Create a handler to handle a request for an account
  *
- * @param accountService Account service instance
+ * @param db Database client
```
allouis, line 46: Are these param comments useful to you?

src/http/api/account.ts
```diff
@@ -39,75 +43,37 @@ type FollowAccount = Pick<
...
+        if (handle === CURRENT_USER_KEYWORD) {
+            accountDto = await accountView.viewBySite(ctx.get('site'));
+        } else {
+            accountDto = await accountView.viewByHandle(handle, {
+                site: ctx.get('site'),
+            });
         }
 
-        // Return response
         return new Response(JSON.stringify(accountDto), {
```
allouis, line 77: We don't check for the `accountDto` being `null` here - I think we wanna 404 in that case probs

src/http/api/views/account.view.ts
```diff
@@ -0,0 +1,331 @@
...
+/**
+ * Additional context that can be passed to the view
+ */
+interface ViewContext {
+    /**
+     * Site instance - This will be used to resolve the default site account
+     * and determine if that account is following / followed by the account
+     * being viewed
+     */
+    site?: Site;
```
allouis, line 20: I _think_ we want context to be an `Account` rather than `Site` - otherwise we're leaking the idea of the default account for a site into this view.

If we pass in an account rather than a site, then when we eventually support multiple accounts for a site, the controller can just pass in like `ctx.get('authenticateAccount')` or something like this?

src/http/api/views/account.view.ts
```diff
@@ -0,0 +1,331 @@
...
+
+        const actor = await lookupObject(fedifyContext, apId);
+
+        if (actor === null) {
+            throw new Error(`Could not find Actor ${apId}`);
+        }
+
+        if (!isActor(actor)) {
+            return null;
+        }
```
allouis, line 230: I think this might just be a refactor/following some existing pattern - but I still think it's good to confirm, and change if we need to.

Why do we throw if `actor` is null, but we return `null` if it's not an actor? In this context I think I would expect a 404 in both cases 🤔

allouis (commented):
Overall structure good, except I think the ViewContext should change from Site -> Account

src/http/api/account.ts
```diff
@@ -39,75 +43,37 @@ type FollowAccount = Pick<
...
- * @param accountService Account service instance
+ * @param db Database client
+ * @param fedifyContextFactory Fedify context factory instance
  */
 export function createGetAccountHandler(
-    accountService: AccountService,
-    accountRepository: KnexAccountRepository,
-    fedify: Federation<ContextData>,
+    db: Knex,
+    fedifyContextFactory: FedifyContextFactory,
```
vershwal, line 51: Is there a reason we are passing `db` and `fedifyContextFactory` to this controller and then passing these two to the `AccountView`? 
Instead of passing `db` and `fedifyContextFactory` to the controller and then into AccountView, can we just create the AccountView instance in app.ts and inject it into the controller?

src/http/api/account.ts
mike182uk: not particularly no, just habit and i get a nice typescript description. Maybe we only include them when extra context is needed? (i.e `@param db Database client that can only read and not write`)

src/http/api/views/account.view.ts
mike182uk: Yeh i went with this so we don't have to double query the default account - If you are requesting your own profile, we cant just return this account, we still need to go via the view to then get the rest of the data needed (so 2 queries for similar data). Maybe not a big deal?

src/http/api/views/account.view.ts
mike182uk: Yeh that was existing code, but I agree, in both cases it should 404

src/http/api/account.ts
vershwal: Yeah, my thinking was — since we’re not using `db` or `fedifyContextFactory` directly in the controller, it felt a bit cleaner to move their usage up into `app.ts` and just pass in the `AccountView`. 
But open to thoughts — happy to go either way depending on how we see it fitting into the overall structure.

mike182uk: 🫡 

updated in https://github.com/TryGhost/ActivityPub/pull/505/commits/b153afa0f7f8e8df10371ec1d4711d80457f37fa

src/http/api/views/account.view.ts
mike182uk: updated in https://github.com/TryGhost/ActivityPub/pull/505/commits/e67df95aeebd6d4458b1f832baeeaf0fbee18e45

src/http/api/views/account.view.ts
mike182uk: updated in https://github.com/TryGhost/ActivityPub/pull/505/commits/731f134ca9eb12b9edef9a82908330fc258403fa

features/account.feature
```diff
@@ -20,8 +20,8 @@ Feature: Account API
 
   Scenario: Get non-existent account
     When an authenticated "get" request is made to "/.ghost/activitypub/account/@nonexistent@fake-external-activitypub"
-    Then the request is rejected with a 500
```
allouis, line 23: 😬 Yeah I think we should never have an explicit 500 check here.

I think we should update that stepdef to throw an error if a 5xx error is returned - what do you think?

mike182uk: Yeh I don't think it makes sense to expect an internal server error 😅 I'll add a check into the stepdef 👍

allouis (approved):
I think we should return Result instead of `AccountDTO | null` - but seeing as we're not throwing errors, this is fine for now, and we can use result when we need to throw/handle errors

PR #509, opened 2025-04-10
Improved perf for followers and following list

vershwal:
Ref: https://linear.app/ghost/issue/AP-1042
Ref: https://linear.app/ghost/issue/AP-1041


src/http/api/account.ts
```diff
@@ -121,30 +126,51 @@ export function createGetAccountFollowsHandler(
...
+        if (handle === 'me') {
+            accountFollows = await accountFollowsView.getFollowsByAccount(
+                siteDefaultAccount,
+                type,
+                Number.parseInt(next || '0'),
+                siteDefaultAccount,
+            );
+        } else {
+            const ctx = fedifyContextFactory.getFedifyContext();
+            const apId = await lookupAPIdByHandle(ctx, handle);
```
vershwal, line 155: I didn't move this lookup to the view because I want the account 
`const account = await accountRepository.getByApId(new URL(apId));` and this account comes from `accountRepository`. Didn't want to duplicate the whole logic of `accountRepository.getByApId` in the view.

PR #520, opened 2025-04-14
Added custom webfinger handler

mike182uk:
ref https://linear.app/ghost/issue/AP-963

Added a custom `GET /.well-known/webfinger` handler that will ensure that a resource on a `www.` domain can be referenced without the `www.` (i.e a webfinger lookup can be done on `@mike@mike.blog` if `mike.blog` is hosted at `www.mike.blog`)


src/http/handler/webfinger.ts
```diff
@@ -0,0 +1,87 @@
...
+        }
+
+        let site: Site | null = null;
+
+        site = await siteService.getSiteByHost(resourceHost);
+        if (site) {
+            // If a site is found for the resource host, we don't need to
+            // do any custom handling, we can just defer to the default
+            // webfinger implementation
+            return next();
```
allouis, line 44: I _think_ we do want to override here, at least sometimes

I think we need/want the `subject` value to always be `username@site.com` without the www. like below:

| resource | site url is www.site.com | site url is site.com |
| -------- | ------------- | -- |
| `acct:username@www.site.com`  | `username@site.com` | `username@site.com` |
| `acct:username@site.com` | `username@site.com` | `username@site.com` |

But I think by forwarding to the default implementation we get

| resource | site url is www.site.com | site url is site.com |
| -------- | ------------- | -- |
| `acct:username@www.site.com`  | `username@www.site.com` | `username@site.com` |
| `acct:username@site.com` | `username@site.com` | `username@site.com` |

allouis: I think it's gonna be a pain in the arse but we should add some cucumber tests for the functionality, because that way if we do forward to fedify still, we can catch any regressions from upgrades

mike182uk: ok cool, so if we always want `subject` to contain no `www` then we shouldn't forward onto the default handler for this case. Will update 👍

mike182uk: > I think it's gonna be a pain in the arse 

Your assessment is correct 😅

I'll look into getting some added though, I agree it would be nice to get some coverage at a higher level

PR #524, opened 2025-04-15
Added endpoint for updating an account information

vershwal:
Ref https://linear.app/ghost/issue/AP-1076/


src/http/api/account.ts
```diff
@@ -361,3 +370,30 @@ export function createGetAccountLikedPostsHandler(
         );
     };
 }
+
+/**
+ * Create a handler to handle a request for an account update
+ */
+export function createUpdateAccountHandler(
+    accountRepository: KnexAccountRepository,
```
allouis, line 378: Just a reminder that we should be passing the service instead of the repository here, so that each layer only speaks to the layer below

allouis (commented):
Just one comment

src/http/api/account.ts
vershwal: Fixed here : https://github.com/TryGhost/ActivityPub/pull/525/files

PR #527, opened 2025-04-15
Migrating modification of Accounts to Entity layer

allouis:
ref https://linear.app/ghost/issue/AP-1071

These are some building blocks for updating an Accounts profile fields

### DIFF
--- a/src/account/account-updated.event.ts
+++ b/src/account/account-updated.event.ts
@@ -1,0 +1,13 @@
+import type { Account } from './account.entity';
+
+export class AccountUpdatedEvent {
+    static getName(): string {
+        return 'account.updated.event';
+    }
+
+    constructor(private readonly account: Account) {}
+
+    getAccount(): Account {
+        return this.account;
+    }
+}
--- a/src/account/account.entity.ts
+++ b/src/account/account.entity.ts
@@ -22,25 +22,44 @@ export type AccountSite = {
     host: string;
 };
 
+export interface ProfileUpdateParams {
+    name?: string | null;
+    bio?: string | null;
+    avatarUrl?: URL | null;
+    bannerImageUrl?: URL | null;
+}
+
 export class Account extends BaseEntity {
     public readonly uuid: string;
     public readonly url: URL;
     public readonly apId: URL;
     public readonly apFollowers: URL;
+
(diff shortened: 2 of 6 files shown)


src/account/account-updated.event.ts
```diff
@@ -0,0 +1,13 @@
+import type { Account } from './account.entity';
+
+export class AccountUpdatedEvent {
+    static getName(): string {
+        return 'account.updated.event';
```
vershwal, line 5: This could just be `account.updated`, but I think it would contradict the current event. Maybe something to clean up later?

src/account/account.entity.ts
```diff
@@ -63,6 +82,40 @@ export class Account extends BaseEntity {
...
+
+    get avatarUrl(): URL | null {
+        return this._avatarUrl;
+    }
+
+    get bannerImageUrl(): URL | null {
+        return this._bannerImageUrl;
+    }
+
+    updateProfile(params: ProfileUpdateParams): void {
```
vershwal, line 101: Do we need to add `username` here? Or maybe that's coming as part of other changes(custom handles)?

src/account/account.repository.knex.ts
```diff
@@ -11,6 +12,28 @@ export class KnexAccountRepository {
         private readonly events: AsyncEvents,
     ) {}
 
+    async save(account: Account): Promise<void> {
+        if (account.isNew) {
+            throw new Error(
+                'Saving of new Accounts has not been implemented yet.',
```
vershwal, line 18: We can rename this method to `update` and use it only for this purpose, and have a different implementation for `save`. My thinking is that this method is updating the account and sending update event.

allouis: Our repository pattern uses a single save even for both update and saves, this means that callers of the code don't need to know if the entity they have a reference to is new or not.

We could change this if you think it's worth it but I think it needs to be part of a larger discussion and the PostRepository would need to be updated too

vershwal: In that case, it’s fine. I get that the single `save` method simplifies things in `PostRepository`, we can keep the same patter here.

vershwal (approved):

PR #528, opened 2025-04-15
Refactored `getPostsByRemoteLookUp` to return the `Result` type

vershwal:
Ref https://linear.app/ghost/issue/AP-1075

### DIFF
--- a/src/http/api/account.ts
+++ b/src/http/api/account.ts
@@ -4,6 +4,7 @@ import type { KnexAccountRepository } from 'account/account.repository.knex';
 import type { AccountService } from 'account/account.service';
 import type { FedifyContextFactory } from 'activitypub/fedify-context.factory';
 import type { AppContext, ContextData } from 'app';
+import { getError, getValue, isError } from 'core/result';
 import { isHandle } from 'helpers/activitypub/actor';
 import { lookupAPIdByHandle } from 'lookup-helpers';
 import type { GetProfileDataResult, PostService } from 'post/post.service';
@@ -296,8 +297,34 @@ export function createGetAccountPostsHandler(
                 if (postResult instanceof Error) {
                     throw postResult;
                 }
-                result.results = postResult.results;
-                result.nextCursor = postResult.nextCursor;
+                if (isError(postResult)) {
+                    const error = getError(postResult);
+                    switch (error) {
+                        case 'invalid-handle':
+                            logger.error(`Invalid handle ${handle}`);
+                            return new Response(null, { status: 400 });
+                        case 'missing-default-account':
+                            logger.error('Missing default account');
+                            return new Response(null, { status: 400 });
+                        case 'invalid-next-parameter':
+                            logger.error('Invalid next parameter');
+                            return new Response(null, { status: 400 });
+                        case 'actor-not-found':
+                            logger.error(`Actor not found for ${handle}`);
+                            return new Response(null, { status: 400 });
+                        case 'error-getting-outbox':
+                            logger.error(`Error getting outbox for ${handle}`);
+                            return new Response(null, { status: 500 });
+                        case 'no-page-found':
+                            logger.error(
+                                `No page found in outbox for ${handle}`,
+                            );
+                            return new Response(null, { status: 500 });
+                    }
+                }
+                const posts = getValue(postResult);
+                result.results = posts.results;
+                result.nextCursor = posts.nextCursor;
             }
         } catch (error) {
             logger.error(`Error getting posts for ${handle}: {error}`, {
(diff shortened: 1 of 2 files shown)


src/post/post.service.ts
```diff
@@ -582,23 +591,23 @@ export class PostService {
...
 
         if (!defaultAccount || !defaultAccount.id) {
-            throw new Error('Default account not found');
+            return error('missing-default-account');
         }
 
         // If the next parameter is not a valid URI, return early
         if (next !== '' && !isUri(next)) {
-            throw new Error('Invalid next parameter');
+            return error('invalid-next-parameter');
```
allouis, line 603: Can we fix this at the type level to require passing a `URL` instead of a `string`?

src/post/post.service.ts
```diff
@@ -582,23 +591,23 @@ export class PostService {
 
         // If the provided handle is invalid, return early
         if (!isHandle(handle)) {
-            throw new Error('Invalid handle');
+            return error('invalid-handle');
         }
 
         if (!defaultAccount || !defaultAccount.id) {
```
allouis, line 597: I think we can simplify this by passing the `accountId` and `accountApId` in for now? Because we know that in this flow this branch never happens

This can become simpler when the migration of this to the View pattern is done

src/post/post.service.ts
```diff
@@ -582,23 +591,23 @@ export class PostService {
...
-            throw new Error('Invalid next parameter');
+            return error('invalid-next-parameter');
         }
 
         // Lookup actor by handle
         const actor = await lookupObject(handle, { documentLoader });
 
         if (!isActor(actor)) {
-            throw new Error('Actor not found');
+            return error('actor-not-found');
```
allouis, line 610: I think this should be `not-an-actor`?

src/post/post.service.ts
```diff
@@ -582,23 +591,23 @@ export class PostService {
 
         // If the provided handle is invalid, return early
         if (!isHandle(handle)) {
-            throw new Error('Invalid handle');
+            return error('invalid-handle');
```
allouis, line 594: We already validate `handle` in the HTTP handler - why do we need to validate it again? Can we pass the `apId` directly to this method instead?

src/http/api/account.ts
```diff
@@ -304,8 +305,34 @@ export function createGetAccountPostsHandler(
...
+                            logger.error(`Actor not found for ${handle}`);
+                            return new Response(null, { status: 400 });
+                        case 'error-getting-outbox':
+                            logger.error(`Error getting outbox for ${handle}`);
+                            return new Response(null, { status: 500 });
+                        case 'no-page-found':
+                            logger.error(
+                                `No page found in outbox for ${handle}`,
+                            );
+                            return new Response(null, { status: 500 });
```
allouis, line 330: This shouldn't be a 500 error, I would expect that we return an empty posts list with a `null` next param

src/http/api/account.ts
```diff
@@ -304,8 +305,34 @@ export function createGetAccountPostsHandler(
...
+                            return new Response(null, { status: 400 });
+                        case 'missing-default-account':
+                            logger.error('Missing default account');
+                            return new Response(null, { status: 400 });
+                        case 'invalid-next-parameter':
+                            logger.error('Invalid next parameter');
+                            return new Response(null, { status: 400 });
+                        case 'actor-not-found':
+                            logger.error(`Actor not found for ${handle}`);
+                            return new Response(null, { status: 400 });
```
allouis, line 322: I think 404 is better suited for a not found issue

src/http/api/account.ts
```diff
@@ -304,8 +305,34 @@ export function createGetAccountPostsHandler(
...
+                        case 'error-getting-outbox':
+                            logger.error(`Error getting outbox for ${handle}`);
+                            return new Response(null, { status: 500 });
```
allouis, line 325: Same here about 500 - we should never explicitly return a 500, they are for unexpected errors only. I think this should probably be an empty posts list too!

src/http/api/account.ts
```diff
@@ -304,8 +305,34 @@ export function createGetAccountPostsHandler(
...
+                        case 'no-page-found':
+                            logger.error(
+                                `No page found in outbox for ${handle}`,
+                            );
+                            return new Response(null, { status: 500 });
+                    }
```
allouis, line 331: Can we add an "exhaustive check" here?

```ts
default:
    const exhaustiveCheck: never = error;
    throw new Error(`Unhandled error ${exhaustiveCheck}`);
```

This ensures that all error types are handled if we ever add new ones! I'll look at getting a helper added to make this simpler

src/post/post.service.ts
vershwal: next can be empty string(`''`) in case when there is no next page, and we handle it differently. I am gonna leave it like this for now and try to come up with a cleaner approach while refactoring it to viewer.

PR #533, opened 2025-04-16
Added Google Cloud Storage (GCS) integration for image uploads

vershwal:
Ref https://linear.app/ghost/issue/AP-1062/

Summary of changes:-

- Integrated Google Cloud Storage (GCS) for handling image uploads.
- Added a new Docker Compose service `fake-gcs` to simulate GCS locally. 
- Configured environment variables in local/test services to connect with the emulator.
- Added @google-cloud/storage as a new dependency.
- Validation on boot:
-- Initializes the GCS client
-- Validates and creates the bucket (if running with the emulator)
- Introduced a new POST API: `./ghost/activitypub/upload/image`
-- Validates file type and size
-- Uploads the image to the GCS bucket with a unique path
-- Returns a public URL to access the image
- Added unit tests for the upload logic.
- Added new Cucumber step definitions and a test scenario to verify the upload flow end-to-end.


src/storage/gcloud-storage/storage.ts
```diff
@@ -0,0 +1,57 @@
...
+        } catch (err) {
+            logger.error('Error uploading file:', err);
+            return new Response(JSON.stringify({}), {
+                status: 200,
+            });
+        }
+    };
+
+    function getStoragePath(account: Account) {
+        return `images/${account.uuid}/${uuidv4()}`;
```
sagzy, line 55: The path starts with `images/`, whereas the rest the API endpoint is media type agnostic.

Think either:
1. We have an API endpoint that is specific to image uploads (which will also allow us to reject any files that are not supported atm)
2. Or create a path helper that takes into account the media type

sagzy: Also think we're missing the file extension

sagzy (commented):
Generally, thinking whether an API endpoint that is specific to image upload would be better:
- we don't intend to support other file type uploads until GA
- allows for a validation on the type of media file (reject any file that is not an image)

src/storage/gcloud-storage/storage.ts
```diff
@@ -0,0 +1,57 @@
...
+    /**
+     * Handle an upload to GCloud Storage bucket
+     */
+    return async function handleUpload(ctx: Context) {
+        const logger = ctx.get('logger');
+        const storage = new Storage();
+        const bucketName = process.env.GCP_BUCKET_NAME;
+
+        if (!bucketName) {
+            return new Response('Bucket name is not configured', {
```
allouis, line 17: I think this should error at application boot time, not on a request!

src/storage/gcloud-storage/storage.ts
```diff
@@ -0,0 +1,57 @@
...
+                status: 400,
+            });
+        }
+
+        const account = await accountService.getAccountForSite(ctx.get('site'));
+        const storagePath = getStoragePath(account);
+
```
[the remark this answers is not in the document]
allouis: I think CodeRabbit is right here, we should validate the existence on boot, like we do for pubsub!

src/storage/gcloud-storage/storage.ts
```diff
@@ -0,0 +1,57 @@
...
+            return new Response(JSON.stringify({ fileUrl }), {
+                status: 200,
+            });
+        } catch (err) {
+            logger.error('Error uploading file:', err);
+            return new Response(JSON.stringify({}), {
+                status: 200,
+            });
+        }
```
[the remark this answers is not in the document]
allouis: Same here, I think we wanna error if something goes wrong

src/app.ts
```diff
@@ -188,6 +190,48 @@ export type ContextData = {
...
+            private_key: 'not-a-real-key',
+        },
+    });
+
+    const bucket = storage.bucket(bucketName);
+
+    const [exists] = await bucket.exists();
+    if (!exists) {
+        logging.info('Creating GCP bucket in fake-gcs');
+        await bucket.create(); //Create a bucket in fake-gcs
```
vershwal, line 231: fake-gcs-server starts with no buckets, so we create it in dev/test if it doesn't exist

src/storage/gcloud-storage/storage.ts
```diff
@@ -0,0 +1,111 @@
...
+import { type Result, error, getError, isError, ok } from 'core/result';
+import type { Context } from 'hono';
+import { v4 as uuidv4 } from 'uuid';
+
+const ALLOWED_IMAGE_TYPES = [
+    'image/jpg',
+    'image/jpeg',
+    'image/png',
+    'image/webp',
+];
```
vershwal, line 13: We can also allow image/gif or image/svg? But doesn't seem necessary for profile/cover images

src/storage/gcloud-storage/storage.ts
```diff
@@ -0,0 +1,111 @@
...
+                            status: 415,
+                        },
+                    );
+            }
+        }
+
+        const bucketName = process.env.GCP_BUCKET_NAME;
+        const emulatorHost = process.env.GCP_STORAGE_EMULATOR_HOST; // This is for dev and testing environments
+
+        if (!bucketName) {
```
vershwal, line 56: I already have this check on boot, but TypeScript complains if I don’t check it again here.
The other option is to use `const bucketName = process.env.GCP_BUCKET_NAME!`;

src/storage/gcloud-storage/storage.ts
```diff
@@ -0,0 +1,111 @@
...
+            return error('file-too-large');
+        }
+
+        if (!ALLOWED_IMAGE_TYPES.includes(file.type)) {
+            return error('file-type-not-supported');
+        }
+
+        return ok(true);
+    }
+}
```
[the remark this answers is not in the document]
vershwal: I think this is a good suggestion, thanks

src/storage/gcloud-storage/storage.ts
```diff
@@ -0,0 +1,106 @@
...
+export type FileValidationError = 'file-too-large' | 'file-type-not-supported';
+
+export function createStorageHandler(
+    accountService: AccountService,
+    bucket: Bucket,
+) {
+    /**
+     * Handle an upload to GCloud Storage bucket
+     */
+    return async function handleUpload(ctx: Context) {
```
allouis, line 24: This is a HTTP handler right? I think this should live in the `src/http` directory, and then use the storage service as a dependency

src/app.ts
allouis: This shouldn't be in the application code if possible - it should be part of the docker stack if we can

We could introduce a script which runs as the container command, and it will execute when the container starts and ensure anything we need

Or we could commit to the repo a pre-populated data directory, and mount that as the volume 🤔

src/app.ts
```diff
@@ -188,6 +190,49 @@ export type ContextData = {
 
 const fedifyKv = await KnexKvStore.create(client, 'key_value');
 
+let storage = new Storage();
```
allouis, line 193: I think all of this logic should be pulled out of `app.ts` like how we have the PubSub initialisation stuff separate. This could be part of the constructor in init method of a GCPStorageService or something?

docker-compose.yml
```diff
@@ -29,6 +31,8 @@ services:
         condition: service_healthy
       pubsub:
         condition: service_healthy
+      fake-gcs:
+        condition: service_started
```
allouis, line 35: Why not service_healthy out of interest?

src/storage/gcloud-storage/storage.ts
vershwal: Done in https://github.com/TryGhost/ActivityPub/pull/533/commits/06a1d0a69cccc4338b88638867afdd7091222c8f

src/app.ts
vershwal: Done in https://github.com/TryGhost/ActivityPub/pull/533/commits/3db475b46662793f806d42d721b704b7234b9740

src/http/api/storage.ts
```diff
@@ -0,0 +1,105 @@
...
+) {
+    /**
+     * Handle an upload to GCloud Storage bucket
+     */
+    return async function handleUpload(ctx: Context) {
+        const logger = ctx.get('logger');
+        const bucket = storageService.getBucket();
+
+        const formData = await ctx.req.formData();
+        const file = formData.get('file') as File;
```
allouis, line 29: I think this is a dangerous assertion - we don't know that `file` is the type of `File`! So the rest of our code becomes unsafe - we should check that is is a `File` with an instanceof check I think

src/storage/gcloud-storage/storage.ts
vershwal: Done!

src/http/api/storage.ts
```diff
@@ -0,0 +1,105 @@
...
+                        },
+                    );
+            }
+        }
+
+        const emulatorHost = process.env.GCP_STORAGE_EMULATOR_HOST; // This is for dev and testing environments
+        const account = await accountService.getAccountForSite(ctx.get('site'));
+        const storagePath = getStoragePath(account, file.name);
+
+        await bucket.file(storagePath).save(file.stream(), {
```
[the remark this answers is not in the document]
vershwal: Done!

src/http/api/storage.ts
```diff
@@ -0,0 +1,48 @@
...
+                    logger.error(`File is too large: ${file.size} bytes`);
+                    return new Response('File is too large', { status: 413 });
+                case 'file-type-not-supported':
+                    logger.error(`File type ${file.type} is not supported`);
+                    return new Response(
+                        `File type ${file.type} is not supported`,
+                        { status: 415 },
+                    );
+            }
+        }
```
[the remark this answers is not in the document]
vershwal: ooh, that's a good one. Forgot the default. Thanks

features/step_definitions/stepdefs.js
```diff
@@ -699,6 +699,26 @@ When(
...
+        this.response = await fetchActivityPub(
+            `http://fake-ghost-activitypub.test${path}`,
+            {
+                method: method || 'post',
+                body: formData,
+            },
+        );
+    },
+);
+
```
vershwal, line 721: I can make this tests in a separate file like we discussed yesterday.

src/app.ts
```diff
@@ -187,6 +189,21 @@ export type ContextData = {
...
 const fedifyKv = await KnexKvStore.create(client, 'key_value');
 
+logging.info('Initialising GCP storage service');
+const gcpStorageService = new GCPStorageService();
+
+try {
+    logging.info('Initialising GCP storage service');
+    await gcpStorageService.init();
+    logging.info('GCP storage service initialised');
+    logging.info('Adding latest code for GCP storage service');
```
vershwal, line 199: This is some extra logging for testing on stg, will remove once testing is complete

allouis (approved):
One small thing, I think we should pass the config in, rather than read from the environment variables

PR #535, opened 2025-04-16
Implemented account profile update

mike182uk:
ref https://linear.app/ghost/issue/AP-1076

Implemented the logic needed for updating an account's profile data via the API

### DIFF
--- a/features/step_definitions/stepdefs.js
+++ b/features/step_definitions/stepdefs.js
@@ -683,7 +683,7 @@ When(
 );
 
 When(
-    /^an authenticated (\"(post|put)\"\s)?request is made to "(.*)" with data:$/,
+    /^an authenticated (\"(post|put)\"\s)?request is made to "(.*)" with the data:$/,
     async function (method, path, data) {
         this.response = await fetchActivityPub(
             `http://fake-ghost-activitypub.test${path}`,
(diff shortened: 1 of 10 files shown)


features/step_definitions/stepdefs.js
```diff
@@ -1511,7 +1511,6 @@ Then('the request is accepted', async function () {
 });
 
 Then('the request is accepted with a {int}', function (statusCode) {
-    assert(this.response.ok);
```
mike182uk, line 1514: I removed this as its not helpful, the assertion below is more useful

vershwal (approved):

src/dispatchers.ts
```diff
@@ -81,7 +81,6 @@ export const keypairDispatcher = (
         ctx: Context<ContextData>,
         handle: string,
     ) {
-        if (handle !== ACTOR_DEFAULT_HANDLE) return [];
```
allouis, line 84: Should we do this for the Actor dispatcher too?

allouis (commented):
Nice, great tests

src/dispatchers.ts
mike182uk: Sure, updated in https://github.com/TryGhost/ActivityPub/pull/535/commits/dc5de38ef4bc2dd57cf3b51f970dd6b4a2101ed8

PR #543, opened 2025-04-17
Improved boot time

rmgpinto:
closes https://linear.app/ghost/issue/AP-1081

- Improved boot time by bundling the app in a single file. This avoids reading multiple files from disk, which can slow down the booting process.
- Since top level `await` are not permitted, `src/app.ts` is now wrapped in a main() function.

### DIFF
--- a/.github/workflows/cicd.yml
+++ b/.github/workflows/cicd.yml
@@ -102,8 +102,8 @@ jobs:
           load: true
           tags: ${{ steps.migrations-docker-metadata.outputs.tags }}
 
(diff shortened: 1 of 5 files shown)


rmgpinto:
@allouis tree-shaking is true by default, format `iife` doesn't seem to work. I guess because we have exports and top level awaits in the src/app.ts. The code isn't ideal because I had to extract the exports outside the main function.
But this is the only way I got it to bundle correctly to decrease boot time by 4 seconds.

PR #544, opened 2025-04-17
Initial refactor

vershwal:


src/http/api/views/account.posts.view.ts
```diff
@@ -0,0 +1,667 @@
...
+export class AccountPostsView {
+    constructor(
+        private readonly db: Knex,
+        private readonly fedifyContextFactory: FedifyContextFactory,
+    ) {}
+
+    async getPostsByHandle(
+        handle: string,
+        account: Account,
+        defaultAccount: Account,
```
allouis, line 87: I don't think `defaultAccount` is a good name here - it's not clear what we're doing with it - I think something that represents this is the "context" we're requesting with

src/http/api/views/account.posts.view.ts
```diff
@@ -0,0 +1,667 @@
...
+                this.mapToPostDTO(result, accountId),
+            ),
+            nextCursor: hasMore ? lastResult.published_date : null,
+        };
+    }
+
+    async getPostsByRemoteLookUp(
+        defaultAccountId: number,
+        defaultAccountApId: URL,
+        handle: string,
```
allouis, line 290: We already have the `apId` of the remote account - I don't think we need to pass `handle` here!

src/http/api/views/account.posts.view.ts
```diff
@@ -0,0 +1,667 @@
...
+        next: string,
+    ): Promise<Result<AccountPosts, GetPostsError>> {
+        const context = this.fedifyContextFactory.getFedifyContext();
+
+        const documentLoader = await context.getDocumentLoader({
+            handle: 'index',
+        });
+
+        // If the next parameter is not a valid URI, return early
+        if (next !== '' && !isUri(next)) {
```
allouis, line 300: We can use the `URL` type instead of checking strings here

src/http/api/account.ts
```diff
@@ -234,119 +236,91 @@ export function createGetAccountPostsHandler(
...
-            if (account?.isInternal && account.id) {
-                const postResult = await postService.getPostsByAccount(
-                    account.id,
-                    defaultAccount.id,
-                    params.limit,
-                    params.cursor,
-                );
+            const account = await accountRepository.getByApId(new URL(apId));
+            if (!account) {
+                return new Response(`Account not found for handle: ${handle}`, {
```
allouis, line 274: Does this mean that we error for all remote accounts that we don't have a record of in our db? Am I misunderstanding this?

src/http/api/views/account.posts.view.ts
```diff
@@ -0,0 +1,667 @@
...
+        defaultAccount: Account,
+        limit: number,
+        cursor: string | null,
+    ): Promise<Result<AccountPosts, GetPostsError>> {
+        if (!account.id || !defaultAccount.id) {
+            throw new Error('Account or default account not found');
+        }
+
+        //If we found the account in our db and it's an internal account, do an internal lookup
+        if (account?.isInternal) {
```
allouis, line 96: I don't think `account` is optional here

src/http/api/views/account.posts.view.ts
```diff
@@ -0,0 +1,667 @@
...
+                'posts.image_url as post_image_url',
+                'posts.published_at as post_published_at',
+                'posts.like_count as post_like_count',
+                this.db.raw('1 AS post_liked_by_current_user'), // Since we are selecting from `likes`, this is always 1
+                'posts.reply_count as post_reply_count',
+                'posts.reading_time_minutes as post_reading_time_minutes',
+                'posts.attachments as post_attachments',
+                'posts.repost_count as post_repost_count',
+                this.db.raw(
+                    `CASE
```
allouis, line 474: This indent looks a bit off!

src/http/api/views/account.posts.view.ts
vershwal: Sorry if I am being blind, but we don't have `apId` of the remote account, the `defaultAccountApId` is the appId of the current user. WE need this handle to look for the remote actor and fetch their outbox.

vershwal: Maybe I need to fix the naming `defaultAccountApId` so that it's clear

src/http/api/account.ts
vershwal: hmm..this doesn't sound right. I'll check. Thanks for pointing it out.

src/http/api/views/account.posts.view.ts
allouis: We have the `apId` in the http handler where we do
```
const apId = await lookupAPIdByHandle(ctx, handle);
```

And we could pass that in instead of the handle

vershwal: oooh...Thanks a lot 🙏

src/http/api/views/account.posts.view.ts
vershwal: I am not good at renaming, I used `currentContextAccount`, but if that's not good please suggest something better 😬

src/http/api/views/account.posts.view.ts
vershwal: Done!

src/http/api/views/account.posts.view.ts
vershwal: Catch here is that next can be an empty string `''` or null and both are totally valid. I used URL type but had to use `new URL('about:blank')` [here](https://github.com/TryGhost/ActivityPub/pull/544/files#diff-5a5f4ba0eac8fe9efad4e5cca046af80781f82b3cd56b9da95bce4b1e8d9dcabR116)  for empty next.

src/http/api/account.ts
vershwal: Done. Removed this check

src/http/api/views/account.posts.view.ts
vershwal: Fixed it!

src/http/api/views/account.posts.view.ts
vershwal: Now it's optional. So I kept it like this

src/http/api/views/account.posts.view.ts
```diff
@@ -0,0 +1,666 @@
...
+    ) {}
+
+    async getPostsByApId(
+        apId: URL,
+        account: Account | null,
+        currentContextAccount: Account,
+        limit: number,
+        cursor: string | null,
+    ): Promise<Result<AccountPosts, GetPostsError>> {
+        if (!currentContextAccount.id) {
```
allouis, line 90: We know that these flows aren't happened - so we should try and stop them at the type level

e.g we can do something like

```
type PersistedAccount = Account & {id: number}
```

My concern here is that we have a bunch of logic to handle cases that we know can't happen - we just need to update the code so the compiler knows they can't happen too!

src/http/api/views/account.posts.view.ts
```diff
@@ -0,0 +1,666 @@
...
+            throw new Error('Current context account id not found');
```
allouis, line 91: We should try to use a Result where possible - in this case it's an invalid code path, but I would prefer to fix this at the type level like I mentioned

src/http/api/views/account.posts.view.ts
```diff
@@ -0,0 +1,666 @@
...
+        }
+
+        //If we found the account in our db and it's an internal account, do an internal lookup
+        if (account?.isInternal) {
+            if (!account.id) {
```
allouis, line 96: Same as above RE: PersistedAccount

src/http/api/views/account.posts.view.ts
```diff
@@ -0,0 +1,666 @@
...
+                ),
+            );
+        }
+
+        //Otherwise, do a remote lookup to fetch the posts
+        return this.getPostsByRemoteLookUp(
+            currentContextAccount.id,
+            currentContextAccount.apId,
+            apId,
+            cursor ? new URL(cursor) : new URL('about:blank'),
```
allouis, line 116: This parsing of the cursor to the correct type I would expect to happen at the controller level, as this is the place where we convert the inputs of a HTTP request into objects usable by our domain

I would also expect `cursor` to be nullable here - rather than using `'about:blank'`

src/http/api/views/account.posts.view.ts
```diff
@@ -0,0 +1,666 @@
...
+                if (!(item instanceof Activity)) {
+                    continue;
+                }
+
+                const object = await item.getObject();
+                const attributedTo = await object?.getAttribution();
+
+                const activity = (await item.toJsonLd({
+                    format: 'compact',
+                    // TODO: Clean up the any type
```
allouis, line 369: We shouldn't introduce more of these comments, we should be cleaning up as we go! `unknown` is the best type to use when we don't know what we're getting back - it means we have to do a few extra checks, but that's good because we're forced to be type-safe and avoid unexpected runtime errors

src/http/api/views/account.posts.view.ts
```diff
@@ -0,0 +1,666 @@
...
+
+        // Return result
+        for await (const item of page.getItems()) {
+            try {
+                if (!(item instanceof Activity)) {
+                    continue;
+                }
+
+                const object = await item.getObject();
+                const attributedTo = await object?.getAttribution();
```
allouis, line 365: This isn't used anywhere - not sure why linting isn't picking it up though

vershwal: Removed it

src/http/api/views/account.posts.view.ts
vershwal: Cleaned this up.

src/http/api/views/account.posts.view.ts
vershwal: Done. Added `PersistedAccount` type and used it.

src/http/api/views/account.posts.view.ts
vershwal: It makes sense to have the correct type in the controller, I kept the cursor as a string to maintain type consistency. In the controller, I didn't change it to a URL type because the local lookup (`getPostsByAccount`) requires the cursor to be a string. For `getPostsByRemoteLookUp`, I adjusted it to a URL where needed.

allouis (approved):

PR #549, opened 2025-04-22
Added moderation service

mike182uk:

Added moderation service to encapsulate the logic for ensuring content from blocked users is not shown to the blocking account

This also adds a new test fixture manager to simplify the creation of test data for integration tests

### DIFF
--- a/src/test/fixtures.ts
+++ b/src/test/fixtures.ts
@@ -23,7 +23,7 @@ export class FixtureManager {
     async createInternalAccount(
         site?: Site | null,
         host = faker.internet.domainName(),
-    ) {
+    ): Promise<[Account, Site]> {
         let _site: Site;
         let _account: Account;
 
@@ -54,10 +54,7 @@ export class FixtureManager {
             _account = accountByApId;
         }
 
-        return {
-            site: _site,
-            account: _account,
-        };
+        return [_account, _site];
     }
 
     async createPost(
--- a/src/test/fixtures.ts
+++ b/src/test/fixtures.ts
@@ -1,0 +1,134 @@
+import { faker } from '@faker-js/faker';
+import type { Knex } from 'knex';
+
+import type { Account } from 'account/account.entity';
+import { KnexAccountRepository } from 'account/account.repository.knex';
+import { AccountService } from 'account/account.service';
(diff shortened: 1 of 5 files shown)


src/test/fixtures.ts
```diff
@@ -0,0 +1,134 @@
...
+            if (!accountByApId) {
+                throw new Error(
+                    `Account not found with ap_id ${accountApId} for site ${_site.id}`,
+                );
+            }
+
+            _account = accountByApId;
+        }
+
+        return {
```
allouis, line 57: This is gonna sound a bit mental, but what if we returned `[Account, Site]` instead of `{account: Account, site: Site}`?

My thinking is that the usage becomes cleaner in the test files, e.g.

```ts
const [
    { account },
    { account: blockedAccount1 },
    { account: blockedAccount2 },
] = await Promise.all([
    fixtureManager.createInternalAccount(null, 'example.com'),
    fixtureManager.createInternalAccount(null, 'blocked1.com'),
    fixtureManager.createInternalAccount(null, 'blocked2.com'),
]);
```

```ts
const [
    [account],
    [blockedAccount1],
    [blockedAccount2],
] = await Promise.all([
    fixtureManager.createInternalAccount(null, 'example.com'),
    fixtureManager.createInternalAccount(null, 'blocked1.com'),
    fixtureManager.createInternalAccount(null, 'blocked2.com'),
]);
```

I mean ideally we could just return `Account` - but I see that we probably need access to the `Site` sometimes? 

Although maybe in that case you have to do it like:

```ts
const site = await fixtureManager.createSite('example.com');
const account = await fixtureManager.createInternalAccount(site);
```

What do you think?

allouis (commented):
Love the fixture manager! One small comment about the return values - but looks good

src/test/fixtures.ts
mike182uk: Yeh nice idea, tuple would work well here, and because its typed, we essentially get the same labelling as we would with the object

Yep returning both as its more than likely you'll want the site for other things, so easier to return at this point (and allows creation of a user without upfront creation of a site)

allouis (approved):

PR #550, opened 2025-04-22
Added handling for feed post removal when account blocked

mike182uk:

Added handling for feed post removal when an account is blocked:

- When an account is blocked remove any posts by the blocked account from the blocking users feed
- When an account is blocked remove any reposts by the blocked account from the blocking users feed

### DIFF
--- a/src/account/account-blocked.event.ts
+++ b/src/account/account-blocked.event.ts
@@ -2,16 +2,16 @@ import type { Account } from 'account/account.entity';
 
 export class AccountBlockedEvent {
     constructor(
-        private readonly actor: Account,
-        private readonly blocked: Account,
+        private readonly account: Account,
+        private readonly blocker: Account,
     ) {}
 
-    getActor(): Account {
-        return this.actor;
+    getAccount(): Account {
+        return this.account;
     }
 
-    getBlocked(): Account {
-        return this.blocked;
+    getBlocker(): Account {
+        return this.blocker;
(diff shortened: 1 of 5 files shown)


src/account/account-blocked.event.ts
```diff
@@ -0,0 +1,20 @@
+import type { Account } from 'account/account.entity';
+
+export class AccountBlockedEvent {
+    constructor(
+        private readonly actor: Account,
+        private readonly blocked: Account,
+    ) {}
+
+    getActor(): Account {
```
allouis, line 9: I don't think this is a good name given Actor means something specific in ActivityPub

allouis: `getBlocker()`? It's a bit anti-dyslexic though 🙃

allouis (commented):
Comment about naming

src/account/account-blocked.event.ts
mike182uk: heh, i though I'd get some resistance on the naming of that 😅 My rationale was `this is the account that did the action`, but i get why this isn't ideal. Looking at `account.followed` we use `getAccount` to represent the account being followed, and `getFollower` to represent the account doing the follow. Maybe we do the same here? `getAccount` to represent the account being blocked and `getBlocker` to represent the account doing the block?

allouis (approved):

PR #551, opened 2025-04-23
Moved Create<Note> creation and publishing to FediverseBridge

vershwal:
Ref https://linear.app/ghost/issue/AP-1094

- Added `handlePostCreated` to `FediverseBridge` that handles the `PostCreatedEvent`
- Removed code for publishing note from `note.ts`
-- `FediverseBridge` will publish the note when `PostCreatedEvent` is received.
-- Create Note API will now return a `postDTO`
- Removed code for publishing article from `webhook.ts`
-- `FediverseBridge` will publish the article when `PostCreatedEvent` is received.
- Added `PostVisibility` enum to post.entity.ts
-- PostVisibility is used webhook.ts, earlier this enum was a part of `src/publishing/types.ts`. Moved it here before removing `src/publishing`
- Removed publishing service and related code
- Updated e2e tests
-- Create note API now returns a postDTO. Earlier this API was returning an Activity. Tests were expecting the old activity structure in the response, updated the tests to accept postDTO as response from the create note API.
- Added type check on attributedTo while adding to inbox
-- Caught this in a failing test. attributedTo can be a `string` and because of this the reply was not being added to our inbox.


features/support/activitypub.js
```diff
@@ -99,3 +121,33 @@ export async function waitForOutboxActivity(
...
+
+    if (options.delay > 0) {
+        await new Promise((resolve) => setTimeout(resolve, options.delay));
+    }
+
+    await waitForOutboxActivity(note, {
+        retryCount: options.retryCount + 1,
+        delay: options.delay + 500,
+    });
+}
```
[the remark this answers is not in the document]
vershwal: That's a great catch thanks!!

src/activitypub/fediverse-bridge.ts
```diff
@@ -36,6 +47,91 @@ export class FediverseBridge {
...
+        const post = event.getPost();
+        if (!post.author.isInternal) {
+            return;
+        }
+        const ctx = this.fedifyContextFactory.getFedifyContext();
+        let fedifyObject = null;
+
+        if (post.type === PostType.Note) {
+            if (post.inReplyTo) {
+                return;
```
vershwal, line 60: Reply activity is already sent in `reply.ts`, this was sending a duplicate activity.
I have added a PR for moving code for sending reply to Fediverse from reply.ts to FediverseBridge [here](https://github.com/TryGhost/ActivityPub/pull/636).  Kept it separate to avoid too many changes in this one as it's already a big PR.

src/activitypub/fediverse-bridge.ts
```diff
@@ -36,6 +47,88 @@ export class FediverseBridge {
...
+        const post = event.getPost();
+        if (!post.author.isInternal) {
+            return;
+        }
+        const ctx = this.fedifyContextFactory.getFedifyContext();
+        let fedifyObject = null;
+
+        if (post.type === PostType.Note) {
+            fedifyObject = new FedifyNote({
+                id: post.apId || ctx.getObjectUri(FedifyNote, { id: uuidv4() }),
```
allouis, line 63: When is `post.apId` not defined?

src/activitypub/fediverse-bridge.ts
```diff
@@ -36,6 +47,88 @@ export class FediverseBridge {
...
+            to: PUBLIC_COLLECTION,
+            cc: post.author.apFollowers,
+        });
+
+        await ctx.data.globaldb.set(
+            [createActivity.id!.href],
+            await createActivity.toJsonLd(),
+        );
+
+        if (fedifyObject?.id!.href) {
```
allouis, line 114: `fedifyObject` should always be defined - so we shouldn't need a `?` here, and we can get rid of the whole conditional!

If typescript tells us that it might not be defined, then we need to understand why and handle that case explicitly

src/http/api/note.ts
```diff
@@ -62,24 +60,16 @@ export async function handleCreateNote(
...
-            apId: post.apId,
-            imageUrl: post.imageUrl,
-        });
-    } catch (err) {
-        ctx.get('logger').error('Failed to publish note: {error}', {
-            error: err,
-        });
-    }
+    const postDTO = postToDTO(post, {
+        authoredByMe: post.author.id === account.id,
```
allouis, line 66: This is always true, right?

src/post/post.entity.ts
```diff
@@ -11,6 +11,28 @@ export enum PostType {
     Tombstone = 2,
 }
 
+/**
+ * Visibility of a post
+ */
+export enum PostVisibility {
```
allouis, line 17: I don't think this belongs here, it hasn't really got anything to do with our concept of a Post and is specific to a GhostPost - which is all handled in the webhook. As far as I can see this is only used to create the validation schema in zod? I would probably get rid of this entirely and just use zod to explicitly define the type for `visibility`

features/support/activitypub.js
```diff
@@ -70,6 +70,28 @@ export async function findInOutbox(activity) {
     return (outbox.orderedItems || []).find((item) => item.id === activity.id);
 }
 
+export async function findNoteInOutbox(note) {
```
allouis, line 73: This is really finding an object, right? I don't think it has to be specific to a note

features/step_definitions/note_steps.js
```diff
@@ -75,19 +76,52 @@ When(
         );
 
         if (this.response.ok) {
-            const activity = await this.response.clone().json();
-
-            this.activities[noteName] = activity;
-            this.objects[noteName] = activity.object;
+            const responseJson = await this.response.clone().json();
+            const post = responseJson.post;
+            this.posts[noteName] = post;
```
allouis, line 81: I'm not 100% sure about adding a new `this.posts` object :thinking: 

I feel like we could have just used `this.objects` - and either stored the postDTO (which may have been a bit buggy) or map the DTO to an ActivityPub object

That way we can get away with pretty much none of the tests changing 🤔 

We would update the "is in our Outbox" step to have something like

```ts
const activity = this.activities[name];
if (activity) return waitForActivityInOutbox(activity);
const object = this.objects[name];
if (object) return waitFroObjectInOutbox(object);
```

allouis (changes requested):
Overall these changes look good, but I've left some comments. Before we merge this we definitely need to smoke test it on staging.

I would love to see the snapshots from the publishing service reinstated and not modified - this will really prove that the refactor does not change functionality, and is just moving the code around.

vershwal:
Thanks for reviewing this @allouis I have updated the PR. 
I have used the existing snapshots for testing. Commit - https://github.com/TryGhost/ActivityPub/pull/551/commits/0c413eee21f52cd2ebe938a5f78095e5b97f6afc

The only change in the snapshot is that earlier we used to return `Actor` in `actor` and `attributedTo` fields, Now we are adding the id in these fields. Changes [here](https://github.com/TryGhost/ActivityPub/pull/551/files#diff-d39d2c709e19a3a8f4bef1ff091b8d62af44d3e51a2f15548075015fcbb51f35R24). This works but if we want to return the actor like we had before I can fetch the actor using apId and add it in the create activity. Do you think this change can have any impact?

features/step_definitions/note_steps.js
vershwal: This was a great suggestion. Got rid of a lot of test changes by using `this.objects` instead of `this.posts`. Updated tests: https://github.com/TryGhost/ActivityPub/pull/551/commits/ed7cda862e6931a298c4971328ed2316cbc4863e

features/support/activitypub.js
vershwal: Updated tests: https://github.com/TryGhost/ActivityPub/pull/551/commits/ed7cda862e6931a298c4971328ed2316cbc4863e

src/http/api/note.ts
vershwal: Correct! Fixed in https://github.com/TryGhost/ActivityPub/pull/551/commits/4a4a080f43440366183d9ed983a77132f8a21e9a

src/activitypub/fediverse-bridge.ts
vershwal: Fixed in https://github.com/TryGhost/ActivityPub/pull/551/commits/4a4a080f43440366183d9ed983a77132f8a21e9a

src/activitypub/fediverse-bridge.ts
vershwal: Fixed in https://github.com/TryGhost/ActivityPub/pull/551/commits/4a4a080f43440366183d9ed983a77132f8a21e9a

vershwal:
Tested this on my staging site (https://princi-vershwal-12.ghost.is/ghost/#)

- [ ] Note created on Ghost site is visible on my profile in mastodon. 
- [ ] Note created on Ghost is visible on the feed of my follower in Mastodon
- [ ] Post created on Ghost site is visible on my profile in mastodon. 
- [ ] Post created on Ghost is visible on the feed of my follower in Mastodon
- [ ] Replies are correctly visible and ordered on both Ghost and Mastodon.

allouis (approved):
This looks good! Let's make sure we have a plan for merging this alongside the client so we don't break things

PR #552, opened 2025-04-23
AP-1092 | 1 | Removed null check on account

vershwal:
Ref: https://linear.app/ghost/issue/AP-1092

### DIFF
--- a/src/http/api/account.ts
+++ b/src/http/api/account.ts
@@ -134,7 +134,7 @@ export function createGetAccountFollowsHandler(
                 return new Response(null, { status: 400 });
             }
 
-            const account = await accountRepository.getByApId(new URL(apId)) as PersistedAccount;
+            const account = await accountRepository.getByApId(new URL(apId));
 
             accountFollows = await accountFollowsView.getFollowsByHandle(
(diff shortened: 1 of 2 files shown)


src/http/api/account.ts
```diff
@@ -134,10 +134,7 @@ export function createGetAccountFollowsHandler(
                 return new Response(null, { status: 400 });
             }
 
-            const account = await accountRepository.getByApId(new URL(apId));
-            if (!account) {
-                return new Response(null, { status: 400 });
-            }
+            const account = await accountRepository.getByApId(new URL(apId)) as PersistedAccount;
```
allouis, line 137: We shouldn't cast to a `PersistedAccount` here - because the value can be `null`

allouis (changes requested):
We shouldn't be casting types without checking them

Also `accountFollows` can be undefined when we use it in the response here

allouis (approved):
Approving with the idea that we add tests for this, but we get the bug fix out now.

Can we make sure the commit message is something like "Fixed ..." rather than "Removed null check ..."

PR #557, opened 2025-04-23
AP-1092 | 3 | Handling non paginated follow lists

vershwal:
Ref https://linear.app/ghost/issue/AP-1092

We are having an issue with - Followers not loading on local setup for Ghost internal account
On prod we fetch the followers locally(from our db) but on local machine they are fetched from remote lookup. In remote lookup ghost's account have followers list in a non-paginated format.
This PR has code for handling non paginated follow lists


src/http/api/account.ts
```diff
@@ -164,15 +164,6 @@ export function createGetAccountFollowsHandler(
                             }),
                             { status: 200 },
                         );
-                    case 'no-page-found':
```
vershwal, line 167: Not needed any more

src/http/api/views/account.follows.view.ts
```diff
@@ -37,6 +36,48 @@ interface AccountRow {
...
+interface FollowsActor {
+    id: string;
+    name: string;
+    preferredUsername: string;
+    icon: {
+        url: string;
+    };
+}
+
+function isValidFollowsActor(obj: unknown): obj is FollowsActor {
```
vershwal, line 48: Kept this validation in a separate method as I had to do these checks at two places.

src/http/api/views/account.follows.view.ts
```diff
@@ -196,45 +322,14 @@ export class AccountFollowsView {
                 }
 
                 const followeeAccount = await this.db('accounts')
-                    .where('ap_id', actor.id?.toString() || '')
+                    .where('ap_id', followsActorObj.id?.toString() || '')
```
allouis, line 325: This is same as above re: `ap_id_hash`

src/http/api/views/account.follows.view.ts
```diff
@@ -141,10 +183,15 @@ export class AccountFollowsView {
             return error('not-an-actor');
         }
 
+        const followeeAccount = await this.db('accounts')
+            .where('ap_id', actor.id?.toString() || '')
```
allouis, line 187: This should be using `ap_id_hash` because `ap_id` is not indexed - there's an example here https://github.com/TryGhost/ActivityPub/blob/de2e070f31c860a027c5d9f22dfaf68dbe70fceb/src/post/post.repository.knex.ts#L154-L156

src/http/api/views/account.follows.view.ts
```diff
@@ -141,10 +183,15 @@ export class AccountFollowsView {
...
+            .first();
+
         let page: CollectionPage | null = null;
 
         try {
-            if (next !== '') {
+            //next would be a number only in case of non-paginated follow lists
```
allouis, line 193: I'm a little bit confused by this, but I think instead of passing around a `string` which is actually a `URL` or a `number` - which should just deal with those types directly.

So we could have `next: URL | number | null`

src/http/api/views/account.follows.view.ts
```diff
@@ -179,11 +226,90 @@ export class AccountFollowsView {
...
+                type === 'following'
+                    ? await actor.getFollowing()
+                    : await actor.getFollowers();
 
-        const accounts: AccountInfo[] = [];
+            if (!follows) {
+                return error('error-getting-follows');
+            }
+
+            console.log('Follows: ', follows);
```
allouis, line 240: I think this is leftover from dev

src/http/api/views/account.follows.view.ts
```diff
@@ -179,11 +226,90 @@ export class AccountFollowsView {
...
+                    const followsActorObj = await lookupObject(item.href, {
+                        documentLoader,
+                    });
+
+                    if (!isActor(followsActorObj)) {
+                        continue;
+                    }
+
+                    const followeeAccount = await this.db('accounts')
+                        .where('ap_id', followsActorObj.id?.toString() || '')
```
allouis, line 262: Same here re: `ap_id_hash`

src/http/api/views/account.follows.view.ts
```diff
@@ -179,11 +226,90 @@ export class AccountFollowsView {
...
+                try {
+                    const followsActorObj = await lookupObject(item.href, {
+                        documentLoader,
+                    });
+
+                    if (!isActor(followsActorObj)) {
+                        continue;
+                    }
+
+                    const followeeAccount = await this.db('accounts')
```
allouis, line 261: If we're going to check our local db - let's do that before calling `lookupObject` and just use our local data? That way we save a network lookup!

src/http/api/views/account.follows.view.ts
```diff
@@ -179,11 +226,90 @@ export class AccountFollowsView {
...
+            const pageNumber = next ? Number.parseInt(next, 10) : 1;
+            const startIndex = (pageNumber - 1) * pageSize;
+
+            const pageUrls = follows.itemIds.slice(
+                startIndex,
+                startIndex + pageSize,
+            );
+
+            for await (const item of pageUrls) {
+                try {
```
allouis, line 252: This logic is (almost?) exactly the same as the loop below - are we able to deduplicate this - the method is getting pretty long!

allouis (changes requested):
We need to update a few of the db queries to make sure they're using indexes for perf 🐢 

We are doing a remote lookup of a follow(ers/ing) list - and then a remote lookup for each actor in it, but we also check if we have the actor locally - I think we should be first checking if we have the actor locally, and then we can skip the subsequent remote lookup, which will improve perf a lot in the cases we have them.

I would love if we can start using specific types, like `URL | number` instead of `string` and then trying to parse it here, I think that would help us split up the code into more manageable chunks too, because we can then have methods which accept _only_ a `URL` or `number`.

src/http/api/views/account.follows.view.ts
vershwal: Done in https://github.com/TryGhost/ActivityPub/pull/557/commits/92b298171accb7c5f0b0ab5cc240b5ddff22d18c

src/http/api/views/account.follows.view.ts
vershwal: Done in https://github.com/TryGhost/ActivityPub/pull/557/commits/92b298171accb7c5f0b0ab5cc240b5ddff22d18c

vershwal:
Latest Changes:-

- Used optimised query for fetching accounts using ap_id_hash and cleaned up dev logs : https://github.com/TryGhost/ActivityPub/pull/557/commits/92b298171accb7c5f0b0ab5cc240b5ddff22d18c
- Doing a remote lookup only when the account is not present locally : https://github.com/TryGhost/ActivityPub/pull/557/commits/2199da093a37b77d93f00ede1087becf8e06a5cb
- Refactoring: extract follow list processing into reusable method : https://github.com/TryGhost/ActivityPub/pull/557/commits/b0204ee194bee737dffdc75fa473d1cd9b730f3a

src/http/api/views/account.follows.view.ts
vershwal: Done here https://github.com/TryGhost/ActivityPub/pull/557/commits/b0204ee194bee737dffdc75fa473d1cd9b730f3a

allouis (commented):
Cool, thanks for the changes! I've left a few more comments

I think the big one is the flow control of different cases (paginated/unpaginated/don't know)

We've gone from 1 case to 3, because before it was just paginated, but by handling non-paginated, we also have to handle the fact we don't know, and following the same patterns we were before kinda breaks down!

I think we should try and be really explicit with each flow - I've outlined one way of doing it (I think - I haven't got full context), but the idea is that we are very clear about what path is what, and we do that using code, not comments!

src/http/api/views/account.follows.view.ts
```diff
@@ -245,6 +252,74 @@ export class AccountFollowsView {
...
+            try {
+                const followeeAccount = await this.db('accounts')
+                    .whereRaw('accounts.ap_id_hash = UNHEX(SHA2(?, 256))', [
+                        item.href || '',
+                    ])
+                    .first();
+
+                if (followeeAccount) {
+                    accounts.push({
+                        id: String(followeeAccount.id),
```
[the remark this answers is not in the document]
vershwal: Fixed in https://github.com/TryGhost/ActivityPub/pull/557/commits/f0f74b2fc8c6c8df85da61b06a4bea767e1bbe65

src/http/api/views/account.follows.view.ts
```diff
@@ -245,6 +252,74 @@ export class AccountFollowsView {
...
+    ) {
+        const ctx = this.fedifyContextFactory.getFedifyContext();
+        const documentLoader = await ctx.getDocumentLoader({
+            handle: 'index',
+        });
+        for await (const item of followsList) {
+            try {
+                const followeeAccount = await this.db('accounts')
+                    .whereRaw('accounts.ap_id_hash = UNHEX(SHA2(?, 256))', [
+                        item.href || '',
```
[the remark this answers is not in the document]
vershwal: Fixed in https://github.com/TryGhost/ActivityPub/pull/557/commits/f0f74b2fc8c6c8df85da61b06a4bea767e1bbe65

src/http/api/views/account.follows.view.ts
```diff
@@ -245,6 +252,74 @@ export class AccountFollowsView {
         });
     }
 
+    private async processFollowsList(
+        followsList: URL[],
+        accounts: AccountInfo[],
```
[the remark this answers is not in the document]
vershwal: Fixed in https://github.com/TryGhost/ActivityPub/pull/557/commits/f0f74b2fc8c6c8df85da61b06a4bea767e1bbe65

src/http/api/views/account.follows.view.ts
```diff
@@ -151,90 +194,54 @@ export class AccountFollowsView {
...
-
-        const accounts: AccountInfo[] = [];
-
-        for await (const item of page.itemIds) {
-            try {
-                const followsActorObj = await lookupObject(item.href, {
-                    documentLoader,
-                });
+        // Handling non-paginated follow lists
+        if (!page || Number(next)) {
```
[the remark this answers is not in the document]
vershwal: Cleaned up `getFollowsByRemoteLookUp` in https://github.com/TryGhost/ActivityPub/pull/557/commits/706e9c3ab7c51e2868be55a154a90a7c4cae47a4

src/http/api/views/account.follows.view.ts
```diff
@@ -90,6 +137,80 @@ export class AccountFollowsView {
...
+        next: URL,
+        documentLoader: DocumentLoader,
+    ): Promise<Result<CollectionPage, GetFollowsError>> {
+        let page: CollectionPage | null = null;
+
+        try {
+            // Ensure the next parameter is for the same host as the actor. We
+            // do this to prevent blindly passing URIs to lookupObject (i.e next
+            // param has been tampered with)
+            const { host: actorHost } = actor?.id || new URL('');
```
allouis, line 151: `new URL('')` will always throw FYI - we have it in a `try/catch` block though so it's okay for now, but just a heads up

allouis (approved):

PR #558, opened 2025-04-23
AP-982 | Attach images to notes/replies

vershwal:
Ref https://linear.app/ghost/issue/AP-982 https://linear.app/ghost/issue/AP-1084

- Store the image url on the PostEntity in `imageUrl`. This is easier to work with for a single image
- Set the imageUrl on the attachment properties
-- Tested this will many options and setting it to the attachments using Image obj works. By works I mean image correctly shows up on mastodon. 
- Updated /note API to accept an image URL
-- Verifing the image url
- Updated /reply API to accept image
-- Verifing the image url

### DIFF
--- a/features/create-note.feature
+++ b/features/create-note.feature
@@ -46,3 +46,19 @@ Feature: Creating a note
       Hello, world!
       """
     Then Activity "Note" is sent to all followers
+
+  Scenario: Creating a note with an image URL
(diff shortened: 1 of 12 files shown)


allouis (commented):
Few comments on the types

allouis (approved):

PR #561, opened 2025-04-24
Added generated columns and indexes to keyvalue to speed up query per…

rmgpinto:
…formance

ref https://linear.app/ghost/issue/AP-1095

- Improve query performance by adding indexes and changing queries to use those indexes.


migrate/migrations/000042_add-generated-columns-and-indexes-to-keyvalue-table.up.sql
```diff
@@ -0,0 +1,11 @@
+ALTER TABLE key_value
+    ADD COLUMN object VARCHAR(255)
+    GENERATED ALWAYS AS (LEFT(JSON_UNQUOTE(value->>"$.object"), 255)) STORED;
```
allouis, line 3: `$.object` is not always a string - will the `LEFT` function handle that?

rmgpinto: It will output the actual object, ex: `{"id": "...", "type": "Follow", "actor": {"id": "https://..." } ..`
The `WHERE` clause will fail on that `=` comparison.

PR #562, opened 2025-04-24
Added handling for saving an account with `null` values for images

mike182uk:
ref https://linear.app/ghost/issue/AP-1097

When trying to save an account with `null` values for `avatarUrl` or `bannerImageUrl` the underlying db record would not be correctly updated due to the logic for saving the account expecting the provided values to be a URL. `null` is a valid value for these fields so we should handle it accordingly

### DIFF
--- a/src/account/account.repository.knex.integration.test.ts
+++ b/src/account/account.repository.knex.integration.test.ts
@@ -1,4 +1,4 @@
-import { beforeAll, describe, expect, it, vi } from 'vitest';
+import { beforeAll, beforeEach, describe, expect, it, vi } from 'vitest';
 
 import assert from 'node:assert';
 import { AsyncEvents } from 'core/events';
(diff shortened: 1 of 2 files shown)


src/account/account.repository.knex.integration.test.ts
```diff
@@ -1,4 +1,4 @@
-import { beforeAll, describe, expect, it, vi } from 'vitest';
```
mike182uk, line 1: Sorry for the messy diff in this test

The test i've added for the functionality change is the last one in the diff

The rest of the changes is just some refactoring to remove code duplication and ensure we have fresh data for each test as we were unintentionally relying on data from previous test runs 😬

src/account/account.repository.knex.ts
```diff
@@ -24,8 +24,12 @@ export class KnexAccountRepository {
                 name: account.name,
                 bio: account.bio,
                 username: account.username,
-                avatar_url: account.avatarUrl?.href,
-                banner_image_url: account.bannerImageUrl?.href,
+                avatar_url:
+                    account.avatarUrl == null ? null : account.avatarUrl?.href,
```
sagzy, line 28: Could also be written as `account.avatarUrl?.href ?? null`

mike182uk: Nice one, updated in https://github.com/TryGhost/ActivityPub/pull/562/commits/3635a0125d946489ab815b0ee7a66632d365ce67

src/account/account.repository.knex.integration.test.ts
allouis: Lol this diff is insane

allouis (approved):

PR #565, opened 2025-04-24
Added error handling in GCPStorageService

vershwal:
ref https://linear.app/ghost/issue/AP-982/

### DIFF
--- a/src/storage/gcloud-storage/gcp-storage.service.ts
+++ b/src/storage/gcloud-storage/gcp-storage.service.ts
@@ -1,4 +1,5 @@
 import { type Bucket, Storage } from '@google-cloud/storage';
+import type { Logger } from '@logtape/logtape';
 import { type Result, error, isError, ok } from 'core/result';
 import { v4 as uuidv4 } from 'uuid';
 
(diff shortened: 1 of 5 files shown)


src/storage/gcloud-storage/gcp-storage.service.ts
```diff
@@ -110,35 +118,47 @@ export class GCPStorageService {
...
-                return url.startsWith(emulatorBaseUrl);
+                if (!url.startsWith(emulatorBaseUrl)) {
+                    return error('invalid-url');
+                }
+                return ok(true);
             }
 
             // Verify if the URL matches the standard Google Cloud Storage public URL pattern for our bucket
             const gcsUrlPattern = new RegExp(
                 `https://storage.googleapis.com/${this.bucketName}/`,
```
allouis, line 136: Why do we do a `startsWith` for the emulatorHost flow - but a regex test here?

src/storage/gcloud-storage/gcp-storage.service.ts
```diff
@@ -110,35 +118,47 @@ export class GCPStorageService {
         return ok(true);
     }
 
-    async verifyImageUrl(url: string): Promise<boolean> {
+    async verifyImageUrl(
+        url: string,
```
allouis, line 122: Would be great if this was a `URL` - then we can do things like `url.host === expectedHost`

src/storage/gcloud-storage/gcp-storage.service.ts
vershwal: Not url is of type `URL` so I am directly checking host now

PR #566, opened 2025-04-24
Refactored Account to improve types & reduce boilerplate

allouis:

The Account entity had grown complex due to its implementation with private
members and public getters for privately mutable state. We were also
encountering issues with nullable `id` fields that required awkward
PersistedAccount type workarounds.

- Account now always has a required `id` field
- Introduced AccountDraft type for creating new accounts without IDs
- Made AccountEntity immutable, returning new instances on save operations
- Aligned `apFollowers` field to be nullable, matching database schema

This refactoring simplifies type checking in the block method implementation,
eliminating the need for ID checking and type casting when ensuring we're
working with persisted accounts. No functional changes were introduced, though
there is a slight type change with `apFollowers` now being nullable to match
the database.


src/account/account.entity.ts
```diff
@@ -1,161 +1,91 @@
...
+    readonly bannerImageUrl: URL | null;
+    readonly apId: URL;
+    readonly apFollowers: URL | null;
+    readonly isInternal: boolean;
+    /**
+     * Returns a new Account instance which needs to be saved.
+     */
+    updateProfile(params: ProfileUpdateParams): Account;
+    /**
+     * @deprecated
```
allouis, line 21: This code is weird and shouldn't really live here, or at least not be exposed like this - why does an Account have a `getApIdForPost` method lol

src/account/account.entity.ts
```diff
@@ -1,161 +1,91 @@
...
+    readonly id: number;
+    readonly uuid: string;
+    readonly username: string;
+    readonly name: string | null;
+    readonly bio: string | null;
+    readonly url: URL;
+    readonly avatarUrl: URL | null;
+    readonly bannerImageUrl: URL | null;
+    readonly apId: URL;
+    readonly apFollowers: URL | null;
```
allouis, line 14: This is nullable now

src/account/account.entity.ts
```diff
@@ -174,28 +104,65 @@ export class Account extends BaseEntity {
...
-            data.avatarUrl,
-            data.bannerImageUrl,
-            data.site,
-            data.apId,
-            data.url,
-            data.apFollowers,
-        );
+    updateProfile(params: ProfileUpdateParams) {
+        type P = ProfileUpdateParams;
+        const get = <K extends keyof P>(prop: K): P[K] =>
```
allouis, line 112: I just wanted to avoid the repetition of
```ts
username: params.username === undefined ? this.username : params.username,
```

But I can put it back if it's preferred

src/account/account.entity.ts
```diff
@@ -174,28 +104,65 @@ export class Account extends BaseEntity {
...
+    avatarUrl: URL | null;
+    bannerImageUrl: URL | null;
+    apId: URL;
+    apFollowers: URL | null;
+};
+
+type AccountDraftData = InternalAccountDraftData | ExternalAccountDraftData;
+
+type Data<T> = {
+    // biome-ignore lint/suspicious/noExplicitAny: These are not publicly usable instances of any
```
allouis, line 166: This is legit - the `any` here is just to match functions, which we strip out - it doesn't leak into our code

src/account/account.entity.ts
```diff
@@ -174,28 +104,65 @@ export class Account extends BaseEntity {
...
+    url: URL | null;
+    avatarUrl: URL | null;
+    bannerImageUrl: URL | null;
+    apId: URL;
+    apFollowers: URL | null;
+};
+
+type AccountDraftData = InternalAccountDraftData | ExternalAccountDraftData;
+
+type Data<T> = {
```
allouis, line 165: Yeah I can probably put this somewhere else, but where?

src/account/account.repository.knex.integration.test.ts
```diff
@@ -197,28 +197,17 @@ describe('KnexAccountRepository', () => {
...
-            account.ap_id ? new URL(account.ap_id) : null,
-            account.url ? new URL(account.url) : null,
-            account.ap_followers_url ? new URL(account.ap_followers_url) : null,
-        );
+        const accountEntity = await accountRepository.getBySite({
+            id: 1,
+        } as Site);
 
-        accountEntity.updateProfile({
+        const updated = accountEntity.updateProfile({
```
allouis, line 204: Immutable - so this had to change

src/account/account.repository.knex.ts
```diff
@@ -13,12 +13,6 @@ export class KnexAccountRepository {
     ) {}
 
     async save(account: Account): Promise<void> {
-        if (account.isNew) {
```
allouis, line 16: Accounts are never new!

src/http/api/views/account.posts.view.ts
```diff
@@ -82,8 +82,8 @@ export class AccountPostsView {
 
     async getPostsByApId(
         apId: URL,
-        account: PersistedAccount,
-        currentContextAccount: PersistedAccount,
+        account: Account | null,
```
allouis, line 85: No more PersistedAccount! But this _is_ nullable, I needed to fix the type. We do `account?.isInternal` below - so this is safe

src/post/post.entity.ts
```diff
@@ -201,17 +201,11 @@ export class Post extends BaseEntity {
     }
 
     addRepost(account: Account) {
-        if (!account.id) {
```
allouis, line 204: Yay

src/account/account.entity.ts
```diff
@@ -174,28 +104,65 @@ export class Account extends BaseEntity {
...
+
+        return AccountEntity.create({
+            ...this,
+            username: get('username'),
+            name: get('name'),
+            bio: get('bio'),
+            avatarUrl: get('avatarUrl'),
+            bannerImageUrl: get('bannerImageUrl'),
+        });
     }
```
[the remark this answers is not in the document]
allouis: I don't think this is true, we explicitly do not return `undefined` if it is that on the `params` object, and instead return the prop from `this` which *does* match the type required in `Data<Account>`

src/account/account.entity.ts
mike182uk: Maybe update the message to include this?

src/account/account.entity.ts
mike182uk: Nah, this is a cool utility, maybe we have generic `utils` module thing somewhere?

src/account/account.entity.ts
```diff
@@ -1,161 +1,91 @@
...
+    /**
+     * @deprecated
+     */
+    getApIdForPost(post: { type: CreatePostType; uuid: string }): URL;
+}
 
-export interface AccountData {
-    id: number;
-    uuid: string | null;
+export interface AccountDraft {
```
mike182uk, line 26: What about this just being `type AccountDraft = Omit<Account, 'id'>` so we don't have to list all of the account properties here?

src/account/account.entity.ts
allouis: I did try to pull it out - but unfortunately it does't work well when you pull it out, because you need to pass in the `this` value and it becomes a bit of a mess like `get(this, params, 'key')` PLUS the types become super fucky

src/account/account.entity.ts
mike182uk: Ah ok, yeh i'm fine with that, i guess it's unlikely they are going to come out of sync without us knowing

PR #572, opened 2025-04-24
Added handling for feed post insertion prevention 

mike182uk:

Added handling for feed post insertion prevention when an account is blocked:
- When a post is created by a blocked account, ensure it is not put into the feeds of any users that have blocked the author
- When a post is reposted by a blocked account, ensure it is not put into the feeds of any users that have blocked the reposter
- When a post is reposted by an account, ensure it is not put into the feeds of any users that have blocked the author
- When a post is created by a an account and then reposted by a blocked account, ensure it is not put into the feeds of any users except for the reposter

### DIFF
--- a/src/moderation/moderation.service.ts
+++ b/src/moderation/moderation.service.ts
@@ -32,7 +32,7 @@ export class ModerationService {
         userIds: number[],
         post: Post,
         repostedBy?: number,
-    ) {
+    ): Promise<number[]> {
         // Map user ids to their corresponding account ids
         const userAccountMap = new Map<number, number>();
 
@@ -47,12 +47,14 @@ export class ModerationService {
         // reposter
         if (repostedBy) {
             const authorHasBlockedReposter = await this.db('blocks')
+                .innerJoin('accounts', 'blocks.blocked_id', 'accounts.id')
                 .where('blocker_id', post.author.id)
                 .andWhere('blocked_id', repostedBy)
+                .select('blocked_id')
                 .first();
 
             if (authorHasBlockedReposter) {
-                return [userAccountMap.get(repostedBy)];
+                return [authorHasBlockedReposter.blocked_id];
             }
         }
 
--- a/src/moderation/moderation.service.ts
+++ b/src/moderation/moderation.service.ts
@@ -1,7 +1,7 @@
-import type { Knex } from 'knex';
-
 import type { Account } from 'account/account.entity';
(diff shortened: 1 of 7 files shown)


src/moderation/moderation.service.ts
```diff
@@ -27,4 +27,56 @@ export class ModerationService {
...
+        repostedBy?: number,
+    ) {
+        // Map user ids to their corresponding account ids
+        const userAccountMap = new Map<number, number>();
+
+        for (const row of await this.db('users')
+            .whereIn('id', userIds)
+            .select('id', 'account_id')) {
+            userAccountMap.set(row.id, row.account_id);
+        }
```
allouis, line 43:
```suggestion
        const rows = await this.db('users')
            .whereIn('id', userIds)
            .select('id', 'account_id'));

        for (const row of rows) {
           userAccountMap.set(row.id, row.account_id);
        }
```

allouis: Maybe a lil easier to parse?

src/moderation/moderation.service.ts
```diff
@@ -27,4 +27,56 @@ export class ModerationService {
...
+        // has blocked the author, and if so, filter out everybody but the
+        // reposter
+        if (repostedBy) {
+            const authorHasBlockedReposter = await this.db('blocks')
+                .where('blocker_id', post.author.id)
+                .andWhere('blocked_id', repostedBy)
+                .first();
+
+            if (authorHasBlockedReposter) {
+                return [userAccountMap.get(repostedBy)];
```
allouis, line 55: I think this can return `[undefined]` - an explicit return type in the function sig would have caught this!

allouis (commented):
One main comment about returning the wrong type - small suggestion on the for loop - otherwise :+1:

src/moderation/moderation.service.ts
mike182uk: Nice catch! Updated in https://github.com/TryGhost/ActivityPub/pull/572/commits/ef06d1aaac15e7fd9631478e7e96b3874073c6f5

src/moderation/moderation.service.ts
mike182uk: Yeh, don't know why I decided to do it this way 🙃

Fixed in https://github.com/TryGhost/ActivityPub/pull/572/commits/755d99fd4a5739566c595c6742f172bcc0333ddc

allouis (approved):

PR #580, opened 2025-04-28
Added handling for notification insertion prevention when an account is blocked

mike182uk:

Added handling for notification insertion prevention when an account is blocked:
- When a blocked account has attempted to follow the blocking account, ensure no notifications are not inserted for the blocking account
- When a blocked account has liked a post from the blocking account, ensure no notifications are not inserted for the blocking account
- When a blocked account has reposted a post from the blocking account, ensure no notifications are not inserted for the blocking account
- When a blocked account has replied to a post from the blocking account, ensure no notifications are not inserted for the blocking account

### DIFF
--- a/src/moderation/moderation.service.ts
+++ b/src/moderation/moderation.service.ts
@@ -31,7 +31,7 @@ export class ModerationService {
     async filterUsersForPost(
         userIds: number[],
         post: Post,
-        repostedBy?: number,
+        interactionAccountId?: number,
     ): Promise<number[]> {
         // Map user ids to their corresponding account ids
         const userAccountMap = new Map<number, number>();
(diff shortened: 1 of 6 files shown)


src/moderation/moderation.service.ts
```diff
@@ -31,7 +31,7 @@ export class ModerationService {
     async filterUsersForPost(
         userIds: number[],
         post: Post,
-        repostedBy?: number,
+        interactionAccountId?: number,
```
mike182uk, line 34: Changed this to `interactionAccountId` so the same logic can be used for `like` as well as `repost` (in both cases we want the same thing to happen - prevent feed insertion and prevent notification)

allouis (approved):

PR #581, opened 2025-04-28
Added image compression to reduce file size

sagzy:
ref https://linear.app/ghost/issue/PROD-1578

- added basic image compression using sharp, with the following output parameters:
    - max 2000px width and 2000px height (keeps the ratio, doesn't enlarge smaller images)
    - quality of 75% for jpeg/webp
    - maximum compression for png (no quality loss)
   
- also reduced the file size validation from 25 MB to 5 MB


package.json
```diff
@@ -82,6 +82,7 @@
     "mysql2": "3.14.1",
     "node-jose": "2.2.0",
     "sanitize-html": "2.16.0",
+    "sharp": "^0.34.1",
```
mike182uk, line 85: Lets pin this to a `0.34.1`

src/storage/gcloud-storage/gcp-storage.service.ts
```diff
@@ -119,6 +121,41 @@ export class GCPStorageService {
...
+            }
+
+            if (format === 'webp') {
+                return sharpPipeline.webp({ quality: 75 }).toBuffer();
+            }
+
+            return fileBuffer;
+        } catch (error) {
+            // Return buffer on the original file if compression failed
+            return fileBuffer;
```
mike182uk, line 155: Do we want to log an error here or something?

src/storage/gcloud-storage/gcp-storage.service.ts
```diff
@@ -119,6 +121,41 @@ export class GCPStorageService {
...
+        }
+        const fileBuffer = Buffer.concat(chunks);
+
+        try {
+            const sharpPipeline = sharp(fileBuffer).resize({
+                width: 1200,
+                withoutEnlargement: true,
+            });
+
+            const format = file.type.split('/')[1];
```
mike182uk, line 138: Should we `toLowerCase` this incase an upload is named like `foo.JPEG`?

src/storage/gcloud-storage/gcp-storage.service.ts
```diff
@@ -119,6 +121,41 @@ export class GCPStorageService {
         return ok(true);
     }
 
+    private async compressFile(file: File): Promise<Buffer> {
+        const chunks: Buffer[] = [];
+
+        for await (const chunk of file.stream()) {
+            chunks.push(Buffer.from(chunk));
```
mike182uk, line 128: Isn't `chunk` already a buffer here?

src/storage/gcloud-storage/gcp-storage.service.ts
mike182uk: or alternatively doesn't sharp provide some way of detecting the format of the image from the contents rather than the name?

src/storage/gcloud-storage/gcp-storage.service.unit.test.ts
```diff
@@ -225,4 +252,120 @@ describe('GCPStorageService', () => {
...
+                2000,
+                2000,
+            );
+            const originalSize = mockFile.size;
+
+            const compressedBuffer = await (
+                service as unknown as {
+                    compressFile: (file: File) => Promise<Buffer>;
+                }
+            ).compressFile(mockFile);
```
mike182uk, line 274: This feels off, but I get why you need to do it (testing privates)

What do you think to moving this image handling to another service, like `ImageService`? This would give us a nice separation of concerns (the storage service specifically deals with writing / reading to storage) and then would make sense for `compressFile` to be on the public interface, making it easier to test

features/step_definitions/stepdefs.js
```diff
@@ -700,12 +701,21 @@ When(
...
+        const buffer = await sharp({
+            create: {
+                width: 200,
+                height: 200,
+                channels: 3,
+                background: { r: 255, g: 0, b: 0 },
+            },
+        })
+            .toFormat('jpg')
+            .toBuffer();
```
mike182uk, line 715: Seeming though this is an e2e test, we could just read a real file from the filesystem instead of programatically creating it? (see `fixtures` in other e2e tests). Maybe not worth the hassle? (just means it we ever decide to not use `sharp` we don't have to rip it out here)

src/storage/gcloud-storage/gcp-storage.service.ts
sagzy: ^ this is using the file MIME type, so e.g. `image/jpeg` for a JPG (not the file extension, e.g. `my-image.jpg`)

There is a validation on the file MIME types in the `validateFile` method ([code](https://github.com/TryGhost/ActivityPub/blob/2465f741cd52340431768bdf748aafef4159277d/src/storage/gcloud-storage/gcp-storage.service.ts#L110)) that allows only the following:
```
const ALLOWED_IMAGE_TYPES = [
    'image/jpg',
    'image/jpeg',
    'image/png',
    'image/webp',
];
```

features/step_definitions/stepdefs.js
sagzy: Yeah good idea! Feels better to work with a real image in a e2e test anyways. 

Fixed in c31dd1ccd9f436532e45e6ee41d627297bc323c9

src/storage/gcloud-storage/gcp-storage.service.ts
sagzy: Good one, fixed in 96b89d10fb4e3b9362da3081a985d3c86903a102

package.json
sagzy: Good catch, added `yarn config set save-prefix ""` to my `~/.yarnrc` config to avoid it in the future

Fixed in 25f417ef23b64c929a63287c9c516e3d02554012

src/storage/gcloud-storage/gcp-storage.service.ts
sagzy: Yeah good one! Updated in 48acfa49443b3faa2df3813d96befa9b20390fb4

src/storage/gcloud-storage/gcp-storage.service.unit.test.ts
sagzy: Yeah I think that's a good idea! Currently the `gcp-storage-service` is mixing logic that is specific to images (e.g. allowed images types, or storage path with images/ or now the image compression) with logic that is generic to any file storage. 

I suggest: I merge this PR with image compression logic, and then start a follow-up refactoring PR to avoid making this one too big / impossible to review. The refactoring is a good one to look at during clean code sesh too

src/storage/gcloud-storage/gcp-storage.service.ts
mike182uk: Ah my bad, all good!

src/storage/gcloud-storage/gcp-storage.service.unit.test.ts
mike182uk: Yeh I didn't look too deeply into what else the storage service is doing, but this approach works for me 👍

src/storage/gcloud-storage/gcp-storage.service.ts
```diff
@@ -119,6 +121,45 @@ export class GCPStorageService {
...
+
+            if (format === 'webp') {
+                return sharpPipeline.webp({ quality: 75 }).toBuffer();
+            }
+
+            return fileBuffer;
+        } catch (error) {
+            this.logger.error(
+                'Image compression failed, keeping original file',
+            );
```
mike182uk, line 158: Should we include the original error / error message somehow to give us an indication of exactly what the error was?

src/storage/gcloud-storage/gcp-storage.service.ts
```diff
@@ -119,6 +121,45 @@ export class GCPStorageService {
         return ok(true);
     }
 
+    private async compressFile(file: File): Promise<Buffer> {
+        const chunks: Buffer[] = [];
+
+        for await (const chunk of file.stream()) {
+            chunks.push(chunk as Buffer);
```
mike182uk, line 128: Ah, are you having to cast this because typescript thinks its not a buffer? If thats the case, maybe we go with `Buffer.isBuffer(chunk) ? chunk : Buffer.from(chunk)` ?

mike182uk (approved):

src/storage/gcloud-storage/gcp-storage.service.ts
sagzy: Yep good one, fixed in 2d5bfa8df9c69c6c7e780ecf74aa659f422fe639

PR #588, opened 2025-04-29
Moved `follow` handler to dedicated class

mike182uk:
no ref

Moved `follow` handler to dedicated class to follow the same pattern as the `create` & `delete` handlers in preparation for further work re: rejecting follow requests for blocked accounts

### DIFF
--- a/src/activity-handlers/follow.handler.ts
+++ b/src/activity-handlers/follow.handler.ts
@@ -1,0 +1,78 @@
+import { Accept, type Context, type Follow } from '@fedify/fedify';
+import { v4 as uuidv4 } from 'uuid';
+
+import type { AccountService } from 'account/account.service';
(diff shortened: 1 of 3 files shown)


src/activity-handlers/follow.handler.ts
```diff
@@ -0,0 +1,78 @@
...
+
+import type { AccountService } from 'account/account.service';
+import { mapActorToExternalAccountData } from 'account/utils';
+import type { ContextData } from 'app';
+import { addToList } from 'kv-helpers';
+
+export class FollowHandler {
+    constructor(private readonly accountService: AccountService) {}
+
+    async handle(ctx: Context<ContextData>, follow: Follow) {
```
mike182uk, line 12: No logic changes here other than referencing the injected `accountService`

PR #593, opened 2025-04-30
Prevented replying to blocked users

allouis:

We do not want to allow users on our service to even attempt to reply to
accounts which have blocked them.


src/post/post.service.integration.test.ts
```diff
@@ -0,0 +1,293 @@
...
+        it('should create a note successfully', async () => {
+            const content = 'This is a test note';
+
+            const result = await postService.createNote(account, content);
+
+            expect(isError(result)).toBe(false);
+
+            if (isError(result)) {
+                throw new Error('Result should not be an error');
+            }
```
mike182uk, line 83: The above assertion should mean we don't need to do this right?

src/post/post.service.integration.test.ts
```diff
@@ -0,0 +1,293 @@
...
+
+            const post = getValue(result);
+            expect(post).toBeInstanceOf(Post);
+            expect(post.author.id).toBe(account.id);
+            expect(post.content).toContain(content);
+            expect(post.type).toBe(0); // PostType.Note
```
mike182uk, line 89: Maybe just use the type directly? 😅

src/post/post.service.integration.test.ts
```diff
@@ -0,0 +1,293 @@
...
+            expect(post.inReplyTo).toBeNull();
+
+            // Verify the post was saved to database
+            const savedPost = await postRepository.getById(post.id!);
+            expect(savedPost).not.toBeNull();
+            expect(savedPost?.id).toBe(post.id);
+            expect(savedPost?.content).toContain(content);
```
mike182uk, line 96: We could use `!` instead of the optional chaining as we know `savedPost` is a`Post`, otherwise the `toBeNull` check would fail

src/post/post.service.integration.test.ts
```diff
@@ -0,0 +1,293 @@
...
+            const imageUrl = new URL('https://example.com/bad-image.jpg');
+
+            const result = await serviceWithFailingStorage.createNote(
+                account,
+                content,
+                imageUrl,
+            );
+
+            expect(isError(result)).toBe(true);
+            expect(getError(result as Err<string>)).toBe('invalid-image-url');
```
mike182uk: How come we have to `as Err<string>` here?

src/post/post.service.integration.test.ts
```diff
@@ -0,0 +1,293 @@
...
+                bucketName: 'test-bucket',
+                uploadBase64Image: vi.fn(),
+                verifyImageUrl: vi
+                    .fn()
+                    .mockResolvedValue(
+                        createError(
+                            'invalid-image-url' as ImageVerificationError,
+                        ),
+                    ),
+                getPublicUrl: vi.fn(),
```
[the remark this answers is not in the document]
allouis: I fixed this

src/http/api/reply.ts
```diff
@@ -0,0 +1,255 @@
...
+    if (!inReplyToId) {
+        return new Response(
+            JSON.stringify({ message: 'ID should be a valid URL' }),
+            {
+                status: 400,
+            },
+        );
+    }
+
+    const objectToReplyTo = await lookupObject(apCtx, id);
```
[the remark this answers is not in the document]
allouis: This matches existing functionality in `handlers.ts`

src/post/post.service.integration.test.ts
allouis: `getError` doesn't accept a `Result` it only accepts an `Error` (aliased to `Err` here)

Usually they're wrapped so like:
```
if (isError(result)) {
  getError(result); // This works because `isError` tells typescript the type
}
```

But here the `isError(result)` call isn't wrapping this we we have to cast it explicitly

src/post/post.service.integration.test.ts
allouis: Yeah good point

src/post/post.service.integration.test.ts
allouis: 😬 Yup!

src/post/post.service.integration.test.ts
allouis: Fixed

src/post/post.service.integration.test.ts
allouis: Fixed

mike182uk (approved):

PR #598, opened 2025-04-30
Ensured follow relationship is removed when account is blocked

mike182uk:

When an account blocks another account, any existing follow relationships between both accounts, in any direction, should be removed. This only removes the relationship on our system and does not federate the change (i.e we do not send an `Undo` of the original `Follow` activity).

### DIFF
--- a/src/account/account.repository.knex.integration.test.ts
+++ b/src/account/account.repository.knex.integration.test.ts
@@ -1,62 +1,31 @@
 import { beforeAll, beforeEach, describe, expect, it, vi } from 'vitest';
 
 import assert from 'node:assert';
-import { AsyncEvents } from 'core/events';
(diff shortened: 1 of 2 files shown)


src/account/account.repository.knex.integration.test.ts
```diff
@@ -1,62 +1,31 @@
...
-            async getSiteSettings(host: string) {
-                return {
-                    site: {
-                        title: 'Test Site',
-                        description: 'A fake site used for testing',
-                        icon: 'https://testing.com/favicon.ico',
-                    },
-                };
-            },
-        });
```
mike182uk, line 59: `FixtureManager` handles all this for us 👍

allouis (approved):

PR #599, opened 2025-04-30
Moved `follow` / `unfollow` API actions to dedicated controller

mike182uk:
no ref

Moved `follow` / `unfollow` API actions to a dedicated controller to follow the same pattern as the `block` controller, in preparation for modification of the `handleFollow` to prevent processing of requests from blocked accounts


mike182uk:
There is no change of logic in the controller actions other than referencing the injected `accountService`

allouis (approved):

PR #601, opened 2025-04-30
Refactored cucumber setup to allow for multiple step definition files

mike182uk:
no refs

Miscellaneous refactorings to allow us to move to using multiple step definition files instead of one large file:

- Introduced `support` directory to organize support files
    - Moved generic helper functions out of `stepdefs.js` and into their respective support files
- Removed nearly all global state from `stepdefs.js`
    - Support files keep track of there global state (i.e `db.js` keeping track of database client instance)
- Added `block_steps.js` to organize block related step definitions
- Moved global static constants to environment variables

I opted to just use the built in `World` rather than a custom one - We could maybe move to this at a later point and use it to keep track of global state etc

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -174,6 +174,8 @@ services:
       - GCP_BUCKET_NAME=activitypub
       - GCP_STORAGE_EMULATOR_HOST=http://fake-gcs:4443
       - TAGS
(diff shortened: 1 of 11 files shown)


allouis (approved):
Such an improvement!

PR #604, opened 2025-05-01
Improved error responses from HTTP API

allouis:

As part of the block work we're handling specific errors from the API to show a message baout not being able to interact with the account. In order to handle errors easily in the frontend we want to standardise all of the errors returned

### DIFF
--- a/src/http/api/reply.ts
+++ b/src/http/api/reply.ts
@@ -48,20 +48,6 @@ export async function handleCreateReply(
         logger,
     });
 
-    if (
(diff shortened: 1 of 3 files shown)


mike182uk (approved):

src/http/api/reply.ts
```diff
@@ -45,11 +48,25 @@ export async function handleCreateReply(
         logger,
     });
 
+    if (
```
allouis, line 51: I'll remove this

PR #615, opened 2025-05-05
Preparing the Repost controller to have moderation restrictions

allouis:
We want to move all of the repost logic into the service layer so we have a single place to add restrictions to blocked accounts


features/repost-activity.feature
```diff
@@ -12,9 +12,7 @@ Feature: Reposting a post/note
     Then the request is accepted
     And the object "Note" should be reposted
     And the object "Note" should have a repost count of 1
-    And a "Announce(Note)" activity is sent to "Alice"
-    And a "Announce(Note)" activity is in the Outbox
-    And the found "Announce(Note)" has property "object.attributedTo" of type "object"
+    And an Announce(Note) is sent to "Alice"
```
allouis, line 15: Trying to use more specific steps here

features/step_definitions/stepdefs.js
```diff
@@ -476,38 +476,40 @@ When('we repost the object {string}', async function (name) {
 
 Then('the object {string} should be reposted', async function (name) {
     const response = await fetchActivityPub(
-        'http://fake-ghost-activitypub.test/.ghost/activitypub/inbox/index',
+        'http://fake-ghost-activitypub.test/.ghost/activitypub/feed',
```
allouis, line 479: Using feed instead of inbox

features/step_definitions/stepdefs.js
```diff
@@ -794,7 +796,7 @@ async function waitForOutboxActivity(
     const found = await findInOutbox(activity);
 
     if (found) {
-        return;
+        return found;
```
allouis, line 799: This is required for `const foundInOutbox = await waitForOutboxActivity(foundActivity);` to work

src/http/api/repost.ts
```diff
@@ -1,105 +1,83 @@
...
+                    return Conflict('Already reposted');
+                case 'upstream-error':
+                    return NotFound('Upstream error fetching post');
+                case 'not-a-post':
+                    return BadRequest('Not a post');
+                case 'missing-author':
+                    return NotFound('Post does not have an author');
+                default:
+                    exhaustiveCheck(error);
+            }
```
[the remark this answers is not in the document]
allouis: `repostByApId` doesn't return this error yet lol

src/post/post.service.ts
```diff
@@ -248,4 +250,30 @@ export class PostService {
...
+        if (reposted) {
+            return error('already-reposted');
+        }
+
+        post.addRepost(account);
+
+        await this.postRepository.save(post);
+
+        return ok(post);
+    }
```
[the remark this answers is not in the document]
allouis: This will come in a separate commit

src/post/post.service.ts
```diff
@@ -29,6 +29,8 @@ export type GetPostsError =
     | 'no-page-found'
     | 'not-an-actor';
 
+export type RepostError = GetByApIdError | 'already-reposted';
+
```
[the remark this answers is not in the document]
allouis: coming in separate commit

src/post/post.service.integration.test.ts
```diff
@@ -300,4 +300,127 @@ describe('PostService', () => {
...
+            expect(isError(firstRepost)).toBe(false);
+
+            // Try to repost it again
+            const result = await postService.repostByApId(
+                reposter,
+                postToRepost.apId,
+            );
+                reposter,
+                postToRepost.apId,
+            );
```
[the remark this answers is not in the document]
allouis: Yeah that was caused by your commit suggestions! I've fixed it now

features/step_definitions/stepdefs.js
```diff
@@ -1184,6 +1186,65 @@ Then('{string} is not in our Followers', async function (actorName) {
...
+    const foundInInbox = await waitForRequest(
+        'POST',
+        inboxUrl.pathname,
+        (call) => {
+            const body = JSON.parse(call.request.body);
+            return body.type === 'Undo' && body.object.type === 'Announce';
+        },
+    );
+
+    const foundActivity = JSON.parse(foundInInbox.request.body);
```
sagzy, line 1206: Out of scope here I guess: but curious how would this test look like, if we wanted to stop relying on the inbox? Do we have currently another way to assert that an activity has been dispatched to the recipient(s) outbox(es)?

sagzy (approved):
LGTM 👍 Added one curiosity, that is probably out of scope for this PR

features/step_definitions/stepdefs.js
allouis: This is checking that we're sending it to their inbox, so it doesn't depend on the KV inbox stuff - we're checking for a network request to an "external" inbox - does that make sense?

sagzy: Ah missed that! Great then :)

PR #616, opened 2025-05-05
Prevented users from reposting content of accounts which have blocked them

allouis:

As part of the block functionality we want to disable interaction entirely for blocked users on our own server. The frontend will display an error when we respond with a 403 indicated that the account has restricted who can interact with them.


sagzy (approved):
LGTM 👍

PR #634, opened 2025-05-07
Add migrations for backfilling `domain` field on `accounts` table

mike182uk:
ref https://linear.app/ghost/issue/PROD-1637

Added 2 sets of migrations that:

1. Backfill the `domain` field extracting the value from an account's `ap_id`
2. Makes `domain` non nullable as we expect this field to always be present


allouis (approved):

allouis (changes requested):
You need to bump the migration numbers - 46 went in just recently

PR #641, opened 2025-05-08
Restricted interations from accounts blocked by domain

allouis:
ref https://linear.app/ghost/issue/PROD-697

Now that we have domain blocks we need to make sure that we check for the blocks when moderating interations.

### DIFF
--- a/src/moderation/moderation.service.integration.test.ts
+++ b/src/moderation/moderation.service.integration.test.ts
@@ -177,5 +177,36 @@ describe('ModerationService', () => {
 
             expect(charlieCanInteractWithAlice).toBe(true);
         });
+
+        it('should prevent interaction when the domain is blocked', async () => {
+            const [[aliceAccount], [bobAccount, bobSite]] = await Promise.all([
+                fixtureManager.createInternalAccount(),
+                fixtureManager.createInternalAccount(null, 'blocked.com'),
+            ]);
+
+            const [charlieAccount] =
+                await fixtureManager.createInternalAccount(bobSite);
+
+            await fixtureManager.createDomainBlock(
+                aliceAccount,
+                bobAccount.apId,
+            );
+
+            const bobCanInteractWithAlice =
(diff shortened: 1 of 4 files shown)


src/moderation/moderation.service.ts
```diff
@@ -64,11 +64,29 @@ export class ModerationService {
...
+            .where('id', interactionAccountId)
+            .first();
+
+        if (!accountRow) {
+            throw new Error(`Account missing ${interactionAccountId}`);
+        }
+
+        const domainBlock = await this.db('domain_blocks')
+            .where('blocker_id', targetAccountId)
+            .andWhere('domain', accountRow.domain)
```
allouis, line 87: Is it better to do a join here, on the `accounts.domain`?

src/moderation/moderation.service.integration.test.ts
```diff
@@ -177,5 +177,38 @@ describe('ModerationService', () => {
...
+
+            const [charlieAccount] =
+                await fixtureManager.createInternalAccount(bobSite);
+
+            const bobDomain = bobAccount.apId.hostname;
+
+            await client('domain_blocks').insert({
+                blocker_id: aliceAccount.id,
+                domain: bobDomain,
+            });
```
mike182uk, line 195: Could use the method you added to the fixture manager?

src/moderation/moderation.service.ts
mike182uk: I guess it would save a query right? something like:

```
select * from domain_blocks db
join accounts a on a.domain = db.domain
where db.domain = <domain>
and db.account_id = <blocked_account_id>
limit 1
```

but its probably not as explicit

mike182uk (approved):

src/moderation/moderation.service.integration.test.ts
allouis: Ah yes, that wasn't merged when I wrote this, but I'll do that

PR #642, opened 2025-05-08
Removed notifications from blocked domain on domain block

mike182uk:

When a domain is blocked by an account, any notifications that they have from accounts that are from the blocked domain are removed


src/notification/notification.service.integration.test.ts
```diff
@@ -843,13 +843,117 @@ describe('NotificationService', () => {
         });
 
         it('should do nothing if user is not found for blocker account', async () => {
```
mike182uk, line 845: I've updated this test as it wasn't doing anything useful

allouis (approved):

PR #645, opened 2025-05-08
Protected feeds and notifications from blocked domains

allouis:
ref https://linear.app/ghost/issue/PROD-697

### DIFF
--- a/src/account/account.repository.knex.integration.test.ts
+++ b/src/account/account.repository.knex.integration.test.ts
@@ -365,4 +365,55 @@ describe('KnexAccountRepository', () => {
         const follows = await client('follows').select('*');
         expect(follows).toStrictEqual([]);
     });
+
+    it('handles removing follow relationships when a domain is blocked', async () => {
+        // Create Alice from a different domain
+        const [aliceAccount] = await fixtureManager.createInternalAccount(
+            null,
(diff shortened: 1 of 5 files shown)


src/account/account.repository.knex.integration.test.ts
```diff
@@ -365,4 +365,50 @@ describe('KnexAccountRepository', () => {
...
+        // Create Charlie using the same site as Bob
+        const [charlieAccount] =
+            await fixtureManager.createInternalAccount(bobSite);
+
+        // Create follow relationships:
+        // 1. Alice follows Bob (will be removed when domain is blocked)
+        // 2. Charlie follows Alice (will be removed when domain is blocked)
+        await fixtureManager.createFollow(aliceAccount, bobAccount);
+        await fixtureManager.createFollow(charlieAccount, aliceAccount);
+
```
mike182uk, line 392: Could be good to create follow between charlie and bob and verify thats still there afterwards instead of checking the db is empty?

mike182uk (approved):

src/account/account.repository.knex.integration.test.ts
allouis: Yeah fair point

PR #646, opened 2025-05-08
Storing notes/reply images in post attachments

vershwal:
Ref https://linear.app/ghost/issue/PROD-1569

- For local actors we had image store in `featuredImage` whereas for remote actors it is stored in `attachments` 
- This PR includes changes to store the image in `attachments` when a note/reply is created. 
- `featuredImage` is kept as null - Not required

### DIFF
--- a/features/step_definitions/note_steps.js
+++ b/features/step_definitions/note_steps.js
@@ -89,7 +89,7 @@ Then(
     'note {string} has the image URL {string}',
     function (noteName, expectedImageUrl) {
         const object = this.objects[noteName];
-        assert.equal(object.attachment.url, expectedImageUrl);
-        assert.equal(object.attachment.type, 'Image');
+        assert.equal(object.attachment[0].url, expectedImageUrl);
+        assert.equal(object.attachment[0].type, 'Image');
     },
 );
--- a/features/step_definitions/note_steps.js
+++ b/features/step_definitions/note_steps.js
@@ -89,7 +89,9 @@ Then(
     'note {string} has the image URL {string}',
     function (noteName, expectedImageUrl) {
         const object = this.objects[noteName];
-        assert.equal(object.attachment[0].url, expectedImageUrl);
-        assert.equal(object.attachment[0].type, 'Image');
+        assert.ok(object.attachment, 'Note does not have attachments');
+
+        assert.equal(object.attachment.url, expectedImageUrl);
+        assert.equal(object.attachment.type, 'Image');
     },
 );
--- a/src/activitypub/fediverse-bridge.ts
+++ b/src/activitypub/fediverse-bridge.ts
@@ -65,12 +65,15 @@ export class FediverseBridge {
                 content: post.content,
                 summary: null,
                 published: Temporal.Now.instant(),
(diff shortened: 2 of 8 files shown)


features/step_definitions/note_steps.js
```diff
@@ -89,7 +89,7 @@ Then(
     'note {string} has the image URL {string}',
     function (noteName, expectedImageUrl) {
         const object = this.objects[noteName];
-        assert.equal(object.attachment.url, expectedImageUrl);
-        assert.equal(object.attachment.type, 'Image');
+        assert.equal(object.attachment[0].url, expectedImageUrl);
```
mike182uk, line 92: Maybe first assert `object.attachment` exists, and has 1 entry before accessing directly? I know we get the same result, but the error message will be a little less cryptic

src/activitypub/fediverse-bridge.ts
```diff
@@ -65,12 +65,15 @@ export class FediverseBridge {
                 content: post.content,
                 summary: null,
                 published: Temporal.Now.instant(),
-                attachments: post.imageUrl
-                    ? [
-                          new Image({
-                              url: post.imageUrl,
-                          }),
-                      ]
+                attachments: post.attachments
```
mike182uk, line 68: I know you've not changed this, but Is `attachments` the correct key for the Fedify object? I noticed in the cucumber test you checked `attachment` (singular)

src/activitypub/fediverse-bridge.ts
```diff
@@ -65,12 +65,15 @@ export class FediverseBridge {
...
+                    ? post.attachments
+                          .filter((attachment) => attachment.type === 'Image')
```
mike182uk, line 70: 🤔 What other attachments would there be? and why wouldn't we want to include them on the note?

src/post/post.entity.ts
```diff
@@ -336,6 +336,17 @@ export class Post extends BaseEntity {
             addPaidContentMessage: false,
         });
 
+        const postAttachment = imageUrl
+            ? [
+                  {
+                      type: 'Image',
+                      mediaType: null,
+                      name: null,
```
mike182uk, line 344: Are these ok just being `null`?

features/step_definitions/note_steps.js
vershwal: Added assert on `object.attachment`. https://github.com/TryGhost/ActivityPub/pull/646/commits/82ee3c72e4651249a908eed0e74744edf437607a

src/activitypub/fediverse-bridge.ts
vershwal: Yes `attachments` is the correct key. Tested that. But if there is a single attachment, fedify internally changes it to `attachment` (singular). Example [here](https://browser.pub/https%3A%2F%2Fprincivershwal.ghost.io%2F.ghost%2Factivitypub%2Fnote%2F5df81770-176f-4357-9784-657685ce232e). This is why I am using attachment (singular) in tests.

mike182uk (approved):

PR #647, opened 2025-05-08
Fixed error on like object network lookup

rmgpinto:
ref https://linear.app/ghost/issue/PROD-1587

- Added a try/catch to like object network lookup. Previously, it was throwing an error if the remote server was not be reachable.


sagzy (approved):
LGTM

PR #648, opened 2025-05-08
Update account system to `follow` / `unfollow` via entity

mike182uk:
ref https://linear.app/ghost/issue/PROD-679

This includes a bunch of modifications to the account system in preparation for usage of the account unfollowed event to remove posts of unfollowed accounts from feeds:

- Emitted `account.unfollowed` event when an account is saved (if applicable)
- Added `follow` & `unfollow` methods to account entity
- Updated account repository to handle persisting of follows based on events
- Updated existing `account.followed` event to use IDs instead of full accounts for consistency with other events
- Added `followAccount` and `unfollowAccount` methods to the account service
- Deprecated usage of `recordAccountFollow` and updated call sites to use `followAccount` instead
  - Have marked as deprecated instead of removing completely as the method is still in-use in a whole bunch of tests - We can clean this up later
- Removed `recordAccountUnfollow` and updated call sites to use `unfollowAccount` instead

### DIFF
--- a/src/account/account-followed.event.ts
+++ b/src/account/account-followed.event.ts
@@ -1,17 +1,19 @@
-import type { Account } from 'account/types';
-
 export class AccountFollowedEvent {
     constructor(
-        private readonly account: Account,
-        private readonly follower: Account,
+        private readonly accountId: number,
+        private readonly followerId: number,
     ) {}
 
-    getAccount(): Account {
-        return this.account;
+    getAccountId(): number {
+        return this.accountId;
     }
 
-    getFollower(): Account {
-        return this.follower;
+    getFollowerId(): number {
+        return this.followerId;
+    }
+
+    getName(): string {
+        return 'account.followed';
     }
 
     static getName(): string {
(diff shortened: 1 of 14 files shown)


allouis:
I rebased this on `main` to fix some conflicts

src/http/api/follow.ts
```diff
@@ -107,11 +68,14 @@ export class FollowController {
...
-            actorToFollow,
+            {
+                id: followedAccount.apId,
+                inboxId: followedAccount.apInbox,
+            },
             follow,
         );
 
-        return new Response(JSON.stringify(await actorToFollow.toJsonLd()), {
+        return new Response(JSON.stringify({}), {
```
allouis, line 75: I think we should return the AccountDTO here

src/http/api/follow.ts
```diff
@@ -153,17 +139,15 @@ export class FollowController {
...
             unfollow,
         );
 
-        await this.accountService.recordAccountUnfollow(
-            accountToUnfollow,
-            account,
-        );
-
-        return new Response(JSON.stringify(unfollowJson), {
+        return new Response(JSON.stringify({}), {
```
allouis, line 150: AccountDTO here too?

src/http/api/follow.ts
allouis: I went against this - because 1. were not using the response and 2. it requires a bunch more queries

PR #661, opened 2025-05-13
Created notification when an user is mentioned

sagzy:
ref https://linear.app/ghost/issue/PROD-1737

- creates a notification of type `Mention` when one of our users is mentioned
- skips the notification if the post author is blocked


src/mention/mention-created.event.ts
```diff
@@ -0,0 +1,13 @@
+import type { Mention } from 'mention/mention.entity';
```
vershwal, line 1: I don't think we need the new `mention.entity.ts`. A good way is to follow the pattern that we have for [post-liked-event](https://github.com/TryGhost/ActivityPub/blob/main/src/post/post-liked.event.ts). This has a post and an accountId. We can have similar for mentions. 

`async createMentionNotification(post: Post, accountId: number)`

We also don't need the entity for saving mentions in db so this won't be reused any where else in the code.

src/notification/notification.service.ts
```diff
@@ -364,4 +365,39 @@ export class NotificationService {
             ])
             .delete();
     }
+
+    async createMentionNotification(mention: Mention) {
```
vershwal, line 369: Product Question: Should we create a notification when someone mentions their own account in a post?

If not, we should add a check for that. 

```
if (post.author.id === accountId) {
            // Do not create a notification if author mentions their own account
            return;
        }
```

src/notification/notification.service.ts
```diff
@@ -364,4 +365,39 @@ export class NotificationService {
...
+            .select('id')
+            .first();
+
+        if (!user) {
+            // If the mention is for an account that no longer exists or is external,
+            // don't create a notification
+            return;
+        }
+
+        const post = await this.db('posts')
```
vershwal, line 381: We can skip this db call if we already pass the post to `createMentionNotification`

vershwal (changes requested):

src/notification/notification.service.ts
sagzy: removed in 731d4d4d24dd984e00f9b9bead0a6c1ac8b463d8

src/notification/notification.service.ts
sagzy: Yeah good one! added in 731d4d4d24dd984e00f9b9bead0a6c1ac8b463d8

src/mention/mention-created.event.ts
sagzy: updated in 731d4d4d24dd984e00f9b9bead0a6c1ac8b463d8

vershwal (approved):

PR #666, opened 2025-05-13
Added mentions for notes and replies

vershwal:
Ref https://linear.app/ghost/issue/PROD-1736

- Parse mentions from the Note/Reply content
- If mentions are present, wrap it into hyperlinks and add to the db
- Send a account mentioned event when post is saved with mentions

### DIFF
--- a/src/lookup-helpers.ts
+++ b/src/lookup-helpers.ts
@@ -6,6 +6,9 @@ import {
     lookupWebFinger,
 } from '@fedify/fedify';
 import type { ContextData } from './app';
+import { type Result, error, ok } from './core/result';
+
+export type LookupError = 'no-links-found' | 'no-self-link' | 'lookup-error';
 
 export async function lookupActor(
     ctx: Context<ContextData>,
@@ -61,6 +64,9 @@ export async function lookupObject(
     return ctx.lookupObject(identifier, { documentLoader });
 }
 
+/**
+ * @deprecated use `lookupActorProfile`
+ */
 export async function lookupAPIdByHandle(
     ctx: Context<ContextData>,
     handle: string,
@@ -110,7 +116,7 @@ export async function lookupAPIdByHandle(
 export async function lookupActorProfile(
     ctx: Context<ContextData>,
     handle: string,
-): Promise<{ profileUrl: URL | null; apId: URL | null }> {
+): Promise<Result<URL, LookupError>> {
     try {
         // Remove leading @ if present
         const cleanHandle = handle.startsWith('@') ? handle.slice(1) : handle;
@@ -124,56 +130,30 @@ export async function lookupActorProfile(
         });
 
         if (!webfingerData?.links) {
-            ctx.data.logger.info(
-                `No links found in WebFinger response for handle ${handle}`,
-            );
-            return { profileUrl: null, apId: null };
+            ctx.data.logger.info('No links found in WebFinger response');
+            return error('no-links-found');
         }
 
-        const profileLink = webfingerData.links.find(
-            (link) => link.rel === 'http://webfinger.net/rel/profile-page',
-        );
-
-        let profileUrl: URL | null = null;
-        if (profileLink?.href) {
-            try {
-                profileUrl = new URL(profileLink.href);
-            } catch (err) {
-                ctx.data.logger.info(
-                    `Invalid profile page URL for handle ${handle}`,
-                );
-            }
-        }
-
-        // Get ActivityPub self link
+        // Find the ActivityPub self link
         const selfLink = webfingerData.links.find(
             (link) =>
                 link.rel === 'self' &&
                 link.type === 'application/activity+json',
         );
 
-        let apId: URL | null = null;
-        if (selfLink?.href) {
-            try {
-                apId = new URL(selfLink.href);
-            } catch (err) {
-                ctx.data.logger.info(
-                    `Invalid self link URL for handle ${handle}`,
-                );
-            }
-        }
-
-        if (!profileUrl && !apId) {
+        if (!selfLink?.href) {
             ctx.data.logger.info(
-                `No valid ActivityPub links found in WebFinger response for handle ${handle}`,
+                'No ActivityPub self link found in WebFinger response',
             );
+            return error('no-self-link');
         }
 
-        return { profileUrl: profileUrl, apId: apId };
+        return ok(new URL(selfLink.href));
     } catch (err) {
         ctx.data.logger.error(
-            `Error looking up actor profile for handle ${handle} - ${err}`,
+            'Error looking up actor by handle ({handle}): {error}',
+            { handle, error: err },
         );
-        return { profileUrl: null, apId: null };
+        return error('lookup-error');
     }
 }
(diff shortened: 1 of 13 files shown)


src/post/post.service.ts
```diff
@@ -176,6 +183,39 @@ export class PostService {
...
+            let account: Account | null = null;
+            const lookupResult = await lookupActorProfile(ctx, mention);
+            if (!lookupResult.apId) {
+                continue;
+            }
+            const accountResult = await this.accountService.ensureByApId(
+                lookupResult.apId,
+            );
+            if (isError(accountResult)) {
+                continue;
```
sagzy, line 201: Shall we log the error?

src/post/content.ts
```diff
@@ -125,6 +141,18 @@ export class ContentPreparer {
         return content + PAID_CONTENT_PREVIEW_HTML(url);
     }
 
+    private addMentions(content: string, mentions: Mention[]) {
+        let preparedContent = content;
+        for (const mention of mentions) {
+            const mentionRegex = new RegExp(mention.name, 'g');
```
sagzy, line 147: Think we need to escape `mention.name` here, otherwise it will match more than we want.

Example: `mention.name` is `@bob@my.site.com`

`new RegExp('@bob@my.site.com', 'g');` will also incorrectly match `@bob@myXsiteYcom` (as `.` matches any characters)

src/post/content.ts
```diff
@@ -170,6 +198,12 @@ export class ContentPreparer {
         return `${text.substring(0, charLimit - 3)}...`;
     }
 
+    private parseMentions(content: string): Set<string> {
+        const handleRegex = /@[\w.-]+@[\w-]+\.[\w.-]+/g;
+        const mentions = content.match(handleRegex) || [];
+        return new Set(mentions.filter(isHandle));
+    }
```
sagzy, line 205: We're checking against two regexes here (handleRegex, then the one defined in isHandle helper) - feels redundant?

Should we extract the regex in `isHandle` as a constant e.g. `export const HANDLE_REGEX = /^@([\w.-]+)@([\w-]+\.[\w.-]+[^.])$/` and then use it in `isHandle` and in `parseMentions`?

src/post/post.entity.unit.test.ts
```diff
@@ -212,6 +212,78 @@ describe('Post', () => {
             expect(note.content).toBe('<p>My first note</p>');
         });
 
+        it('creates a reply with mentions', () => {
+            const account = internalAccount();
+            const inReplyTo = Post.createNote(account, 'Parent');
+            // TODO: Clean up the any type
+            // biome-ignore lint/suspicious/noExplicitAny: Legacy code needs proper typing
+            (inReplyTo as any).id = 'fake-id';
```
sagzy, line 220: To avoid the `any`, we could do something like `(inReplyTo as unknown as { id: string }).id = 'fake-id';` - which is still a workaround, but at least avoid us adding more any to the codebase?

src/lookup-helpers.ts
```diff
@@ -110,7 +110,7 @@ export async function lookupAPIdByHandle(
 export async function lookupActorProfile(
     ctx: Context<ContextData>,
     handle: string,
-): Promise<URL | null> {
+): Promise<{ profileUrl: URL | null; apId: URL | null }> {
```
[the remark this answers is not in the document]
allouis: It's also worth noting that we are passing in the entire `Context<>` object - but we only need a logger!

src/post/post.entity.ts
```diff
@@ -221,6 +228,19 @@ export class Post extends BaseEntity {
         };
     }
 
+    addMention(account: Account) {
+        if (!account.id) {
```
allouis, line 232: This case is not possible

src/post/post.service.ts
```diff
@@ -176,6 +183,39 @@ export class PostService {
...
+            }
+            account = getValue(accountResult);
+
+            const profileUrl = lookupResult.profileUrl
+                ? lookupResult.profileUrl
+                : lookupResult.apId;
+
+            processedMentions.push({
+                name: mention,
+                href: profileUrl,
```
allouis, line 211: Can we use `account.url` here? This should be the profile url, and falls back to the ap id if not present

allouis (commented):
Left a few comments!

src/post/post.service.ts
sagzy: Oh - in that case, do we even need the new `lookupActorProfile` helper? https://github.com/TryGhost/ActivityPub/blob/804e17d18653802e5b8d599cd18a0c37aecb1635/src/lookup-helpers.ts#L110C1-L179C2

allouis: Yeah I think maybe we don't! Having said that, currently our `getApIdByHandle` lookup returns a `string` instead of a `URL` - so maybe we keep the `lookupActorProfile` and use it to return `Promise<Result<URL, LookupError>>` - and deprecate the old one? Not a big deal either way but does improve the codebase

vershwal: Thanks!! I had this same qs about the new `lookupActorProfile` helper.  
I'll create a new method that would return return `Promise<Result<URL, LookupError>>` and use it to return only `apId`.

src/post/post.entity.ts
vershwal: Removed the check here https://github.com/TryGhost/ActivityPub/pull/666/commits/4adb4d63b88e42c797ef80a882cf288e5f96fa1c#diff-3522562e572113f90d24c5500be2ae5bf5647a64867dfb920f6727a32a18facfL232

src/post/post.service.ts
vershwal: Did the changes in commit https://github.com/TryGhost/ActivityPub/pull/666/commits/4adb4d63b88e42c797ef80a882cf288e5f96fa1c

src/lookup-helpers.ts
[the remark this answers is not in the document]
vershwal: Did the changes in commit https://github.com/TryGhost/ActivityPub/commit/4adb4d63b88e42c797ef80a882cf288e5f96fa1c

src/post/post.entity.unit.test.ts
vershwal: Done in https://github.com/TryGhost/ActivityPub/pull/666/commits/ac30087a15cead75ae3a68a74bc6ae8eb5857d03

src/post/content.ts
vershwal: Done https://github.com/TryGhost/ActivityPub/pull/666/commits/ac30087a15cead75ae3a68a74bc6ae8eb5857d03

src/post/content.ts
vershwal: Great point. I also added a test for this and it failed 😅 Nice catch. Fixed here https://github.com/TryGhost/ActivityPub/pull/666/commits/ac30087a15cead75ae3a68a74bc6ae8eb5857d03

src/post/post.service.ts
vershwal: Done!

sagzy (approved):

PR #667, opened 2025-05-13
Fixed fetch failed errors

rmgpinto:
ref https://linear.app/ghost/issue/PROD-1587

- Wrapped code in a try/catch block and avoided throwing errors on ETIMEOUT or EHOSTUNREACH.

### DIFF
--- a/src/activity-handlers/create.handler.ts
+++ b/src/activity-handlers/create.handler.ts
@@ -24,7 +24,25 @@ export class CreateHandler {
             return;
         }
 
-        const sender = await create.getActor(ctx);
+        let sender = null;
+        try {
+            sender = await create.getActor(ctx);
+        } catch (err) {
+            if (
+                err instanceof Error &&
+                'code' in err &&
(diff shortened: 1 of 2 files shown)


src/activity-handlers/create.handler.ts
```diff
@@ -48,23 +41,23 @@ export class CreateHandler {
                             postId: create.objectId.href,
                         },
                     );
-                    break;
```
allouis, line 51: We don't want to continue if there is an error

src/activity-handlers/create.handler.ts
```diff
@@ -96,21 +89,22 @@ export class CreateHandler {
...
 
-        shouldAddToInbox = await isFollowedByDefaultSiteAccount(
-            sender,
-            site,
-            this.accountService,
+        const post = getValue(postResult);
+
+        const account = await this.accountService.getAccountForSite(site);
+
+        const isFollowing = await this.accountService.checkIfAccountIsFollowing(
```
allouis, line 102: Removes the dependency on `sender`

PR #671, opened 2025-05-14
Added mentions to `Create(Note)` activity

vershwal:
Ref https://linear.app/ghost/issue/PROD-1759

- Attached a tag field to the Note object that contains the Mentions
- Added the mentioned user in the cc field of the Create(Note) activity for both notes and replies

### DIFF
--- a/src/activitypub/fediverse-bridge.ts
+++ b/src/activitypub/fediverse-bridge.ts
@@ -6,7 +6,7 @@ import {
     Note as FedifyNote,
     Follow,
     Image,
-    type Mention,
+    Mention,
     PUBLIC_COLLECTION,
     Reject,
     Update,
@@ -19,7 +19,6 @@ import { addToList } from 'kv-helpers';
 import { PostCreatedEvent } from 'post/post-created.event';
 import { PostDeletedEvent } from 'post/post-deleted.event';
 import { PostType } from 'post/post.entity';
-import type { PostService } from 'post/post.service';
 import { v4 as uuidv4 } from 'uuid';
 import type { FedifyContextFactory } from './fedify-context.factory';
 
@@ -28,7 +27,6 @@ export class FediverseBridge {
         private readonly events: EventEmitter,
         private readonly fedifyContextFactory: FedifyContextFactory,
         private readonly accountService: AccountService,
-        private readonly postService: PostService,
     ) {}
 
     async init() {
@@ -65,7 +63,13 @@ export class FediverseBridge {
             if (post.inReplyTo) {
                 return;
             }
-            mentions = await this.postService.getMentionsForPost(post);
+            mentions = post.mentions.map(
+                (account) =>
+                    new Mention({
+                        name: `@${account.username}@${account.apId.hostname}`,
+                        href: account.apId,
+                    }),
+            );
             ccs = [
                 post.author.apFollowers,
                 ...mentions.map((mention) => mention.href),
--- a/src/activitypub/fediverse-bridge.ts
+++ b/src/activitypub/fediverse-bridge.ts
@@ -6,6 +6,7 @@ import {
     Note as FedifyNote,
     Follow,
     Image,
+    type Mention,
     PUBLIC_COLLECTION,
     Reject,
     Update,
@@ -18,6 +19,7 @@ import { addToList } from 'kv-helpers';
 import { PostCreatedEvent } from 'post/post-created.event';
 import { PostDeletedEvent } from 'post/post-deleted.event';
 import { PostType } from 'post/post.entity';
+import type { PostService } from 'post/post.service';
 import { v4 as uuidv4 } from 'uuid';
 import type { FedifyContextFactory } from './fedify-context.factory';
 
@@ -26,6 +28,7 @@ export class FediverseBridge {
         private readonly events: EventEmitter,
         private readonly fedifyContextFactory: FedifyContextFactory,
         private readonly accountService: AccountService,
+        private readonly postService: PostService,
     ) {}
 
     async init() {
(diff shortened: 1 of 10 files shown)


src/activitypub/fediverse-bridge.ts
```diff
@@ -106,7 +121,7 @@ export class FediverseBridge {
             actor: post.author.apId,
             object: fedifyObject,
             to: PUBLIC_COLLECTION,
-            cc: post.author.apFollowers,
+            ccs: ccs,
```
vershwal, line 124: We want to add cc in both `fedifyObject` and `createActivity`. Sample [Create activity from Mastodon](https://browser.pub/https%3A%2F%2Fmastodon.social%2Fusers%2Fprincivershwal%2Fstatuses%2F114506263585764479%2Factivity)

src/http/api/reply.ts
```diff
@@ -89,13 +90,12 @@ export async function handleCreateReply(
...
 
     const to = PUBLIC_COLLECTION;
-    const cc = [attributionActor, apCtx.getFollowersUri(ACTOR_DEFAULT_HANDLE)];
 
     const conversation = objectToReplyTo.replyTargetId || objectToReplyTo.id!;
     const mentions = [
         new Mention({
             href: attributionActor.id,
-            name: attributionActor.name,
+            name: getHandle(attributionActor),
```
vershwal, line 98: This wasn't really useful as it was not adding the handle here but just the name. See tag [here](https://browser.pub/https://princivershwal.ghost.io/.ghost/activitypub/note/c1c0ce3b-6dac-48f9-8ecc-ff502b21efe6)

src/post/post.service.ts
```diff
@@ -341,4 +341,8 @@ export class PostService {
 
         return ok(post);
     }
+
+    async getMentionsForPost(post: Post) {
+        return this.postRepository.getMentionsForPost(post);
```
vershwal, line 346: We don't want the controllers or helpers to directly access the `postRepository` so added this function here.

src/activitypub/fediverse-bridge.ts
```diff
@@ -95,7 +110,7 @@ export class FediverseBridge {
                 preview,
                 url: post.url,
                 to: PUBLIC_COLLECTION,
-                cc: post.author.apFollowers,
+                ccs: ccs,
```
sagzy, line 113: Curious about the use of `ccs` instead of `cc`

I can only find docs for `cc` (https://www.w3.org/TR/activitypub/#create-activity-outbox, https://www.w3.org/TR/activitystreams-vocabulary/#dfn-cc) and that seems to be what Mastodon & co also use?

src/post/post.repository.knex.ts
```diff
@@ -947,4 +948,26 @@ export class KnexPostRepository {
...
+                post_id: post.id,
+            });
+
+        return rows.map(
+            (row) =>
+                new Mention({
+                    name: `@${row.account_username}@${row.account_domain}`,
+                    href: new URL(row.account_apid),
+                }),
+        );
```
sagzy, line 971: Wonder if we should be mapping to Fedify's Mentions directly in PostRepository here 🤔 

Think I would have expected us to return an array of the internal `Mention` interface here, and then map that to its Fediverse equivalent in Fediverse Bridge?

src/activitypub/fediverse-bridge.ts
```diff
@@ -55,10 +58,19 @@ export class FediverseBridge {
...
         let fedifyObject: FedifyNote | Article;
 
+        let mentions: Mention[] = [];
+        let ccs: URL[] = [];
+
         if (post.type === PostType.Note) {
             if (post.inReplyTo) {
                 return;
             }
+            mentions = await this.postService.getMentionsForPost(post);
```
allouis, line 68: What about using `post.mentions` here? Having a `PostService#getMentionsForPost` doesn't seem right to me - maybe if it were a `MentionsService`...

But a mention does seem kinda core to a post at this point, especially as it could potentially trigger notifications/emails/etc... 

I would think something like `post.mentions: Account[]` or `post.mentions: Account["id"][]` would keep everything in the domain, and reduce the number of methods we need on the repository and services.

vershwal: Done [here](https://github.com/TryGhost/ActivityPub/pull/671/files#diff-3522562e572113f90d24c5500be2ae5bf5647a64867dfb920f6727a32a18facfR100)

src/post/post.repository.knex.ts
vershwal: Removed this method as now we have mentions added in post.

src/activitypub/fediverse-bridge.ts
vershwal: While creating the FedifyNote we can't use cc to add multiple URLs or Objects

This is how it's defined in constructor of Note Object :-

```
cc?: Object | URL | null;
ccs?: (Object | URL)[];
```

However, I know what you mean, when you look at an activity you'll see the array 'cc'. This is because Fedify internally converts ccs -> cc

I have created [this activity](https://princis-macbook-pro-2.tail5da2a.ts.net/.ghost/activitypub/note/07444f77-12fb-46f1-8af6-78fe899fc512) with my local code. I added all recipients in 'ccs' but in the activity you'll see 'cc' (similar to Mastodon)

sagzy: Ahhh it's a Fedify thing - sorry! Got confused

src/post/post.entity.ts
```diff
@@ -97,6 +97,7 @@ export class Post extends BaseEntity {
     private repostsToRemove: Set<number> = new Set();
     private deleted = false;
     public readonly content: string | null;
+    public readonly mentions: Account[] = [];
```
allouis, line 100: I think we need to populate this when reading from the repository too- I'm imagining a situation where the `PostCreatedEvent` switches to storing a `postId` instead of a full `Post` in future (because we are sending over a queue, and need to limit bandwidth e.g.), then we call `getById` and everything looks good, but federated mentions are silently failing 😢

allouis (changes requested):

src/post/post.entity.ts
vershwal: Added changes in https://github.com/TryGhost/ActivityPub/pull/671/commits/ed90d4e675370e0ef3329eccdfc4437857e89c68 and https://github.com/TryGhost/ActivityPub/pull/671/commits/cb7bc1b340a7108a7870352c03a0737282ff78c6

allouis (approved):

PR #672, opened 2025-05-14
Remove notifications when a post is deleted

vershwal:
Ref https://linear.app/ghost/issue/PROD-1764

- Removes all notifications for a post on PostDeletedEvent

### DIFF
--- a/src/notification/notification.service.integration.test.ts
+++ b/src/notification/notification.service.integration.test.ts
@@ -1023,6 +1023,10 @@ describe('NotificationService', () => {
             ]);
 
             const post = await fixtureManager.createPost(aliceAccount);
+            const replyPost = await fixtureManager.createReply(
+                bobAccount,
+                post,
+            );
 
             // Create multiple notifications for the post
             await Promise.all([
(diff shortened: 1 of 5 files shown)


src/notification/notification.service.ts
```diff
@@ -343,6 +343,10 @@ export class NotificationService {
             .delete();
     }
 
+    async removePostNotifications(post: Post) {
+        await this.db('notifications').where('post_id', post.id).delete();
```
allouis, line 347: Do we need to delete where `in_reply_to_post_id` matches too?

src/notification/notification.service.integration.test.ts
```diff
@@ -1014,4 +1014,59 @@ describe('NotificationService', () => {
...
+                ),
+                fixtureManager.createNotification(
+                    aliceAccount,
+                    bobAccount,
+                    NotificationType.Mention,
+                ),
+            ]);
+
+            // Add post_id to the notifications
+            await client('notifications')
```
allouis, line 1047: I think `createNotification` should be updated to accept the post the notification is for (if it's a post related notification) - what do you think?

allouis (commented):
Left a couple questions

src/notification/notification.service.ts
vershwal: Yes 💯 Thanks for pointing this out! Fixed in https://github.com/TryGhost/ActivityPub/pull/672/commits/756195618e85042570bef614556b45ac1a552554

src/notification/notification.service.integration.test.ts
vershwal: Yeah, made things a lot more easier. Added https://github.com/TryGhost/ActivityPub/pull/672/commits/756195618e85042570bef614556b45ac1a552554

PR #675, opened 2025-05-15
Handled incoming mentions

sagzy:
ref https://linear.app/ghost/issue/PROD-698

- we currently don't alter the incoming note: if it doesn't contain a properly formed hyperlink, it will not render as link in the feed

### DIFF
--- a/features/handle-mention.feature
+++ b/features/handle-mention.feature
@@ -1,0 +1,7 @@
+Feature: Incoming mentions
+
+    Scenario: We receive a Create(Note) with a mention
+        Given an Actor "Person(Alice)"
+        And a "Create(Note)" Activity "Note" by "Alice" with content "Hello @index@site.com" that mentions "Us"
+        When "Alice" sends "Note" to the Inbox
+        Then the request is accepted
--- a/src/post/post.service.ts
+++ b/src/post/post.service.ts
@@ -1,7 +1,7 @@
 import {
     Article,
-    Note,
     Mention as FedifyMention,
+    Note,
     lookupObject,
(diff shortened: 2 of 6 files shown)


features/handle-mention.feature
```diff
@@ -0,0 +1,7 @@
+Feature: Incoming mentions
+
+    Scenario: We receive a Create(Note) with a mention
+        Given an Actor "Person(Alice)"
+        And a "Create(Note)" Activity "Note" by "Alice" with content "Hello @index@site.com" that mentions "Us"
+        When "Alice" sends "Note" to the Inbox
+        Then the request is accepted
```
sagzy, line 7: I wanted to do:

Then `a notification of type "Mention" is created`, in which I request the notifications endpoint and check that a mention-type notification is present - but couldn't get that to work. Skipping it for now for the sake of getting the rest of the code in, I'll try to get that working in another PR

src/post/post.service.ts
```diff
@@ -79,6 +84,35 @@ export class PostService {
...
+            if (tag instanceof FedifyMention) {
+                if (!tag.href) {
+                    continue;
+                }
+
+                if (!tag.name) {
+                    continue;
+                }
+
+                const accountResult = await this.accountService.ensureByApId(
```
allouis, line 101: This can potentially make a network request, but as far as I can see it would only be in the case of:

1. An external Note
2. Mentions an external Account

In that case we probably don't actually care about the mention, but I don't think it's harmful to collect and store it anyway

allouis (approved):

PR #676, opened 2025-05-15
Fixed duplicate mention and reply notifications, when mentioning the author in a reply

sagzy:
ref https://linear.app/ghost/issue/PROD-698

- if I reply to Bob and also mention Bob in my reply, don't create a mention-type notification (as they're get a reply-type notification already)

### DIFF
--- a/src/post/post.entity.ts
+++ b/src/post/post.entity.ts
@@ -338,6 +338,11 @@ export class Post extends BaseEntity {
         );
 
         for (const mention of data.mentions ?? []) {
+            if (data.inReplyTo && mention.id === data.inReplyTo.author.id) {
+                // If Bob replies to my content and also mentions me, don't create a mention (only a reply)
+                continue;
+            }
+
             post.addMention(mention);
         }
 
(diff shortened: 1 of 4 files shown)


allouis (commented):
This puts notification logic/requirements into the Post entity, and it's implicit because it relies on the link between a Mention & a notification for a Mention, which isn't immediately obvious!

IMO This logic would sit a lot better inside of the NotificationsService - so we have a single place to look for all the logic of why/when/how a notification is created!

I think we have enough information in the `AccountMentionedEvent` to do this - we can read the Post from the `postId` on the event

Though this kinda links to what I mentioned in slack about using the `PostCreatedEvent`, where we could have a super explicit flow which details all of the notifications for an incoming post!
```
if (post.inReplyTo) {
 // create reply notification
}
for (mention in post.mentions) {
  if (mention.account === post.inReplyTo.author) continue;
  // create mention notification
}
```

What do you think about this? Does the logic of creating a Mention belong on the Post?

I think if this is the case, then we should remove the `Mention` that we create in `reply.ts` when we create replies 🤔 Ideally we can get to a point where there's an obvious 1-1 mapping between a `Post` object and the `Note` object which we federate

src/post/post.entity.ts
```diff
@@ -338,6 +338,11 @@ export class Post extends BaseEntity {
         );
 
         for (const mention of data.mentions ?? []) {
+            if (data.inReplyTo && mention.id === data.inReplyTo.author.id) {
```
vershwal, line 341: My concern with this approach is that we won't save the mentions in db. If someone mentions the author in a reply we won't have that mention saved in the db(although it's a valid mention) 

I agree with fabien that this should be handled on the notification level.

allouis (approved):

vershwal (approved):

PR #689, opened 2025-05-19
Added a `data-profile` attribute with the social web handle for mentions

vershwal:
Ref https://linear.app/ghost/issue/PROD-1785/

- In Feed and Profile views, we want to override mention hyperlinks so that the link opens in the Ghost app. 
- Added a `data-profile` attribute to the hyperlink when generating an outgoing Mention
- Frontend can override the default on-click behaviour using this `data-profile` attribute

### DIFF
--- a/src/post/post.service.integration.test.ts
+++ b/src/post/post.service.integration.test.ts
@@ -203,7 +203,7 @@ describe('PostService', () => {
 
             // Verify mention is wrapped in hyperlink in content
             expect(post.content).toBe(
-                `<p>This is a test note mentioning <a href="${mentionedAccount.apId}" rel="nofollow noopener noreferrer">@${mentionedAccount.username}@${mentionedAccount.apId.hostname}</a></p>`,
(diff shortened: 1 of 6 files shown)


src/post/post.service.integration.test.ts
```diff
@@ -203,7 +203,7 @@ describe('PostService', () => {
 
             // Verify mention is wrapped in hyperlink in content
             expect(post.content).toBe(
-                `<p>This is a test note mentioning <a href="${mentionedAccount.apId}" rel="nofollow noopener noreferrer">@${mentionedAccount.username}@${mentionedAccount.apId.hostname}</a></p>`,
+                `<p>This is a test note mentioning <a href="${mentionedAccount.apId}" data-profile="@${mentionedAccount.username}@${mentionedAccount.apId.hostname}" rel="nofollow noopener noreferrer">@${mentionedAccount.username}@${mentionedAccount.apId.hostname}</a></p>`,
```
sagzy, line 206: We can also use `mentionedAccount.domain` now

sagzy (approved):
LGTM!

src/post/post.service.integration.test.ts
vershwal: This is a nice suggestion sag but the tests are using [this function](https://github.com/TryGhost/ActivityPub/blob/main/src/account/account.repository.knex.ts#L202) to get the account and domain is not selected here. So I am gonna leave it like this for now.

PR #694, opened 2025-05-20
Add user row creation for existing accounts after site migration

betschki:
While testing ActivityPub, I discovered a bug affecting migrated users:

A user previously hosted their site on Ghost(Pro) and activated ActivityPub as `@index@example.com`. A user on my ActivityPub server then interacted with this account, creating an entry in the `accounts` table.

When the site was later migrated to Magic Pages and the ActivityPub beta was activated, a new entry was created in the `sites` table, as expected. Normally, a default user is created for the site, but in this case, user creation failed.

The `createOrFetchAccountAndUser` function tries to insert into `accounts` and then `users`. If the `accounts` insert fails due to a duplicate key (because the account already exists from the previous interaction on a different ActivityPub server), it fetches the existing account but skips creating the corresponding `user` row. This leaves the new site without a user linked to the account.

As a result, the newly created site is unusable: requests to `/.well-known/webfinger` return a 401, and Ghost UI requests fail since no user can be found for the site.

This fix aims to ensure that when an account already exists, a `user` row is always created linking the site to the account if it does not already exist.

I'm just starting out with understanding ActivityPub, so if there are any edge cases or other impacts I might have missed, I'm happy for any pointers.


allouis:
The tests passed, but GCP actions failed because they're run on a fork

allouis:
Hey @betschki just a thought - have you tested this flow?

It looks as though this will end up with a local Account - but they will be missing the `ap_private_key` - so will be unable to sign outgoing activities 🤔

I think we will need to generate a new keypair for the account in this case too

betschki:
@allouis thank you, good catch! I added this to the `AccountService` and also added test cases

allouis:
Thank you!!

PR #696, opened 2025-05-20
Removed writes to the Liked collection

sagzy:
ref https://linear.app/ghost/issue/PROD-1734/drop-support-for-liked-collection-until-ga

- we are dropping support for the Liked collection until GA, as it relies on the legacy key-value store and we don't have appetite to migrate it to the new database structure before GA


allouis (approved):
Hells yeah! Tests are way nicer :D

PR #697, opened 2025-05-20
Fixed `authoredByMe` property when reading local posts

allouis:
ref https://linear.app/ghost/issue/PROD-1638

We were incorrectly passing `accountId` instead of `defaultAccountId` to the `mapPostToDTO` method, which resulted in always having true for `authoredByMe`.

This meant that we showed the "delete" button on posts which aren't ours.

I've renamed `defaultAccountId` to `contextAccountId` so it's clearer what it is, added JSDoc to the `mapPostToDTO` method to explain the parameter and fixed which param we pass. I've also fixed the test which should have caught this to expect the correct data.

### DIFF
--- a/src/http/api/views/account.posts.view.ts
+++ b/src/http/api/views/account.posts.view.ts
@@ -110,7 +110,7 @@ export class AccountPostsView {
 
     async getPostsByAccount(
         accountId: number,
-        defaultAccountId: number,
+        contextAccountId: number,
         limit: number,
         cursor: string | null,
     ): Promise<AccountPosts> {
@@ -238,7 +238,7 @@ export class AccountPostsView {
                 ).andOnVal(
                     'current_user_likes.account_id',
                     '=',
-                    defaultAccountId.toString(),
+                    contextAccountId.toString(),
                 );
             })
             .leftJoin('reposts as current_user_reposts', function () {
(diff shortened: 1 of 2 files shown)


src/http/api/views/account.posts.view.ts
```diff
@@ -273,7 +273,7 @@ export class AccountPostsView {
 
         return {
             results: paginatedResults.map((result: GetProfileDataResultRow) =>
-                this.mapToPostDTO(result, accountId),
+                this.mapToPostDTO(result, contextAccountId),
```
allouis, line 276: This is the actual fix

src/http/api/views/account.posts.view.ts
```diff
@@ -604,7 +613,7 @@ export class AccountPostsView {
                 url: result.author_url ?? '',
                 avatarUrl: result.author_avatar_url ?? '',
             },
-            authoredByMe: result.author_id === accountId,
+            authoredByMe: result.author_id === contextAccountId,
```
sagzy, line 616: Another way of solving this could be to use the `post_reposted_by_current_user` / `post_liked_by_current_user` pattern, currently in use for `repostedByMe` / `likedByMe` - that way we don't need to pass any account ID to `mapToPostDTO` and that logic stay in the query?

sagzy (approved):
Nice catch 💯  

Added a suggestion for an alternative fix, but feel free to discard and merge!

PR #698, opened 2025-05-20
Fixed actor & keypair mismatch when sending activities

allouis:
ref https://linear.app/ghost/issue/PROD-1775

When calling `sendActivity` fedify uses the passed username to construct the id of the keypair, which is incorrect, because for existing account the id is always based on the "index" URL. We still only support one account per site, so we can safely hardcode the username which we pass to sendActivity.


sagzy (approved):
Looking good - think the rabbit is spot on in calling out using `username` over `handle` :)

PR #704, opened 2025-05-21
Updated Actor dispatcher to use identifier

allouis:
ref https://linear.app/ghost/issue/PROD-1775

Our actor dispatcher now correctly uses identifier, so that 'index' is the identifier. We then have a username/handle mapper which is currently hardcoded to 'index'. This allows us to remove a lot of the hardcoding elsewhere in the application, so that we can correctly pass `username`.

Now we're in a good position to move toward multiple accounts, as we will just need to update the `mapHandle` and `actorDispatcher` to correctly fetch from the DB and the `sendActivity` calls will all work as expected.

### DIFF
--- a/src/activity-handlers/follow.handler.ts
+++ b/src/activity-handlers/follow.handler.ts
@@ -89,7 +89,7 @@ export class FollowHandler {
             accountToFollow,
         );
 
-        await this.sendAccept(ctx, follow, parsed.handle, sender);
+        await this.sendAccept(ctx, follow, parsed.identifier, sender);
     }
 
     private async persistActivity(
@@ -113,7 +113,7 @@ export class FollowHandler {
     private async sendAccept(
         ctx: Context<ContextData>,
         follow: Follow,
-        handle: string,
+        identifier: string,
         sender: Actor,
     ): Promise<void> {
         const acceptId = ctx.getObjectUri(Accept, { id: uuidv4() });
@@ -126,13 +126,13 @@ export class FollowHandler {
 
         await ctx.data.globaldb.set([accept.id!.href], acceptJson);
 
-        await ctx.sendActivity({ username: 'index' }, sender, accept);
+        await ctx.sendActivity({ identifier }, sender, accept);
     }
 
     private async sendReject(
         ctx: Context<ContextData>,
         follow: Follow,
-        handle: string,
+        identifier: string,
         sender: Actor,
     ): Promise<void> {
         const rejectId = ctx.getObjectUri(Reject, { id: uuidv4() });
@@ -145,6 +145,6 @@ export class FollowHandler {
 
         await ctx.data.globaldb.set([reject.id!.href], rejectJson);
 
-        await ctx.sendActivity({ username: 'index' }, sender, reject);
+        await ctx.sendActivity({ identifier }, sender, reject);
     }
 }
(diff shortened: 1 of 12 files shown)


src/activity-handlers/follow.handler.ts
```diff
@@ -89,7 +89,7 @@ export class FollowHandler {
             accountToFollow,
         );
 
-        await this.sendAccept(ctx, follow, parsed.handle, sender);
+        await this.sendAccept(ctx, follow, parsed.identifier, sender);
```
allouis, line 92: `handle` is deprecated

src/activitypub/activity.ts
```diff
@@ -28,8 +28,11 @@ export class FedifyActivitySender implements ActivitySender<Activity, Actor> {
     constructor(private readonly fedifyCtx: FedifyRequestContext) {}
 
     async sendActivityToActorFollowers(activity: Activity, actor: Actor) {
+        if (!actor.preferredUsername) {
+            throw new Error(`Actor ${actor.id} has no preferred username`);
```
allouis, line 32: This should never happen - a Sentry error for this will alert us to a problem

src/app.ts
```diff
@@ -369,17 +365,20 @@ function ensureCorrectContext<B, R>(
 fedify
     // actorDispatcher uses RequestContext so doesn't need the ensureCorrectContext wrapper
     .setActorDispatcher(
-        '/.ghost/activitypub/users/{handle}',
+        '/.ghost/activitypub/users/{identifier}',
         spanWrapper(actorDispatcher(siteService, accountService)),
     )
+    .mapHandle(async () => {
+        return 'index'; // All identifiers are 'index'
```
allouis, line 372: In the future this will be something like:
```
if (isDefaultAccount) {
  return 'index';
} else {
  return uuid;
}
```

src/http/api/like.ts
```diff
@@ -48,54 +47,49 @@ export class LikeController {
             );
         }
 
-        const account = await this.accountRepository.getBySite(ctx.get('site'));
-        if (account !== null) {
```
allouis, line 52: `ctx.get('account')` is never null, so we don't need this check anymore

src/http/api/like.ts
```diff
@@ -201,39 +201,36 @@ export class LikeController {
...
```
allouis, line 205: Same as before, never null now

src/activitypub/fediverse-bridge.ts
```diff
@@ -139,7 +139,7 @@ export class FediverseBridge {
 
         await ctx.sendActivity(
             {
-                username: 'index',
+                username: post.author.username,
```
sagzy, line 142: If the post author has changed their username to `changed`, isn't this changing the call from

```
await ctx.sendActivity({username: 'index'}, ...)
``` 

to
```
await ctx.sendActivity({username: 'changed'}, ...)
```

?

src/app.ts
sagzy: nice

src/activitypub/fediverse-bridge.ts
allouis: Yeah that's right! But we have updated the actor dispatcher configured with fedify to work with identifiers now, so the flow is like this:

1. We call sendActivity with `{username: 'changed'}`
2. It's not the `identifier` so fedify will first call the `mapHandle` definition in `app.ts`
3. This returns `'index'`
4. Then fedify calls the actor dispatcher with `'index'`
5. We get the correct actor, with the correct identifier for the URLs!

sagzy: Ah 💡 Thanks for explaining, I understand now!

sagzy (approved):

PR #706, opened 2025-05-21
[WIP] Remove dependency on GET /inbox in feature tests

sagzy:
WIP


features/delete-post.feature
```diff
@@ -5,28 +5,21 @@ Feature: Delete a post
...
       """
       Hello
       World
       """
-    And an authenticated request is made to "/.ghost/activitypub/feed"
-    And the request is accepted
-    And "OurNote" is in the feed
-    And "AliceNote" is in the feed
+    And the note "OurNote" is in our feed
+    And the note "AliceNote" is in our feed
```
sagzy, line 14: The main change across most of these specs is:

Before
```
Given "Alice" sends "AliceNote" to the Inbox
And "AliceNote" is in our Inbox
And an authenticated request is made to "/.ghost/activitypub/feed"
And the request is accepted
And "AliceNote" is in the feed
```

After:
```
Given "Alice" sends "AliceNote" to the Inbox
Then the note "AliceNote" is in our feed
```

The "is in our Inbox" served as a wait step to make sure the activity was processed.
Now, as we're removing it, we need to "wait" for the note to be in our feed. The new step "the note {string} is our feed" uses the `waitForXInFeed()` method

PR #707, opened 2025-05-21
Incoming Mentions: Added `data-profile` and `rel` attributes to mention links

vershwal:
Ref https://linear.app/ghost/issue/PROD-1810/

The updateMentions function handles incoming mentions by:
 1. Adding `data-profile` and `rel` attributes to existing mention links

For existing links:
- Preserves the original href and other attributes
- Adds `data-profile` attribute with the mention handle
- Adds `rel="nofollow noopener noreferrer"`

For unwrapped mentions:
- Do Nothing

### DIFF
--- a/src/post/content.ts
+++ b/src/post/content.ts
@@ -1,5 +1,6 @@
 import { htmlToText } from 'html-to-text';
 import linkifyHtml from 'linkify-html';
+import { parse } from 'node-html-parser';
 import { HANDLE_REGEX } from '../constants';
 import type { Mention } from './post.entity';
 
@@ -76,6 +77,10 @@ export class ContentPreparer {
 
     static parseMentions(content: string) {
         return ContentPreparer.instance.parseMentions(content);
+    }
+
+    static updateMentions(content: string, mentions: Mention[]) {
+        return ContentPreparer.instance.updateMentions(content, mentions);
     }
 
     /**
@@ -205,6 +210,44 @@ export class ContentPreparer {
         const mentions =
             content.match(new RegExp(HANDLE_REGEX.source, 'g')) || [];
         return new Set(mentions);
+    }
+
+    /**
+     * Updates existing hyperlinks with data-profile attribute and rel attributes
+     */
+    private updateMentions(content: string, mentions: Mention[]) {
+        try {
+            const html = parse(content);
+            const links = html.querySelectorAll('a');
+
+            for (const mention of mentions) {
+                const apId = mention.account.apId
+                    .toString()
+                    .replace(/\/+$/, '');
+                const url = mention.account.url.toString().replace(/\/+$/, '');
+
+                // Find all links that matches the mentioned account
+                for (const link of links) {
+                    const href = link.getAttribute('href')?.replace(/\/+$/, '');
+                    if (href === apId || href === url) {
+                        // Update the link attributes
+                        link.setAttribute(
+                            'data-profile',
+                            `@${mention.account.username}@${mention.account.apId.hostname}`,
+                        );
+                        link.setAttribute(
+                            'rel',
+                            'nofollow noopener noreferrer',
+                        );
+                    }
+                }
+            }
+
+            return html.toString();
+        } catch (error) {
+            // If anything goes wrong during parsing or processing, return the original content
+            return content;
+        }
     }
 
     /**
--- a/package.json
+++ b/package.json
@@ -83,6 +83,7 @@
     "linkify-html": "4.3.1",
     "linkifyjs": "4.3.1",
     "mysql2": "3.14.1",
+    "node-html-parser": "7.0.1",
     "node-jose": "2.2.0",
     "sanitize-html": "2.17.0",
     "sharp": "0.34.1",
--- a/src/post/content.unit.test.ts
+++ b/src/post/content.unit.test.ts
@@ -417,4 +417,291 @@ describe('ContentPreparer', () => {
             expect(result).toEqual(new Set(['@user@example.com']));
         });
     });
+
+    describe('updateMentions', () => {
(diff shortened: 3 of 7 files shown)


src/post/content.ts
```diff
@@ -207,6 +211,77 @@ export class ContentPreparer {
...
+            const escapedHandle = mention.name.replace(
+                /[.*+?^=!:${}()|\[\]\/\\]/g,
+                '\\$&',
+            );
+
+            // Handling existing hyperlinks in content
+            const existingLinkRegex = new RegExp(
+                `<a[^>]*href=["'](?:${escapedAccountApId}|${escapedAccountUrl})["'][^>]*>.*?</a>`,
+                'g',
+            );
```
sagzy, line 237: There may be trailing slashes in the href:

- https://example.com/users/bob
- https://example.com/users/bob/

So may want to do something like 
```
const apId = mention.account.apId.toString().replace(/\/+$/, '');
const url = mention.account.url.toString().replace(/\/+$/, '');
```

before adding it to the regex?

src/post/content.ts
```diff
@@ -207,6 +211,77 @@ export class ContentPreparer {
...
+                preparedContent.includes(`<a href="${mention.account.apId}"`);
+
+            if (!hasExistingLink) {
+                const unwrappedMentionRegex = new RegExp(escapedHandle, 'g');
+                preparedContent = preparedContent.replace(
+                    unwrappedMentionRegex,
+                    `<a href="${mention.account.url}" data-profile="@${mention.account.username}@${mention.account.apId.hostname}" rel="nofollow noopener noreferrer">${mention.name}</a>`,
+                );
+            }
+        }
```
sagzy, line 280: This part feels a bit risky - if our hyperlink matching logic failed somehow, then we may be constructing invalid HTML (e.g. nested hyperlinks)?

src/post/content.ts
```diff
@@ -207,6 +211,77 @@ export class ContentPreparer {
...
+
+            // Replace existing hyperlinks with newly constructed hyperlinks with data-profile and rel attributes
+            preparedContent = preparedContent.replace(
+                existingLinkRegex,
+                (match) => {
+                    // Extract the href value from the existing link
+                    const hrefMatch = match.match(/href=["']([^"']*)["']/);
+                    const originalHref = hrefMatch
+                        ? hrefMatch[1]
+                        : mention.account.url;
```
sagzy, line 247: Aren't we potentially overriding the original href here?

src/post/content.ts
sagzy: I wonder if we should drop the requirement to wrap plaintext mentions into hyperlinks for now: Mastodon sends us mentions in hyperlinks, and we don't know which AP implementations sends plaintext mentions.

This feels easier / less risky to ship?

We can always back to this in context and then decide whether we want to support plaintext mentions or get the AP implementation to fix their context

sagzy (commented):
Generally, wonder if we can simplify the logic to:

- if there is a hyperlink in the original content
    - if there is no data-profile attribute already, add a `data-profile` attribute
    - if there is no rel attribute already, add a `rel` attribute
- otherwise, don't alter the content (i.e. drop the requirement to support plaintext mentions)

src/post/content.ts
vershwal: > then we may be constructing invalid HTML (e.g. nested hyperlinks)

You are right about this. I faced this while writing tests. Test case: if the added link is not an apId or url, this part of code will add a nested hyperlink.

Generally if we aren't sure if there is any AP implementations that sends plaintext mentions, I am happy to drop this requirement


> This feels easier / less risky to ship?

💯

src/post/content.ts
vershwal: Thanks for this! Nice suggestion

src/post/content.ts
vershwal: Removed the support for plaintext mentions for now

src/post/content.ts
```diff
@@ -207,6 +212,44 @@ export class ContentPreparer {
...
+    private updateMentions(content: string, mentions: Mention[]) {
+        try {
+            const html = parse(content);
+            const links = html.querySelectorAll('a');
+
+            for (const mention of mentions) {
+                const apId = mention.account.apId
+                    .toString()
+                    .replace(/\/+$/, '');
+                const url = mention.account.url.toString().replace(/\/+$/, '');
```
allouis, line 227: I would love to see this abstracted a little bit so this `updateMentions` is focused more on "what" than "how"

Maybe a utility like:
```ts
equals(a: URL | string, b: URL | string) {
  if (a instanceof URL) return equals(a.href, b);
  if (b instanceof URL) return equals(a, b.href);
  
  return a.replace(trailingSlashRegex, '') === b.replace(trailingSlashRegex, '');
}
```

```ts
for (mention of mentions) {
  for (link of links) {
    if (equals(mention.account.apId, link.href) || equals(mention.account.url, link.href)) {
      // update da link
    }
  }
}
```

src/post/content.ts
```diff
@@ -207,6 +212,44 @@ export class ContentPreparer {
...
+
+                // Find all links that matches the mentioned account
+                for (const link of links) {
```
allouis, line 230: We've got a nested loop here, which we could probably flatten - but honestly I think it's fine for now - I don't see us dealing with a number of a tags or mentions that's going to be an issue

allouis (approved):
Looks great! I left one comment about cleaning up the equality checks, but it's not a big deal

src/post/content.ts
vershwal: Added here ; https://github.com/TryGhost/ActivityPub/pull/716

PR #708, opened 2025-05-21
Support storing expiry in KnexKvStore

allouis:
ref https://linear.app/ghost/issue/PROD-685

We want to start cleaning up old keys from the key_value table, for example the idempotency keys created by fedify. The first step toward that is storing the expiry, and then we can run a cron style job to delete expired rows

### DIFF
--- a/src/knex.kvstore.integration.test.ts
+++ b/src/knex.kvstore.integration.test.ts
@@ -1,5 +1,6 @@
 import { beforeAll, describe, expect, it } from 'vitest';
 
+import { Temporal } from '@js-temporal/polyfill';
 import type { Knex } from 'knex';
 import { createTestDb } from 'test/db';
 import { KnexKvStore } from './knex.kvstore';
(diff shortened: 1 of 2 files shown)


src/knex.kvstore.integration.test.ts
```diff
@@ -87,4 +88,29 @@ describe('KnexKvStore', () => {
...
+
+        expect(row.expires).not.toBeNull();
+        expect(row.expires).toBeInstanceOf(Date);
+
+        const now = new Date();
+        const diff = row.expires.getTime() - now.getTime();
+
+        const differenceFromOneDay = Math.abs(diff - 1000 * 60 * 60 * 24);
+
+        expect(differenceFromOneDay).toBeLessThan(1000);
```
allouis, line 114: Testing locally I could get a few 100 milliseconds difference

sagzy (approved):
LGTM 👍

PR #713, opened 2025-05-21
Using `PostCreatedEvent` to create mention notifications

vershwal:
Ref https://linear.app/ghost/issue/PROD-1845

- Now that `mentions` are a part of post and we can access them directly using `post.mentions`, it is redundant to use another event for mentions notifications
- Using existing `PostCreatedEvent` to create mention notifications
- Deleted `AccountMentionedEvent`


allouis (approved):
🔥

PR #719, opened 2025-05-22
Supported running cleanup job on GCP

allouis:
ref https://linear.app/ghost/issue/PROD-685

Connecting to CloudSQL isn't done in the usual way, we need to use the cloud-sql-proxy, or a socket, or some other way of connecting. Using this proxy seemed like the easiest way to have a shared image for both local and cloud use

### DIFF
--- a/jobs/cleanup-expired-key-value-records/Dockerfile
+++ b/jobs/cleanup-expired-key-value-records/Dockerfile
@@ -1,8 +1,5 @@
 FROM mysql:8.3
 
-RUN curl -Lo /bin/cloud-sql-proxy https://storage.googleapis.com/cloud-sql-connectors/cloud-sql-proxy/v2.10.1/cloud-sql-proxy.linux.amd64
-RUN chmod +x /bin/cloud-sql-proxy
-
 COPY cleanup-expired-key-value-records /bin
 
 ENTRYPOINT ["cleanup-expired-key-value-records"]


jobs/cleanup-expired-key-value-records/cleanup-expired-key-value-records
```diff
@@ -1,9 +1,21 @@
 #!/bin/sh
 set -e
 
+PORT=${MYSQL_PORT:-3306}
+
+if [ -n "$INSTANCE_CONNECTION_NAME" ]; then
```
rmgpinto, line 6: if `MYSQL_SOCKET_PATH` is set: 
```
mysql \
   --socket="$MYSQL_SOCKET_PATH" \
   --user="$MYSQL_USER" \
   --password="$MYSQL_PASSWORD" \
   --database="$MYSQL_DATABASE"
...
```

else:

```
mysql \
   --host="$MYSQL_SOCKET_PATH" \
   --port="$MYSQL_PORT" \
   --user="$MYSQL_USER" \
   --password="$MYSQL_PASSWORD" \
   --database="$MYSQL_DATABASE"
...
```

jobs/cleanup-expired-key-value-records/Dockerfile
```diff
@@ -1,5 +1,8 @@
 FROM mysql:8.3
 
+RUN curl -Lo /bin/cloud-sql-proxy https://storage.googleapis.com/cloud-sql-connectors/cloud-sql-proxy/v2.10.1/cloud-sql-proxy.linux.amd64
```
rmgpinto, line 3: Not needed now 😄

allouis: Ooopsie, yup!

PR #725, opened 2025-05-26
Removed Inbox collection

sagzy:
ref https://linear.app/ghost/issue/PROD-683

- in the past, we used the GET `/inbox/:handle` endpoint to generate the Inbox/Feed feeds from the legacy key-value store
- now, we don't need the endpoint anymore, as we have new API endpoints (`/inbox`, `/feed`) reading from the new relational database
- this commit deletes the GET `/inbox/:handle` endpoint and related code


src/http/api/derepost.ts
```diff
@@ -111,7 +110,6 @@ export function createDerepostActionHandler(
         await ctx.get('globaldb').set([undo.id!.href], undoJson);
 
         // Remove announce activity from database
-        await removeFromList(ctx.get('db'), ['reposted'], announceId.href);
```
sagzy, line 114: Note for reviewer: this `'reposted'` collection doesn't seem to be in use:
- we never write to it
- we read from it in the Inbox collection (see `inboxHandler` function), but that code is now also deleted

So don't think we have any reason to keep those `removeFromList` calls for the `reposted` collection

src/helpers/activitypub/activity.ts
```diff
@@ -1,11 +1,3 @@
-import { createHash } from 'node:crypto';
-import { Announce, type Context, type KvStore, Like } from '@fedify/fedify';
-
-import type { ContextData } from '../../app';
-import { getActivityChildrenCount, getRepostCount } from '../../db';
-import { lookupActor } from '../../lookup-helpers';
-import { sanitizeHtml } from '../html';
-
 export interface ActivityObjectAttachment {
```
allouis, line 1: Are these used anywhere? Would be so nice to delete this whole file

allouis (approved):
🔥

PR #726, opened 2025-05-26
Added `outboxes` table

vershwal:
Ref https://linear.app/ghost/issue/PROD-1877

- Added `outboxes` table, this table contains an outgoing feed of posts, articles, reposts and replies.


allouis (commented):
I don't see that we have a way to filter reposts here - unless `item_type` should store that information?

I'm wondering if we want something like:
```
post_type TINYINT UNSIGNED NOT NULL. -- Filtering on Post/Article/etc...
outbox_type TINYINT UNSIGNED NOT NULL. -- Naming? Filtering on Repost/Reply/etc...
```

Do we need `in_reply_to` - if we have a flag that says it _is_ a reply?

allouis (approved):

PR #735, opened 2025-05-27
Removed gated content that is visible to members only

sagzy:
ref https://linear.app/ghost/issue/PROD-1883

- in Ghost, a publisher can create a CTA card with different visibility options: visible to public visitors, free members or paid members
- in ActivityPub, we want to display only content that is visible publicly
- this commit removes gated content that is visible to members only, as well as gated content markers (`<!--kg-gated-block:begin>...<!--kg-gated-block:end-->`)


allouis (approved):
Looks good! You might wanna add a test which has multiple blocks in

PR #736, opened 2025-05-27
Added support for `summary` to posts

allouis:
ref https://linear.app/ghost/issue/PROD-1884
ref https://codeberg.org/fediverse/fep/src/branch/main/fep/b2b8/fep-b2b8.md

We want to be able to render the custom_excerpt of a Ghost Post as a sort of subheading in the UI. To do this we're going to store it as a `summary` in the db, as well as on the created ActivityPub objects, as suggested in fep-b2b8


src/http/api/thread.unit.test.ts
```diff
@@ -133,6 +135,7 @@ describe('Thread API', () => {
                                 Audience.Public,
                                 'Test Post 789 (reply to Test Post 123)',
                                 'Test Post 789 Excerpt',
+                                null,
```
sagzy, line 138: Out of scope for this PR given - but wonder if we should be using named parameters to improve readability / not having to insert params in the same exact order everywhere?

sagzy (approved):
LGTM!

PR #737, opened 2025-05-27
Backfilled the `outboxes` table from `posts` & `reposts`

allouis:
ref https://linear.app/ghost/issue/PROD-1881

As part of the work to optimise reading the posts & reposts of an internal user we have created the `outboxes` table, which functions very similar to the `feeds` table. In order to use this table for existing users content we need to backfill it.


vershwal (approved):

vershwal (changes requested):
We'll have to update this migration to use `account_id` instead of `user_id`

vershwal (approved):

PR #745, opened 2025-05-28
Handled EXIF orientation for image uploads and added HEIC support

minimaluminium:
- Added automatic EXIF-based rotation using Sharp's rotate() method to fix 90-degree rotation issue with iPhone photos
- Added HEIC/HEIF formats to allowed image types
- Converted HEIC/HEIF images to JPEG format for web compatibility


vershwal (approved):
Looks good overall, It would be good to have a small unit test for this that can check if it converts HEIC/HEIF files to JPEG format.

PR #751, opened 2025-05-29
Added `getPostsFromOutbox` function to read profile posts from outbox

vershwal:
Ref https://linear.app/ghost/issue/PROD-1879/

- Returns posts in descending order (newest first)
- Filter out replies
- Includes repost author information for reposted posts
- Adds likedByMe and repostedByMe status for the context account
- Supports pagination for posts
- The endpoint already have e2e tests [here](https://github.com/TryGhost/ActivityPub/blob/main/features/profile-posts.feature) and all of these tests pass with the new query

### DIFF
--- a/src/http/api/account.ts
+++ b/src/http/api/account.ts
@@ -249,12 +249,24 @@ export function createGetAccountPostsHandler(
 
         // We are using the keyword 'me', if we want to get the posts of the current user
         if (handle === 'me') {
-            accountPosts = await accountPostsView.getPostsByAccount(
-                currentContextAccount.id,
-                currentContextAccount.id,
-                params.limit,
-                params.cursor,
-            );
+            const accountPostsResult =
+                await accountPostsView.getPostsFromOutbox(
+                    currentContextAccount,
+                    currentContextAccount.id,
+                    params.limit,
+                    params.cursor,
+                );
+            if (isError(accountPostsResult)) {
+                logger.error(`Error getting outbox for ${handle}`);
+                return new Response(
+                    JSON.stringify({
+                        posts: [],
+                        next: null,
+                    }),
+                    { status: 200 },
+                );
+            }
+            accountPosts = getValue(accountPostsResult);
         } else {
             const ctx = fedifyContextFactory.getFedifyContext();
             const apId = await lookupAPIdByHandle(ctx, handle);
--- a/src/http/api/views/account.posts.view.integration.test.ts
+++ b/src/http/api/views/account.posts.view.integration.test.ts
@@ -11,8 +11,10 @@ import { AsyncEvents } from 'core/events';
 import { getValue, isError } from 'core/result';
 import type { Knex } from 'knex';
 import { Audience, Post, PostType } from 'post/post.entity';
+import { KnexPostRepository } from 'post/post.repository.knex';
 import { generateTestCryptoKeyPair } from 'test/crypto-key-pair';
 import { createTestDb } from 'test/db';
+import { type FixtureManager, createFixtureManager } from 'test/fixtures';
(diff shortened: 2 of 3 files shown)


src/http/api/views/account.posts.view.ts
```diff
@@ -112,6 +110,11 @@ export class AccountPostsView {
         );
     }
 
+    /**
+     * Get posts by account
+     *
+     * @deprecated Use getPostsFromOutbox instead
+     */
```
vershwal, line 117: Will remove this in the next PR. On removing this, Github was creating too many diff between this and the new query.

src/http/api/views/account.posts.view.integration.test.ts
```diff
@@ -248,4 +220,151 @@ describe('AccountPostsView', () => {
...
+    describe('getPostsFromOutbox', () => {
+        let account: Account;
+        let contextAccount: Account;
+        let postRepository: KnexPostRepository;
+
+        beforeEach(async () => {
+            [account] = await fixtureManager.createInternalAccount();
+            [contextAccount] = await fixtureManager.createInternalAccount();
+            postRepository = new KnexPostRepository(db, events);
+        });
```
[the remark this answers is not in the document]
vershwal: Good point. Resolved in https://github.com/TryGhost/ActivityPub/pull/751/commits/de13fb94bcf28564eddfcd9a752ddcc2d0c01d13

src/http/api/views/account.posts.view.ts
```diff
@@ -286,6 +289,118 @@ export class AccountPostsView {
         };
     }
 
+    async getPostsFromOutbox(
+        account: Account,
+        contextAccountId: number,
+        limit: number,
+        cursor: string | null,
+    ): Promise<Result<AccountPosts, GetPostsError>> {
+        if (!account.isInternal) {
```
vershwal, line 298: Currently this will never be the case as this function is only called for internal accounts but I added this check so that if this function is used elsewhere, in the future, it gives an error while reading for external accounts.

allouis (commented):
Few comments on the error stuff!

Is the existing stuff covered by a test? I can't see a test for `getPostsByApId` - maybe I'm being blind. Or is there a test in cucumber? It would be really nice to know that the refactor is returning the same data

src/http/api/views/account.posts.view.ts
```diff
@@ -90,16 +92,14 @@ export class AccountPostsView {
         currentContextAccount: Account,
         limit: number,
         cursor: string | null,
-    ): Promise<Result<AccountPosts, GetPostsError>> {
+    ): Promise<Result<AccountPosts, GetPostsError | GetPostsFromOutboxError>> {
```
allouis, line 95: This should never return a `GetPostsFromOutboxError` I think - because this method should support both internal and external accounts. 

I don't think the caller of this method is in a position to be able to handle a `GetPostsFromOutboxError` - it means there's a bug in our code if they get one.

I think we need to check the return of the call to `getPostsFromOutbox` on line 98 - and if it's an error - we can `throw` an error, because at that point we're in an undefined state?

src/http/api/account.ts
```diff
@@ -249,12 +249,24 @@ export function createGetAccountPostsHandler(
...
+                    currentContextAccount.id,
+                    params.limit,
+                    params.cursor,
+                );
+            if (isError(accountPostsResult)) {
+                const error = getError(accountPostsResult);
+                switch (error) {
+                    case 'not-internal-account':
+                        logger.error(`Account is not internal for ${handle}`);
+                        return new Response(null, { status: 404 });
```
allouis, line 264: I think I would expect a 500 here - this means that something is wrong with our code 🤔

PR #758, opened 2025-05-29
Reduced boot time by bundling app code

rmgpinto:
closes https://linear.app/ghost/issue/PROD-675

- Reduced boot time by bundling app code. The service will be faster to serve requests on scaling events.


rmgpinto:
Waiting for dependency injection.

PR #782, opened 2025-06-04
Implemented ReplyChain API

allouis:
The commits have been broken down in a way that _should_ be easier to review, please let me know if they're not.

We have no support for pagination here, but this API is not used in production, only behind a flag.

Pagination is a little tricky because we have three ways of paginating, we have the need for more ancestors, more children, or deeper reply chains.

### DIFF
--- a/src/http/api/reply-chain.ts
+++ b/src/http/api/reply-chain.ts
@@ -1,244 +1,31 @@
-import { unsafeUnwrap } from 'core/result';
-import type { PostService } from 'post/post.service';
+import { exhaustiveCheck, getError, getValue, isError } from 'core/result';
 import type { AppContext } from '../../app';
-import { postToDTO } from './helpers/post';
+import { NotFound } from './helpers/response';
+import type { ReplyChainView } from './views/reply.chain.view';
 
 export class ReplyChainController {
-    constructor(private readonly postService: PostService) {}
+    constructor(private readonly replyChainView: ReplyChainView) {}
 
     async handleGetReplies(ctx: AppContext) {
         const account = ctx.get('account');
-        const postResult = await this.postService.getByApId(
+
+        const replyChainResult = await this.replyChainView.getReplyChain(
+            account.id,
             new URL(ctx.req.param('post_ap_id')),
         );
 
-        const post = unsafeUnwrap(postResult);
+        if (isError(replyChainResult)) {
+            const error = getError(replyChainResult);
+            switch (error) {
+                case 'not-found':
+                    return NotFound('Post not found');
+                default:
+                    return exhaustiveCheck(error);
+            }
+        }
 
-        const data = {
-            ancestors: {
-                chain: [
-                    {
-                        id: 'https://activitypub.ghost.org/.ghost/activitypub/article/6e36a7e3-f6b5-403a-bf40-3b8b1850afd8',
-                        type: 0,
-                        title: 'Metabolizing mentions',
-                        excerpt: '',
-                        summary:
-                            "We got those pesky little @ symbols doing the things they're supposed to",
-                        content:
-                            "We got those pesky little @ symbols doing the things they're supposed to",
-                        url: 'https://activitypub.ghost.org/metabolizing-mentions/',
-                        featureImageUrl:
-                            'https://activitypub.ghost.org/content/images/2025/05/johnonolan_a_lightly_scifi_themed_photograph_of_a_pug_astronaut_d454cc41-ce26-4ec1-8848-3a79a4524699.png',
-                        publishedAt: '2025-05-29T08:46:35.000Z',
-                        likeCount: 0,
-                        likedByMe: false,
-                        replyCount: 3,
-                        readingTimeMinutes: 1,
-                        attachments: [],
-                        author: {
-                            id: 'https://activitypub.ghost.org/.ghost/activitypub/users/index',
-                            name: 'Building ActivityPub',
-                            handle: '@index@activitypub.ghost.org',
-                            avatarUrl:
-                                'https://activitypub.ghost.org/content/images/2024/09/ghost-orb-white-squircle-07.png',
-                            url: 'https://activitypub.ghost.org/',
-                        },
-                        authoredByMe: false,
-                        repostCount: 0,
-                        repostedByMe: false,
-                        repostedBy: null,
-                        metadata: {
-                            ghostAuthors: [],
-                        },
-                    },
-                ],
-                next: null,
-            },
-            post: postToDTO(post, {
-                authoredByMe: post.author.id === account.id,
-                likedByMe: false,
-                repostedByMe: false,
-                repostedBy: null,
-            }),
-            children: [
-                {
-                    post: {
-                        id: 'https://www.jannis.io/.ghost/activitypub/note/9d8615b7-dfcb-4442-a1e4-969de84ce88e',
-                        type: 0,
-                        title: '',
-                        excerpt: '',
-                        summary: null,
-                        content:
-                            "<p>Awesome! My shoutout: I really enjoyed reading about the journey @jbaty@social.lol took exploring ActivityPub on Ghost. There's some really good feedback in his journey 🤗</p>",
-                        url: 'https://www.jannis.io/.ghost/activitypub/note/9d8615b7-dfcb-4442-a1e4-969de84ce88e',
-                        featureImageUrl: null,
-                        publishedAt: '2025-05-29T08:55:33.097Z',
-                        likeCount: 0,
-                        likedByMe: false,
-                        replyCount: 1,
-                        readingTimeMinutes: 1,
-                        attachments: [],
-                        author: {
-                            id: 'https://www.jannis.io/.ghost/activitypub/users/index',
-                            name: 'Jannis Fedoruk-Betschki',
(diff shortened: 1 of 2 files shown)


src/http/api/views/reply.chain.view.integration.test.ts
```diff
@@ -0,0 +1,168 @@
...
+ * The post will have an ancestor chain of MAX_ANCESTOR_DEPTH + 5 posts - to ensure pagination of ancestors
+ * The post will have MAX_CHILDREN_COUNT + 5 replies - to ensure pagination of children
+ * Each reply will have:
+ *  - If it is even indexed, it will have a number of replies equal to its index - this covers the case where a reply has no chain
+ *  - If it is odd indexed, it will have a reply chain of length equal to its index - this covers the case where a chain is paginated
+ */
+async function setupPosts(fixtureManager: FixtureManager) {
+    const [account] = await fixtureManager.createInternalAccount();
+
+    let postsToCreate = ReplyChainView.MAX_ANCESTOR_DEPTH + 6;
```
allouis, line 20: +6 because one of these posts becomes _the_ post, so this creates the ancestors and the "current"

src/http/api/views/reply.chain.view.integration.test.ts
```diff
@@ -0,0 +1,168 @@
...
+            }
+
+            const resultIds = children.map((c) => c.post_ap_id);
+
+            expect(resultIds).toEqual(expectedResults);
+        });
+    });
+
+    describe('getReplyChain', () => {
+        it('should return the reply chain for a post', async () => {
```
allouis, line 145: This test is a little shit for now - but the functionality is covered in the above two tests pretty extensively IMO

src/http/api/views/reply.chain.view.ts
```diff
@@ -0,0 +1,357 @@
...
+        post: PostDTO;
+        chain: PostDTO[];
+        next: string | null;
+    }[];
+    next: string | null;
+};
+
+export type ReplyChainError = 'not-found';
+
+const PostRowSchema = z.object({
```
allouis, line 23: Might be worth sharing this one day - looking for critique on where you might find DB rows which don't match this schema (and so will cause a 500 here - that is intentional)

src/http/api/views/reply.chain.view.ts
```diff
@@ -0,0 +1,357 @@
...
+                                .whereNotNull('ai.in_reply_to')
+                                .andWhere(
+                                    'ai.depth',
+                                    '<',
+                                    ReplyChainView.MAX_ANCESTOR_DEPTH,
+                                );
+                        });
+                })
+                .from('ancestor_ids')
+                .where('depth', '>', 0)
```
allouis: This means we don't both getting the current post, which is at depth 0 - only the ancestors

src/http/api/views/reply.chain.view.ts
```diff
@@ -0,0 +1,356 @@
...
+                                'posts.reply_count',
+                                'posts.in_reply_to',
+                                db.raw(
+                                    'ROW_NUMBER() OVER (ORDER BY created_at) AS top_level_child_index',
+                                ),
+                            )
+                                .from('posts')
+                                .orderBy('created_at', 'asc')
+                                .where('in_reply_to', postId)
+                                .limit(ReplyChainView.MAX_CHILDREN_COUNT + 1)
```
allouis: +1 so we can determine if there are more pages later

src/http/api/views/reply.chain.view.ts
```diff
@@ -0,0 +1,356 @@
...
+                                .join(
+                                    'child_ids as ci',
+                                    'p.in_reply_to',
+                                    'ci.id',
+                                )
+                                .where('ci.reply_count', '=', 1)
+                                .andWhere(
+                                    'ci.depth',
+                                    '<',
+                                    ReplyChainView.MAX_CHILDREN_DEPTH + 1,
```
allouis: +1 so we can determine if there are more pages later

src/http/api/reply-chain.ts
```diff
@@ -1,244 +1,31 @@
...
-    constructor(private readonly postService: PostService) {}
+    constructor(private readonly replyChainView: ReplyChainView) {}
 
     async handleGetReplies(ctx: AppContext) {
         const account = ctx.get('account');
-        const postResult = await this.postService.getByApId(
+
+        const replyChainResult = await this.replyChainView.getReplyChain(
+            account.id,
             new URL(ctx.req.param('post_ap_id')),
```
vershwal, line 14: I think we should have a check on `ctx.req.param('post_ap_id')` and return an error if it doesn't exits? 
Also what coderabbitai suggested, this might result in a 500 if `'post_ap_id'` is not a valid URL, in this case 400(bad req) is a more suitable response.

src/http/api/views/reply.chain.view.integration.test.ts
```diff
@@ -0,0 +1,168 @@
...
+            const { post } = await setupPosts(fixtureManager);
+
+            const replyChainView = new ReplyChainView(db);
+
+            // @ts-expect-error Property 'getChildren' is private and only accessible within class 'ReplyChainView'
+            await replyChainView.getChildren(post.author.id, post.id!);
```
vershwal: This call is not required. We are not using it's return value.

src/http/api/views/reply.chain.view.integration.test.ts
```diff
@@ -0,0 +1,168 @@
...
+    });
+
+    describe('getChildren', () => {
+        it('should return the paginated children of a post with replies chain', async () => {
+            const { ancestors, post, replies, chains } =
+                await setupPosts(fixtureManager);
+
+            const replyChainView = new ReplyChainView(db);
+            // @ts-expect-error Property 'getChildren' is private and only accessible within class 'ReplyChainView'
+            const children = await replyChainView.getChildren(
```
vershwal, line 121: These tests are checking small pieces of code, which is good. But I'm not sure if it is a right pattern for testing private methods. Maybe we could add tests that directly test the public function(`getReplyChain`) and from it's response, we can verify that the internal private methods are functioning correctly?

sagzy (approved):

src/http/api/views/reply.chain.view.ts
vershwal: DRY Alert!! 
Whenever we share it, we should extract this in a common place. We have this similar row structure defined at multiple places in the code.

src/http/api/views/reply.chain.view.ts
```diff
@@ -0,0 +1,356 @@
...
+type PostRow = z.infer<typeof PostRowSchema>;
+
+export class ReplyChainView {
+    static readonly MAX_ANCESTOR_DEPTH = 10;
+    static readonly MAX_CHILDREN_COUNT = 10;
+    static readonly MAX_CHILDREN_DEPTH = 5;
+
+    constructor(private readonly db: Knex) {}
+
+    private mapToPostDTO(result: PostRow, contextAccountId: number): PostDTO {
```
vershwal, line 68: DRY Alert again 😬 
Don't want you to change this immediately but this mapping is also something we have at multiple places in the code.

src/http/api/views/reply.chain.view.ts
```diff
@@ -0,0 +1,356 @@
...
+                    result.author_username,
+                ),
+                name: result.author_name ?? '',
+                url: result.author_url ?? '',
+                avatarUrl: result.author_avatar_url ?? '',
+            },
+            authoredByMe: result.author_id === contextAccountId,
+            repostCount: result.post_repost_count,
+            repostedByMe: result.post_reposted_by_current_user === 1,
+            repostedBy: null,
```
vershwal, line 104: Don't we need the reposter's details? I see that the query also doesn't fetch the reposter's details. I understand for the reply and chain we don't need this but for the current post? I am still thinking on this one(do we need reposter's details in any case)..maybe you have a clear idea.

src/http/api/reply-chain.ts
allouis: Yup totally agree - this is still WIP and focused on the happy path, but I can add those before merging

src/http/api/views/reply.chain.view.integration.test.ts
allouis: These tests allowed me to build this incrementally, one method at a time, and I think they're valuable tests, I wouldn't want to delete them.

We should absolutely be testing the `getReplyChain` method and checking everything through that - but these private methods are complex enough that I think we should test them too.

It does make this class more brittle, if we change the implementation, but keep the public interface the same, we can break the tests - but in this case I think it's a worthwhile trade off.

src/http/api/views/reply.chain.view.ts
allouis: I don't think we _can_ fetch this for the current post, because a reposter only exists in the context of a feed, where a post has been reposted.

The client has more context, and can retain the knowledge of the reposter if we render a reply chain in the context of a feed - but I'm not sure there's much we can do here on the server.

src/http/api/views/reply.chain.view.integration.test.ts
vershwal: Yes I do agree that these tests are valuable and we should have them!

PR #785, opened 2025-06-05
Added logic to update like/repost counts for remote posts

sagzy:
ref https://linear.app/ghost/issue/PROD-699/better-likerepost-counts-in-feedpost-view

- for remote posts, we're going to periodically fetch the remote object to get updated like / repost counts


src/post/post.repository.knex.ts
```diff
@@ -679,6 +679,24 @@ export class KnexPostRepository {
...
+    async updateInteractionCounts(
+        postId: number,
+        likeCount?: number,
+        repostCount?: number,
+    ) {
+        await this.db('posts').where({ id: postId }).update({
+            like_count: likeCount,
+            repost_count: repostCount,
+        });
+    }
```
sagzy: Note to reviewer: we're bypassing the Post entity here, as we don't want to remove the `readonly` constraint on `likeCount` / `repostCount`:

https://github.com/TryGhost/ActivityPub/blob/1786516e640907390a3c49ca95b09bb1cdc2a606/src/post/post.entity.ts#L114-L162

Curious if there is a better way we can think of?

sagzy: Continued here: https://ghost.slack.com/archives/C06TE92R3JR/p1749124819037289

src/post/post.repository.knex.ts
```diff
@@ -543,7 +545,13 @@ export class KnexPostRepository {
                     wasDeleted = true;
                 }
             } else {
-                if (likesToAdd.length > 0 || likesToRemove.length > 0) {
+                if (newLikeCount !== null) {
```
allouis: This `if/else` doesn't do exactly what we want it to do.

It says that if we've manually set the likeCount, then to skip the block of code which inserts/removes likes, and therefore skip the code which updates the `like_count` based on that.

But, we can save an internal post later, after not having set the likeCount - and that block will run - which is undesirable

Also, in theory, we can register a like on a post AND refresh it's remote counts in the same transaction, and with this if/else block - we'd lose the register of the like

I think instead what we want to do is unconditionally run the count which inserts/removes likes, and then we want to do something like

```
if it's an internal post
  set the like_count from likeCount
else if there are likes to update/remove
  set the like_count based on the current +- the diff
```

sagzy: Thank you! Fixed in [b5750f3](https://github.com/TryGhost/ActivityPub/pull/785/commits/b5750f388545c152e27549e501bc3846d52c6bd1) and [6e611f6](https://github.com/TryGhost/ActivityPub/pull/785/commits/6e611f60e316414c317168bb61cf44f41fee6ce2)

PR #786, opened 2025-06-05
Added pubsub events implementation to allow background processing of events

mike182uk:
ref https://linear.app/ghost/issue/PROD-1925

Added pubsub events implementation to allow background processing of events - This will primarily be utilised by the background post interactions count task, but also sets the ground work for moving our existing event processing (`AsyncEvents`) to use pubsub in the future

### DIFF
--- a/src/app.ts
+++ b/src/app.ts
@@ -233,39 +233,12 @@ let queue: GCloudPubSubPushMessageQueue | undefined;
 if (process.env.USE_MQ === 'true') {
     globalLogging.info('Message queue is enabled');
 
-    if (!process.env.MQ_PUBSUB_HOST) {
-        throw new Error('MQ_PUBSUB_HOST is not set');
-    }
-    if (!process.env.MQ_PUBSUB_PROJECT_ID) {
-        throw new Error('MQ_PUBSUB_PROJECT_ID is not set');
-    }
-    if (!process.env.MQ_PUBSUB_TOPIC_NAME) {
-        throw new Error('MQ_PUBSUB_TOPIC_NAME is not set');
-    }
-    if (!process.env.MQ_PUBSUB_GHOST_TOPIC_NAME) {
-        throw new Error('MQ_PUBSUB_GHOST_TOPIC_NAME is not set');
-    }
-    if (!process.env.MQ_PUBSUB_SUBSCRIPTION_NAME) {
-        throw new Error('MQ_PUBSUB_SUBSCRIPTION_NAME is not set');
-    }
-    if (!process.env.MQ_PUBSUB_GHOST_SUBSCRIPTION_NAME) {
-        throw new Error('MQ_PUBSUB_GHOST_SUBSCRIPTION_NAME is not set');
-    }
-
-    const pubSubClient = await initPubSubClient({
-        host: process.env.MQ_PUBSUB_HOST,
+    const pubSubClient = initPubSubClient({
+        host: process.env.MQ_PUBSUB_HOST || 'unknown_pubsub_host',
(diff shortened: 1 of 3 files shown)


allouis (approved):

src/app.ts
```diff
@@ -912,19 +939,21 @@ app.use(async (ctx, next) => {
...
-    let data: unknown;
-
-    try {
-        data = await ctx.req.json();
-    } catch (err) {
-        globalLogging.error('Failed to parse JSON: {err}', { err });
-
-        return new Response(null, { status: 400 });
-    }
+    const pubSubEvents = container.resolve('pubSubEvents');
```
allouis: We should be trying not to resolve dependencies manually here, maybe something like this would work
```ts

container.register('pubSubMessageHandler', asFunction(function (
    pubSubEvents,
    fedify,
    fedifyContextFactory,
    logging,
    fedifyKv,
  ) {
    return createIncomingPubSubMessageHandler(
       pubSubEvents,
       fedify,
       fedifyContextFactory,
       {
           globaldb: fedifyKv,
           logger: logging,
       },
       logging,
     ),
  }
).singleton());

app.post('/.ghost/activitypub/pubsub/ghost/push', async (ctx) => {
  const handler = spanWrapper(container.resolve('pubSubMessageHandler'));
  return handler(ctx);
});
```

src/events/pubsub-http.ts
```diff
@@ -0,0 +1,85 @@
...
+        data: z.string(),
+        attributes: z.record(z.string(), z.string()),
+    }),
+});
+
+export function createIncomingPubSubMessageHandler(
+    events: PubSubEvents,
+    fedify: Federation<ContextData>,
+    fedifyContextFactory: FedifyContextFactory,
+    newCtxData: ContextData,
```
allouis: Can we read this from the `Context` in the function below, instead of passing it in here?

src/events/event.ts
```diff
@@ -0,0 +1,32 @@
+export interface SerializableEvent {
+    toJSON(): Record<string, unknown>;
+}
+
+export interface DeserializableEventConstructor {
+    fromJSON(data: Record<string, unknown>): SerializableEvent;
+}
+
+export class EventSerializer {
```
allouis, line 9: I wonder, if we're just calling `toJSON` and `fromJSON` can we jsut delete this file altogether and have the PubSubEvents call those instead?

allouis (commented):
Few comments, overall structure looks good - but I think we can clean up signatgures - and rely a bit more on the awilix container

src/app.ts
mike182uk: Updated in https://github.com/TryGhost/ActivityPub/pull/786/commits/836e8d8ab5fe1e41aff3f1fecb6fd0730db0d56b

src/events/pubsub-http.ts
mike182uk: Updated in https://github.com/TryGhost/ActivityPub/pull/786/commits/83704c34621493adeb85a765e3619a0b622b3893

src/events/event.ts
mike182uk: Discussed on Slack, we need to be able to register events so that they can be deserialised correctly

src/events/pubsub-http.ts
mike182uk: Moved `globalDb` to earlier middleware in https://github.com/TryGhost/ActivityPub/pull/786/commits/a8bd5093edc3b67fed0ea5d5a83ada98a1f22059

PR #792, opened 2025-06-06
Mentions bug fix

vershwal:


mike182uk:
@vershwal Any reason we couldn't just modify the regex to account for the fact that handle-like strings may appear in a URL, and only replace on a word boundary? (i.e space before and after the handle)

Something like:

```
/(?<!\S)@mike@mike\.world(?!\S)/g
```

Executed against:

```
<p>Hello @mike@mike.world is this link to your profile correct? <a href="https://example.xyz/@mike@mike.world">@mike@mike.world</a></p>
```

Should correctly match the first `@mike@mike.world` but none of the others 🤔

vershwal:
> and only replace on a word boundary? (i.e space before and after the handle)

@mike182uk the reason we don't do this is because IRL people generally use mentions with punctuation and we want to acknoladge those mentions too:-

For example:-

`<p>Hello @mike@mike.world,  how are things today</p>`

or 

`<p>It's been great working with @index@activitypub.ghost.org.<p>`

And there can be many different cases. We have good tests for these cases [here](https://github.com/TryGhost/ActivityPub/blob/main/src/post/content.unit.test.ts#L441-L496). I want to make this fix without changing the existing regex as it works really well for all these cases.

mike182uk:
Ah, yeh good point, we could still accommodate that with the regex by doing some kind of look ahead:

```
(?<!\S)@mike@mike\.world(?!\w)
```

So match `@mike@mike.world` only when there is a space before and a non word character after, which works good for:

```
<p>Hello @mike@mike.world! is this link to your profile correct? <a href="https://example.xyz/@mike@mike.world">@mike@mike.world</a></p>
```

But this feels like it could end up becoming whack-a-mole trying to account for all edge cases 😅

Lets go with the URL replacement for now 👍

PR #795, opened 2025-06-06
Added support for the `next` param when reading replies

allouis:
ref https://linear.app/ghost/issue/PROD-1945

This just forwards on the `next` param to `getReplyChain` as well as exposes the `next` param based off of the published_at property.

### DIFF
--- a/src/http/api/views/reply.chain.view.ts
+++ b/src/http/api/views/reply.chain.view.ts
@@ -347,6 +347,7 @@ export class ReplyChainView {
     public async getReplyChain(
         accountId: number,
         postApId: URL,
+        cursor?: string,
     ): Promise<Result<ReplyChain, ReplyChainError>> {
         const selectPostRow = this.selectPostRow(accountId);
         const exists = await selectPostRow(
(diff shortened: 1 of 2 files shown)


src/http/api/views/reply.chain.view.ts
```diff
@@ -368,49 +369,51 @@ export class ReplyChainView {
...
+            allChildren.length > ReplyChainView.MAX_CHILDREN_COUNT;
+
+        const children = allChildren
+            .slice(0, ReplyChainView.MAX_CHILDREN_COUNT)
+            .map((child) => ({
+                post: child.post,
+                chain: child.chain.slice(0, ReplyChainView.MAX_CHILDREN_DEPTH),
+                next:
+                    child.chain.length > ReplyChainView.MAX_CHILDREN_DEPTH
+                        ? 'TODO'
```
mike182uk, line 403: This deliberate?

allouis: Yeah, but I think we're good with the cursor as the timestamp - so I will update this

allouis: Oh excuse me, this isn't the param for more children, but more replies. In which case yes - this is intentional, as there is no `next` param for them yet!

PR #802, opened 2025-06-09
Moved AccountUpdatedEvent into Entity

allouis:
ref https://linear.app/ghost/issue/PROD-1986

We don't want to emit an AccountUpdatedEvent for every time an entity is saved, we only want to do it when we have actually updated the data. This may also have knock-on effects where an account was updated, but in the context of another accounts request, and the event handler was attempting to call `sendActivity` but the keys used were from the wrong context, so it fails.

### DIFF
--- a/src/account/account.entity.unit.test.ts
+++ b/src/account/account.entity.unit.test.ts
@@ -6,6 +6,7 @@ import { AccountBlockedEvent } from './account-blocked.event';
 import { AccountFollowedEvent } from './account-followed.event';
 import { AccountUnblockedEvent } from './account-unblocked.event';
 import { AccountUnfollowedEvent } from './account-unfollowed.event';
+import { AccountUpdatedEvent } from './account-updated.event';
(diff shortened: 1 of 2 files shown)


src/account/account.entity.ts
```diff
@@ -142,19 +143,36 @@ export class AccountEntity implements Account {
...
+            this.events,
+        );
+
+        if (
+            account.username !== this.username ||
+            account.name !== this.name ||
+            account.bio !== this.bio ||
+            account.avatarUrl !== this.avatarUrl ||
+            account.bannerImageUrl !== this.bannerImageUrl
+        ) {
```
mike182uk: Hm, just a thought, we already do this check in the account service:

https://github.com/TryGhost/ActivityPub/blob/main/src/account/account.service.ts#L613-L623

allouis: True, I think I should remove the check there then!

PR #804, opened 2025-06-09
Added logic to trigger interaction count update, when an update is due

sagzy:
ref https://linear.app/ghost/issue/PROD-699

- for remote posts, we want to update interactions counts:
    - if they're actively rendered in feeds
    - at a lower frequency for older posts
    
- the update frequency is based on when the post was published:

| Post published            | Refresh interaction counts           |
|----------------------|----------------------------------|
| < 6 hours ago             | At most once every 10 minutes    |
| 6–24 hours ago          | At most once every 2 hours         |
| 1-7 days ago               | At most once every 6 hours         |
| > 7 days ago               | At most once per day                    |

### DIFF
--- a/src/post/post-interaction-counts.service.ts
+++ b/src/post/post-interaction-counts.service.ts
@@ -53,9 +53,9 @@ export class PostInteractionCountsService {
     }
 
     /**
-     * Computes whether the post is due an update of like/repost counts.
+     * Computes whether the post is due an update of like/repost counts,
+     * based on the following rules:
      *
-     * @table
      * | Post published       | Refresh interaction counts       |
      * |----------------------|----------------------------------|
      * | < 6 hours ago        | At most once every 10 minutes    |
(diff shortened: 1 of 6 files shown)


src/post/post-interaction-counts.service.ts
```diff
@@ -0,0 +1,98 @@
...
+     * @param {Date|null} updatedAt - The date and time the post was last updated
+     * @returns {boolean}
+     */
+    private isUpdateDue(publishedAt: Date, updatedAt: Date | null): boolean {
+        let lastUpdate = updatedAt;
+        if (lastUpdate === null) {
+            lastUpdate = publishedAt;
+        }
+
+        const timeSinceLastUpdate = new Date().getTime() - lastUpdate.getTime();
```
mike182uk: We could extract this out so we don't have to construct it twice:

```ts
const now = Date.now()

const timeSinceLastUpdate = now - lastUpdate.getTime()
const timeSincePublished = now - publishedAt.getTime()
```

src/post/post-interaction-counts.service.ts
```diff
@@ -0,0 +1,98 @@
...
+     * Computes whether the post is due an update of like/repost counts.
+     *
+     * @table
+     * | Post published       | Refresh interaction counts       |
+     * |----------------------|----------------------------------|
+     * | < 6 hours ago        | At most once every 10 minutes    |
+     * | 6–24 hours ago       | At most once every 2 hours       |
+     * | 1-7 days ago         | At most once every 6 hours       |
+     * | > 7 days ago         | At most once per day             |
+     *
```
mike182uk: Nice, although is `@table` a valid JSDoc thing? (either way the table is still useful)

src/post/post-interaction-counts.service.unit.test.ts
```diff
@@ -0,0 +1,278 @@
...
+            ).not.toHaveBeenCalled();
+            expect(mockLogger.error).toHaveBeenCalledWith(
+                'Post with ID {postId} not found when updating interaction counts - Skipping',
+                { postId },
+            );
+        });
+
+        it('logs an error if updating interaction counts fails', async () => {
+            const postId = 1;
+            const publishedAt = new Date(Date.now() - 2 * 60 * 60 * 1000);
```
mike182uk: It might be good to put these time values into consts to make it easier to read / re-use

```ts
const TWO_HOURS = 2 * 60 * 60 * 1000;
```

mike182uk (approved):

src/post/post-interaction-counts.service.ts
sagzy: Ah good one - I've seen it recently in some code but it doesn't seem to be part of JSDoc core, removed!

src/post/post-interaction-counts.service.ts
sagzy: Good one, improved in d91d5ecf3b0a76f5a712d5bbc4e88523ac284d5f

src/post/post-interaction-counts.service.unit.test.ts
sagzy: Good one, improved readability in d91d5ecf3b0a76f5a712d5bbc4e88523ac284d5f

PR #807, opened 2025-06-10
Improved error logging for parent post fetching for replies

vershwal:
ref https://linear.app/ghost/issue/PROD-1917/

- Currently if there is an error while fetching the thread for a reply it silently fails.
- Added logs and thrown error so that we can track these failures.
- Earlier we used to save the reply without complete thread
- We won't save the reply now and exit with an error


src/post/post.service.ts
```diff
@@ -180,14 +181,27 @@ export class PostService {
...
+                        Sentry.captureException(errorMessage);
+                        context.data.logger.error(errorMessage);
                         break;
                     case 'missing-author':
+                        errorMessage = `Parent post for reply ${foundObject.id}, parent id : ${foundObject.replyTargetId}, has no author`;
+                        Sentry.captureException(errorMessage);
+                        context.data.logger.error(errorMessage);
                         break;
                     default: {
+                        errorMessage = `Unknown error while fetching parent post for reply ${foundObject.id}, parent id : ${foundObject.replyTargetId}, error : ${error}`;
```
allouis: I don't think we need logging here - this branch should never be called, but if it is then the `exhaustiveCheck` will explode and throw an error - which will cause a 500 etc... and notify sentry

src/post/post.service.ts
```diff
@@ -180,14 +181,27 @@ export class PostService {
             const found = await this.getByApId(foundObject.replyTargetId);
             if (isError(found)) {
                 const error = getError(found);
+                let errorMessage: string;
                 switch (error) {
                     case 'upstream-error':
+                        errorMessage = `Failed to fetch parent post for reply ${foundObject.id}, parent id : ${foundObject.replyTargetId}`;
+                        Sentry.captureException(errorMessage);
```
allouis: I'm not 100% sure how Sentry behaves if we pass a string to it - everywhere else we pass an instance of `Error` 

I think it's probably worth creating an instance of error here to pass to Sentry - just to be safe

src/post/post.service.ts
```diff
@@ -180,14 +181,27 @@ export class PostService {
             const found = await this.getByApId(foundObject.replyTargetId);
             if (isError(found)) {
                 const error = getError(found);
+                let errorMessage: string;
```
allouis, line 184: `errorMessage` is defined here, rather than in the `case` statements - but we don't use it outside of the case statements.

What if we take advantage of the scope of `errorMessage` and do something like:

```ts
let errorMessage: string;
switch (error) {
  case 'something':
    errorMessage = 'something';
    break;
  // etc...
}

const err = new Error(errorMessage);
ctx.data.logging.error(errorMessage);
Sentry.captureException(err);
```

src/post/post.service.ts
vershwal: Done

src/post/post.service.ts
vershwal: This is really neat. Done https://github.com/TryGhost/ActivityPub/pull/807/commits/5764543dcdaf3d411e7e0d5ad32f098bbecc597a

allouis (approved):

PR #812, opened 2025-06-10
Fixed `hasMore` when ancestors is empty

allouis:
If the ancestors array is empty, `hasMore` will be true as undefined !== null


vershwal (approved):
Looks good, nice tests 😊

PR #827, opened 2025-06-11
Added API endpoint for count of unread notifications

vershwal:
closes https://linear.app/ghost/issue/PROD-2002/

- We want to show a count of unread notifications to the user on AP UI

### DIFF
--- a/features/notifications.feature
+++ b/features/notifications.feature
@@ -6,3 +6,19 @@ Feature: Notifications
   Scenario: Requests for notifications with limit over 100 are rejected
     When an authenticated request is made to "/.ghost/activitypub/notifications?limit=200"
     Then the request is rejected with a 400
+
+  Scenario: Requests for unread notifications count
+    Given an Actor "Person(Alice)"
+    And an Actor "Person(Bob)"
+    And we publish an article
+    And "Bob" likes our article
+    And the like is in our notifications
+    And "Alice" likes our article
+    And the like is in our notifications
+    And "Bob" sends us a reply to our article
+    And the reply is in our notifications
+    And "Alice" sends us a reply to our article
+    And the reply is in our notifications
+    When an authenticated request is made to "/.ghost/activitypub/notifications/unread/count"
+    Then the request is accepted with a 200
+    And the unread notifications count is 4
--- a/features/notifications.feature
+++ b/features/notifications.feature
@@ -12,13 +12,7 @@ Feature: Notifications
     And an Actor "Person(Bob)"
     And we publish an article
     And "Bob" likes our article
-    And the like is in our notifications
     And "Alice" likes our article
-    And the like is in our notifications
     And "Bob" sends us a reply to our article
-    And the reply is in our notifications
     And "Alice" sends us a reply to our article
-    And the reply is in our notifications
-    When an authenticated request is made to "/.ghost/activitypub/notifications/unread/count"
-    Then the request is accepted with a 200
-    And the unread notifications count is 4
+    Then the unread notifications count is 4
--- a/src/app.ts
+++ b/src/app.ts
@@ -114,6 +114,7 @@ import {
     createGetNotificationsHandler,
     createGetPostHandler,
     createGetThreadHandler,
+    createGetUnreadNotificationsCountHandler,
     createPostPublishedWebhookHandler,
     createSearchHandler,
     createStorageHandler,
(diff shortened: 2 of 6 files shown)


src/http/api/notification.ts
```diff
@@ -116,3 +117,44 @@ export function createGetNotificationsHandler(
...
+            switch (error) {
+                case 'not-internal-account':
+                    ctx.get('logger').error(
+                        `User not found for account ${account.id}`,
+                    );
+                    return new Response(null, { status: 500 });
+                default:
+                    return exhaustiveCheck(error);
+            }
+        }
```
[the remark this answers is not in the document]
vershwal: I would expect a 500 here as this means that something is wrong with our code. We should always find an user for an account that is requesting unread notification count.

features/notifications.feature
```diff
@@ -6,3 +6,19 @@ Feature: Notifications
   Scenario: Requests for notifications with limit over 100 are rejected
     When an authenticated request is made to "/.ghost/activitypub/notifications?limit=200"
     Then the request is rejected with a 400
+
+  Scenario: Requests for unread notifications count
```
allouis, line 10: We want to move away from have the implementation details in the cucumber tests!

So ideally we would get rid of `And the like is in our notifications` and `When an authenticated request is made to "/.ghost/activitypub/notifications/unread/count` and have something like

```
Given we receive a like from Alice
And we receive a like from Bob
And we receive a reply from Alice
And we receive a reply from Bob
Then our unread notifications count is 4
```

allouis: This is just an idea, maybe even there's a single step for the setup idk - @mike182uk might have better ideas of what the test looks like - but the point is that we're making it simpler, and pushing the implementation into the step defs

src/app.ts
```diff
@@ -547,6 +548,16 @@ container.register(
     ).singleton(),
 );
 
+container.register(
+    'getUnreadNotificationsCountHandler',
+    asFunction((accountService, notificationService) =>
```
allouis: NABD - You should be able to do `asFunction(createGetUnreadNotificationsCountHandler)` here instead

src/http/api/notification.ts
```diff
@@ -116,3 +117,44 @@ export function createGetNotificationsHandler(
...
+    accountService: AccountService,
+    notificationService: NotificationService,
+) {
+    /**
+     * Handle a request for a user's unread notifications count
+     *
+     * @param ctx App context instance
+     */
+    return async function handleGetUnreadNotificationsCount(ctx: AppContext) {
+        const account = await accountService.getDefaultAccountForSite(
```
allouis: This uses the old `AccountType` - we should instead use `ctx.get('account')` which gives back the `Account` for the currently authenticated user

src/http/api/notification.ts
```diff
@@ -116,3 +117,44 @@ export function createGetNotificationsHandler(
...
+
+        const unreadNotificationsCountResult =
+            await notificationService.getUnreadNotificationsCount(account.id);
+
+        if (isError(unreadNotificationsCountResult)) {
+            const error = getError(unreadNotificationsCountResult);
+            switch (error) {
+                case 'not-internal-account':
+                    ctx.get('logger').error(
+                        `User not found for account ${account.id}`,
```
allouis: NABD - Something more descriptive here might help us identify the error faster if it ever happens, like `Cannot get notifications count for external account ${account.id}`

features/notifications.feature
vershwal: Done here https://github.com/TryGhost/ActivityPub/pull/827/commits/150396f481123c88e40453dd2892e5aa805ab869

src/app.ts
vershwal: Done https://github.com/TryGhost/ActivityPub/pull/827/commits/f3f97aa479e446af97c88c7b2aa11dd87b40e67e

src/http/api/notification.ts
vershwal: Yeah, I missed this. Done here https://github.com/TryGhost/ActivityPub/pull/827/commits/f3f97aa479e446af97c88c7b2aa11dd87b40e67e

src/http/api/notification.ts
vershwal: Done

src/http/api/notification.ts
```diff
@@ -49,12 +42,12 @@ export function createGetNotificationsHandler(
             });
         }
 
-        const account = await accountService.getDefaultAccountForSite(
+        const account = await this.accountService.getDefaultAccountForSite(
```
allouis, line 45: If you're changing this line we might as well use `ctx.get('account')` here too - nabd tho

vershwal: Will add a small tidying PR for this 😊

PR #833, opened 2025-06-12
Added API endpoint for resetting unread notifications count

vershwal:
closes https://linear.app/ghost/issue/PROD-2004/

### DIFF
--- a/src/account/account.entity.ts
+++ b/src/account/account.entity.ts
@@ -28,7 +28,7 @@ export interface Account {
     unblockDomain(domain: URL): Account;
     follow(account: Account): Account;
     unfollow(account: Account): Account;
-    resetUnreadNotificationsCount(): Account;
+    readAllNotifications(): Account;
     /**
      * Returns a new Account instance which needs to be saved.
      */
@@ -231,7 +231,7 @@ export class AccountEntity implements Account {
         );
     }
 
-    resetUnreadNotificationsCount(): Account {
+    readAllNotifications(): Account {
         return AccountEntity.create(
             this,
             this.events.concat(new NotificationsReadEvent(this.id)),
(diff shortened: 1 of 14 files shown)


src/account/account.repository.knex.ts
```diff
@@ -10,6 +10,7 @@ import { AccountUnfollowedEvent } from './account-unfollowed.event';
 import { type Account, AccountEntity } from './account.entity';
 import { DomainBlockedEvent } from './domain-blocked.event';
 import { DomainUnblockedEvent } from './domain-unblocked.event';
+import { NotificationReadEvent } from './notifications-read-event';
```
allouis: The filename and eventname are plural, but the class name is singular - maybe we should rename it to `NotificationsReadEvent` 🤔

src/account/account.repository.knex.ts
```diff
@@ -142,6 +143,20 @@ export class KnexAccountRepository {
                             following_id: event.getAccountId(),
                         })
                         .delete();
+                } else if (event instanceof NotificationReadEvent) {
```
allouis: It's not a big deal though, but you might consider putting this logic in the notifications service so it is all grouped together. Let's imagine a world where we split notifications out into a separate deployment/api/whatever - we wouldn't have to touch the account code. 

We don't do that for blocks & follows, but if we had a block/follows service, we might want to!

src/account/account.entity.ts
```diff
@@ -27,6 +28,7 @@ export interface Account {
     unblockDomain(domain: URL): Account;
     follow(account: Account): Account;
     unfollow(account: Account): Account;
+    resetUnreadNotificationsCount(): Account;
```
allouis: Not a big deal, but this kinda describes the result, rather than the what - `markNotificationsAsRead` or `readAllNotifications` might be more appropriate

allouis (approved):

PR #836, opened 2025-06-12
Added logging when remote post does not expose interaction counts

sagzy:
ref https://linear.app/ghost/issue/PROD-699/better-likerepost-counts-in-feedpost-view

- before, when the remote object didn't expose like/repost counts, we would log a misleading "Successfully updated interaction counts for post with ID {postId}"
- now, we instead log "Post with ID {postId} does not expose interaction counts - Skipping"
- also, renamed `postService.update()` to `postService.updateInteractionCounts()` to be more explicit

### DIFF
--- a/src/post/post.service.ts
+++ b/src/post/post.service.ts
@@ -55,11 +55,13 @@ export type RepostError =
 
 export type GhostPostError = CreatePostError | 'post-already-exists';
 
+export const INTERACTION_COUNTS_NOT_FOUND = 'interaction-counts-not-found';
 export type UpdateInteractionCountsError =
     | 'post-not-found'
     | 'post-is-internal'
     | 'upstream-error'
-    | 'not-a-post';
+    | 'not-a-post'
+    | typeof INTERACTION_COUNTS_NOT_FOUND;
 
 export class PostService {
     constructor(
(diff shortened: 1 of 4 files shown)


src/post/post.service.ts
```diff
@@ -432,7 +434,7 @@ export class PostService {
         return ok(post);
     }
 
-    async update(
+    async updateInteractionCounts(
```
sagzy, line 437: @mike182uk in the Post Service, I think it's good to keep the more explicit `updateInteractionCounts` (versus in the `PostInteractionCountsService` it's okay to just call it `update()`)

src/post/post.service.ts
```diff
@@ -456,6 +458,10 @@ export class PostService {
         const likeCount = await getLikeCountFromRemote(object);
         const repostCount = await getRepostCountFromRemote(object);
 
+        if (likeCount === null && repostCount === null) {
+            return error(INTERACTION_COUNTS_NOT_FOUND);
+        }
+
```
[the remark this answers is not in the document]
sagzy: ^ that's a good point

@mike182uk what do you think of adding a more permanent boolean flag to the `posts` e.g. `skipInteractionCountsUpdate` that is set to true once we know the remote object doesn't expose the counts? So that it's skipped going forward?

Bit polluting to the `posts` table, but looks like an easy way to make fewer unnecessary network calls?

src/post/post.service.ts
mike182uk: My bad, i was a bit trigger happy with the previous rename 🤠

PR #838, opened 2025-06-12
Refactored GCP storage service to allow for more storage adapters

sagzy:
ref https://linear.app/ghost/issue/PROD-2017

- we want to enable other storage adapters for self-hosters
- before doing so, the current code needs a bit of refactoring to be able to easily plug in different storage adapters


src/app.ts
```diff
@@ -233,6 +236,39 @@ try {
...
 }
 
+container.register(
+    'storageAdapter',
+    asFunction(() => {
+        const bucketName = process.env.GCP_BUCKET_NAME || '';
+        return new GCPStorageAdapter(
+            bucketName,
+            process.env.GCP_STORAGE_EMULATOR_HOST ?? undefined,
+        );
```
sagzy, line 235: As we add more storage adapters, I imagine this block could select the right adapter based on config

src/app.ts
```diff
@@ -233,6 +236,39 @@ try {
...
+} catch (err) {
+    globalLogging.error('Failed to initialise storage adapter {error}', {
+        error: err,
+    });
+    process.exit(1);
+}
+
+container.register(
+    'imageProcessor',
+    asFunction((logging) => new ImageProcessor(logging)).singleton(),
```
allouis: You can just use `asClass(ImageProcessor)` here!

src/app.ts
```diff
@@ -233,6 +236,39 @@ try {
...
+);
+
+container.register(
+    'imageStorageService',
+    asFunction(
```
allouis: Same as above `asClass(ImageStorageProcessor)`!

src/app.ts
allouis: Exactly my thinking too

src/app.ts
```diff
@@ -233,6 +236,39 @@ try {
...
+        const bucketName = process.env.GCP_BUCKET_NAME || '';
+        return new GCPStorageAdapter(
+            bucketName,
+            process.env.GCP_STORAGE_EMULATOR_HOST ?? undefined,
+        );
+    }).singleton(),
+);
+
+try {
+    await container.resolve('storageAdapter').init();
```
allouis: I think we should get rid of this, and instead handle missing buckets as a 500/runtime error.

We've done the same with the PubSub initialisation checks (cc @mike182uk)

The reason being - just because it exists at boot, doesn't mean it exists later - so we should handle it missing later anyway. And also we should be resilient to the external services being down/misconfigured - just because there's no storage doesn't mean we can't serve the feed/profiles/etc....

src/http/api/image-upload.ts
```diff
@@ -0,0 +1,53 @@
...
+
+        if (!file || !(file instanceof File)) {
+            return new Response('No valid file provided', { status: 400 });
+        }
+
+        const account = await accountService.getAccountForSite(ctx.get('site'));
+        const isHeicFile =
+            file.type.includes('heic') || file.type.includes('heif');
+        const extension = isHeicFile ? 'jpg' : file.type.split('/')[1];
+        const path = `/images/${account.uuid}/${uuidv4()}.${extension}`;
```
allouis: Would love to see this handled by the `imageStorageService`

src/app.ts
sagzy: Yes this makes total sense! Will update

src/http/api/image-upload.ts
sagzy: Yep been a little on the fence here, as passing an Account object or the account uuid to the image storage service felt a bit weird! But agree, would help to keep the handler thin

src/app.ts
sagzy: Removed check on app boot in feb9216ea24f24bdaf8a5d5a23aabb8200fe11af and added error handling on save in 30bffd8a593f52c221eab5c9552b7e7b485392db

src/app.ts
sagzy: Ah yes - that's the beauty of it! Updated in 89364a96981c53649ca9f88f47f87aeff6d5876a

src/app.ts
sagzy: Updated in 89364a96981c53649ca9f88f47f87aeff6d5876a

src/http/api/image-upload.ts
sagzy: Updated in 5dfac2c2d9cbc690b7c74df76fed509223bcd880

src/app.ts
```diff
@@ -579,9 +589,9 @@ container.register(
 );
 
 container.register(
-    'storageHandler',
-    asFunction((accountService, storageService) =>
-        createStorageHandler(accountService, storageService),
+    'imageUploadHandler',
+    asFunction((accountService, imageStorageService) =>
```
allouis: This can just be `asFunction(createImageUploadHandler)` 😝

src/storage/adapters/gcp-storage-adapter.ts
```diff
@@ -0,0 +1,121 @@
...
+            await this.bucket.file(path).save(file.stream(), {
+                metadata: {
+                    contentType: file.type,
+                },
+                // resumable uploads (default: true) use a session and chunked uploads.
+                // Disabled it in dev/testing because resumable mode can be unreliable with GCS emulators.
+                // This is fine for small files like images in dev/testing.
+                resumable: !this.emulatorHost,
+            });
+
```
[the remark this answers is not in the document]
sagzy: Good one! Fixing

src/http/api/image-upload.ts
```diff
@@ -20,7 +17,9 @@ export function createStorageHandler(
         }
 
         const account = await accountService.getAccountForSite(ctx.get('site'));
-        const result = await storageService.saveFile(file, account.uuid);
+        const path = imageStorageService.storagePath(file, account.uuid);
```
allouis: NABD - I would probably even put this inside `save` tbh

allouis (approved):

src/storage/image-storage.service.ts
```diff
@@ -0,0 +1,48 @@
...
+    ) {}
+
+    storagePath(file: File, accountUUID: string): string {
+        // HEIC/HEIF files are converted to JPEG format
+        const isHeicFile =
+            file.type.includes('heic') || file.type.includes('heif');
+        const extension = isHeicFile ? 'jpg' : file.type.split('/')[1];
+
+        return `/images/${accountUUID}/${uuidv4()}.${extension}`;
+    }
```
mike182uk: How about making this more generic so we don't have to pass `accountUUID` or hard code the path as `/images/..`?

We could maybe have a helper in the application somewhere that generates the base path based on the account, then we pass it in to this fn, and then this fn just deals with any other stuff (i.e changing heic extension, using `uuid` as file name)?

src/test/fixtures/dog.jpg
mike182uk, line 1: Could we keep this file co-located with the test? Similar to what we do with snapshots? (maybe a `__fixtures__` dir  in `src/storage`)?

src/app.ts
sagzy: true! updated in d4c5dabae26cbc8f5cd3e9c880a1e78f4bfb1b50

src/test/fixtures/dog.jpg
sagzy: Sure thing! Wasn't sure of our code convention here, updated in 20554d69ac7407f25f3d117aa996e2b4163bc0cd

src/storage/image-storage.service.ts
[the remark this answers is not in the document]
sagzy: Fixed in 5789968b3582224d8985b16472fd3430cb7efaeb

src/storage/adapters/gcp-storage-adapter.ts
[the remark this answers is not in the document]
sagzy: Fixed in 8e88bf559bb56a95c09454fbd869485fbf992f20

src/storage/image-storage.service.ts
sagzy: Yep - also don't like passing the account UUID to the service just for the path generation.

In 7944aaa22a71a5d48d1561b49371fdd06d7ae83c, I cleaned up this HEIC/HEIF code (handled now directly in the image compressor) and used a path prefix for the image storage service instead of coupling it with the account UUID

What do you think?

src/http/api/image-upload.ts
sagzy: Updated in 7944aaa22a71a5d48d1561b49371fdd06d7ae83c

src/storage/image-storage.service.ts
mike182uk: Nice one, looks good 👍

PR #843, opened 2025-06-13
Implemented populate-explore-json job

allouis:
ref https://linear.app/ghost/issue/PROD-2011

We use `fedify` for handling fetching from the fediverse.

The basic gist of it is:
1. Get following of curator account
2. Spin up some in process workers to fetch all the actors
3. Convert them to DTOs
4. Sort and save in S3


jobs/populate-explore-json/index.ts
```diff
@@ -1,19 +1,343 @@
...
+    followerCount: number;
+    followsMe: boolean;
+    followedByMe: boolean;
+    blockedByMe: boolean;
+    domainBlockedByMe: boolean;
+}
+
+// Parse handle like @user@domain
+function parseHandle(handle: string): { username: string; domain: string } {
+    const match = handle.match(/^@([\w]+)@([\w.-]+)$/);
```
sagzy: The regex we use in AP service:
```
/@([\w.-]+)@([\w-]+\.[\w.-]+[a-zA-Z])/
```

Main differences I see:
- AP regex accepts usernames with `.` and `-` (and it should probably accept `_` too, e.g. @simba_the_yorkie@mastodon.social)
- AP regex looks for a domain with at least a `.` and a TLD ending with a letter (not sure how useful that second part is)

For simplicity, probably best to stick to the same regex?

With the exception that here we can keep the anchors `^` / `$` as we're not parsing handles inside larger strings

jobs/populate-explore-json/README.md
```diff
@@ -34,3 +37,35 @@ docker compose run --rm app
...
+      "bannerImageUrl": null,
+      "customFields": {},
+      "postCount": 100,
+      "likedCount": 0,
+      "followingCount": 50,
+      "followerCount": 200,
+      "followsMe": false,
+      "followedByMe": false,
+      "blockedByMe": false,
+      "domainBlockedByMe": false
```
sagzy:
```
      "followsMe": false,
      "followedByMe": false,
      "blockedByMe": false,
      "domainBlockedByMe": false
```

Curious whether we need these placeholder fields here, as they'll be always false. I imagine the idea is to override them in the frontend when we compare the results to the current user following - but curious if we need to return them here, or could directly append these in the frontend as we do the comparison

jobs/populate-explore-json/index.ts
allouis: Yeah good shout! I will steal that one

jobs/populate-explore-json/README.md
allouis: I copied the `AccountDTO` structure from our main app - but you're probably right, we can leave these out - will do that :)

jobs/populate-explore-json/index.ts
allouis: Actually tbh - this is just used on the input validation, I'm gonna simplify it further

PR #851, opened 2025-06-16
Handled deleted posts in reply chains

allouis:
ref https://linear.app/ghost/issue/PROD-2063

Because posts are soft-deleted in the database we need to explicitly handle them. For children & replies chains we can just act as if they don't exist - but for the ancestors, we need to return a tombstone object instead.

### DIFF
--- a/src/http/api/views/reply.chain.view.ts
+++ b/src/http/api/views/reply.chain.view.ts
@@ -1,6 +1,7 @@
 import { getAccountHandle } from 'account/utils';
 import { type Result, error, ok } from 'core/result';
 import type { Knex } from 'knex';
+import { PostType } from 'post/post.entity';
 import z from 'zod';
 import type { PostDTO } from '../types';
 
@@ -71,7 +72,7 @@ export class ReplyChainView {
         if (result.post_deleted_at !== null) {
             return {
                 id: result.post_ap_id,
-                type: 2, // PostType.Tombstone
+                type: PostType.Tombstone,
                 title: '',
                 excerpt: '',
                 summary: null,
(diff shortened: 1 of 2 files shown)


src/http/api/views/reply.chain.view.ts
```diff
@@ -66,6 +67,42 @@ export class ReplyChainView {
...
+                readingTimeMinutes: result.post_reading_time_minutes,
+                attachments: [],
+                author: {
+                    id: result.author_id.toString(),
+                    handle: getAccountHandle(
+                        result.author_url
+                            ? new URL(result.author_url).host
+                            : '',
+                        result.author_username,
+                    ),
```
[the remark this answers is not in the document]
allouis: I think we're happy exploding for invalid data here - means we're in an undefined state

src/http/api/views/reply.chain.view.ts
```diff
@@ -66,6 +67,42 @@ export class ReplyChainView {
...
+                type: 2, // PostType.Tombstone
+                title: '',
+                excerpt: '',
+                summary: null,
+                content: '',
+                url: result.post_url,
+                featureImageUrl: null,
+                publishedAt: result.post_published_at,
+                likeCount: result.post_like_count,
+                likedByMe: result.post_liked_by_current_user === 1,
```
vershwal: I think we can leave this `false`. It doesn't make sense for the UI to show if a deleted post is liked/reposted. Same for the `repostedByMe` field below.

vershwal (approved):
Looks good overall. Added one small comment.

src/http/api/views/reply.chain.view.ts
allouis: Yeah, I think you're right, thanks!

PR #864, opened 2025-06-17
Added LocalStorageAdapter

allouis:
ref https://linear.app/ghost/issue/PROD-2009

This is an alternative implementation of the StorageAdapter interface which can write files to the local file system. This removes the current dependency on GCS which makes self hosting easier.


src/storage/adapters/local-storage-adapter.ts
```diff
@@ -0,0 +1,56 @@
...
+            this.hostingUrl = new URL(`${hostingUrl.href}/`);
+        } else {
+            this.hostingUrl = hostingUrl;
+        }
+    }
+
+    async save(
+        file: File,
+        path: string,
+        // TODO this should return Result<URL>
```
allouis, line 27: This needs to happen at the interface - I can remove this comment if we want but I plan to clean it up soon

src/storage/adapters/local-storage-adapter.ts
```diff
@@ -0,0 +1,56 @@
...
+            const directory = dirname(fullPath);
+
+            await fs.mkdir(directory, { recursive: true });
+
+            await fs.writeFile(fullPath, buffer);
+
+            const url = new URL(path, this.hostingUrl).toString();
+
+            return ok(url);
+        } catch (err) {
```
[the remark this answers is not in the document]
allouis: We're conforming to an interface here, so we need to first update that and all the consumers of it - outside of scope of this PR IMO

src/storage/adapters/local-storage-adapter.ts
```diff
@@ -0,0 +1,56 @@
...
+            return error('error-saving-file');
+        }
+    }
```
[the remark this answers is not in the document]
allouis: I think this is fixed now?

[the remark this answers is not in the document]
allouis: Cool cool - will address

src/storage/adapters/local-storage-adapter.ts
sagzy: Yep good one!

src/storage/adapters/local-storage-adapter.ts
```diff
@@ -0,0 +1,61 @@
...
+
+    async save(
+        file: File,
+        path: string,
+        // TODO this should return Result<URL>
+    ): Promise<Result<string, StorageError>> {
+        if (path.startsWith('/')) {
+            return this.save(file, path.slice(1));
+        }
```
mike182uk, line 31: Is this so just so you don't have modify `path` or is there a more elaborate reason for the recursion?

src/storage/adapters/local-storage-adapter.ts
```diff
@@ -0,0 +1,61 @@
...
+
+        try {
+            const buffer = Buffer.from(await file.arrayBuffer());
+
+            const fullPath = normalize(join(this.storagePath, path));
+
+            if (!fullPath.startsWith(this.storagePath)) {
+                return error('error-saving-file');
+            }
```
mike182uk, line 40: We could do this before creating the buffer so we don't create the buffer if this fails

src/storage/adapters/local-storage-adapter.ts
```diff
@@ -0,0 +1,61 @@
...
+        if (path.startsWith('/')) {
+            return this.save(file, path.slice(1));
+        }
+
+        try {
+            const buffer = Buffer.from(await file.arrayBuffer());
+
+            const fullPath = normalize(join(this.storagePath, path));
+
+            if (!fullPath.startsWith(this.storagePath)) {
```
mike182uk, line 38: Is this still vulnerable to path traversal? (not sure if it would even be possible to somehow get a path like `../../foo/bar`)

mike182uk (approved):

src/storage/adapters/local-storage-adapter.ts
allouis: Yeah we have a lint rule about not modifying parameters, and I think this is cleaner than
```ts
let actualPath = path;
if (path.startsWith('/')) {
    actualPath = path.slice(1);
}
```

src/storage/adapters/local-storage-adapter.ts
allouis: The `normalize` takes care of this, as it returns an absolute path - so the `..` will be "expanded" - then we check the base path here

PR #866, opened 2025-06-17
Allowed LocalStorageAdapter to be configured

allouis:
ref https://linear.app/ghost/issue/PROD-2009

This configures nginx so that it can serve our static files from the shared docker volume, and mounts the docker volume so it can be used by the LocalStorageAdapter.

The DI container can conditionally register the LocalStorageAdapter based on the env variables, by default it uses the existing setup.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -7,6 +7,7 @@ services:
     volumes:
       - ./src:/opt/activitypub/src
       - ./dist:/opt/activitypub/dist
+      - dev-content:/opt/activitypub/content
     environment:
       - PORT=8080
       - MYSQL_USER=ghost
@@ -27,6 +28,8 @@ services:
       - GCP_BUCKET_NAME=activitypub
       - GCP_STORAGE_EMULATOR_HOST=http://fake-gcs:4443
       - ACTIVITYPUB_COLLECTION_PAGE_SIZE=20
+      - LOCAL_STORAGE_PATH=/opt/activitypub/content
+      # - LOCAL_STORAGE_HOSTING_URL=https://<tailscale-url>/.ghost/activitypub/local-storage
     command: yarn build:watch
     depends_on:
       migrate:
@@ -78,6 +81,8 @@ services:
     build: nginx
     ports:
       - "80:80"
+    volumes:
+      - dev-content:/content:ro
     depends_on:
       - activitypub
 
@@ -310,3 +315,4 @@ networks:
 
 volumes:
   test-content:
+  dev-content:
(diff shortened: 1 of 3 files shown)


docker-compose.yml
```diff
@@ -27,6 +28,8 @@ services:
       - GCP_BUCKET_NAME=activitypub
       - GCP_STORAGE_EMULATOR_HOST=http://fake-gcs:4443
       - ACTIVITYPUB_COLLECTION_PAGE_SIZE=20
+      - LOCAL_STORAGE_PATH=/opt/activitypub/content
+      # - LOCAL_STORAGE_HOSTING_URL=https://<tailscale-url>/.ghost/activitypub/local-storage
```
allouis, line 32: This needs to be uncommented, and updated and then local storage will be enabled

src/configuration/registrations.ts
```diff
@@ -103,12 +104,24 @@ export function registerDependencies(
...
+            if (
+                process.env.LOCAL_STORAGE_PATH &&
+                process.env.LOCAL_STORAGE_HOSTING_URL
+            ) {
+                return new LocalStorageAdapter(
+                    process.env.LOCAL_STORAGE_PATH || '',
+                    new URL(process.env.LOCAL_STORAGE_HOSTING_URL || ''),
+                );
+            }
+            if (process.env.GCP_BUCKET_NAME) {
```
allouis: Hmm - this is a slight functionality change, I think I should remove the `if` statement here

src/configuration/registrations.ts
```diff
@@ -103,6 +104,15 @@ export function registerDependencies(
...
         asFunction(() => {
+            if (
+                process.env.LOCAL_STORAGE_PATH &&
+                process.env.LOCAL_STORAGE_HOSTING_URL
+            ) {
+                return new LocalStorageAdapter(
+                    process.env.LOCAL_STORAGE_PATH || '',
+                    new URL(process.env.LOCAL_STORAGE_HOSTING_URL || ''),
+                );
+            }
```
sagzy, line 115: I wonder if we can get rid of the `|| ''`, as empty strings are not valid and we are only doing that to please Typescript :D Maybe something like:

```
const localStoragePath = process.env.LOCAL_STORAGE_PATH || ''; // string
const localStorageHostingUrl = process.env.LOCAL_STORAGE_HOSTING_URL || ''

if (localStoragePath && localStorageHostingUrl) {
    ...
}
```

?

sagzy (approved):
LGTM, 1 minor comment :)

src/configuration/registrations.ts
allouis: That's a good point - you're right!

PR #879, opened 2025-06-18
Added a helper that builds create activity for a post

vershwal:
ref https://linear.app/ghost/issue/PROD-2066/

- Refactored building create activity and Fedify object into a helper function so that it can be used for creating the outbox
- Didn't change any functionality in this PR just moved the create activity logic into a helper and added tests

### DIFF
--- a/src/activitypub/fediverse-bridge.ts
+++ b/src/activitypub/fediverse-bridge.ts
@@ -83,6 +83,7 @@ export class FediverseBridge {
             return;
         }
 
+        // TODO: Replies are currently handled in the handler file. Move that logic here.
         if (post.type === PostType.Note && post.inReplyTo) {
             return;
         }


src/activitypub/fediverse-bridge.ts
```diff
@@ -87,81 +82,15 @@ export class FediverseBridge {
...
-                ),
-                preview,
-                url: post.url,
-                to: PUBLIC_COLLECTION,
-                ccs: ccs,
-            });
-        } else {
-            throw new Error(`Unsupported post type: ${post.type}`);
+
+        if (post.type === PostType.Note && post.inReplyTo) {
```
allouis, line 87: Maybe a TODO comment above this explaining that replies are handled in the handler, but they should be moved here eventually?

allouis (approved):
This looks good! IMO the helper should live in the `src/activitypub` directory - but not a big deal

src/activitypub/fediverse-bridge.ts
vershwal: Added the comment. I am gonna leave it in the helpers dir for now.

PR #884, opened 2025-06-19
Added outbox for Ghost AP accounts

vershwal:
closes https://linear.app/ghost/issue/PROD-2096


src/dispatchers.ts
```diff
@@ -859,22 +861,75 @@ export function followingFirstCursor() {
...
+        );
+
+        const host = ctx.request.headers.get('host')!;
+        const site = await siteService.getSiteByHost(host);
+        if (!site) {
+            throw new Error(`Site not found for host: ${host}`);
+        }
+
+        const siteDefaultAccount =
+            await accountService.getDefaultAccountForSite(site);
```
vershwal: This returns the old AccountType. I want to clean this but here we are using the Fedify context so can't directly get the account from context. Second, [this method](https://github.com/TryGhost/ActivityPub/blob/main/src/account/account.repository.knex.ts#L158) `getBySite` to get AccountEntity using site is also deprecated.

src/post/post.repository.knex.ts
```diff
@@ -1091,4 +1091,80 @@ export class KnexPostRepository {
...
+        const posts: Post[] = [];
+        for (const row of rows) {
+            const post = await this.mapRowToPostEntity(row);
+            posts.push(post);
+        }
+
+        return posts;
+    }
+
+    async getOutboxItemCount(accountId: number): Promise<number> {
```
vershwal, line 1178: The above query uses the already existing index`KEY idx_outboxes_account_id_outbox_type_published_at_desc (account_id, outbox_type, published_at DESC)`
But this count query will need a new one - `(account_id, outbox_type)`

src/dispatchers.ts
allouis: Yeah fair fair - I do think `getBySite` is preferable here though - that was deprecated to try get us to use `ctx.get('account')` - but as you mention we don't have that on the Fedify context!

Not a big deal though because we're only using the `id`

src/dispatchers.ts
```diff
@@ -859,22 +861,75 @@ export function followingFirstCursor() {
...
+
+        const outbox = await postService.getOutboxForAccount(
+            siteDefaultAccount.id,
+            cursor,
+            pageSize,
+        );
+        const outboxItems = await Promise.all(
+            outbox.map(async (post: Post) => {
+                return (await buildCreateActivityAndObjectFromPost(post, ctx))
+                    .createActivity;
```
allouis: Minor, but this gave me a small wtf moment - I thought that we forgot to call `createActivity()`! 

Maybe something like
```ts
const {createActivity} = await buildCreat...
return createActivity;
```

but really up to you!

src/post/post.repository.knex.ts
allouis: I think it should be able to reuse the same index 🤔 because the first one starts with `account_id, outbox_type` it effectively works as an index over that! We should confirm this though

vershwal: You are right! It's using the index :-

```
-> Aggregate: count(0) (cost=8.07 rows=1) (actual time=0.0407..0.0409 rows=1 loops=1) 
-> Covering index lookup on outboxes using idx_outboxes_account_id_outbox_type_published_at_desc (account_id=2, outbox_type=0) (cost=4.17 rows=39) (actual time=0.0289..0.036 rows=39 loops=1)
```

Query

```
EXPLAIN ANALYZE SELECT COUNT(*) AS count
FROM outboxes
WHERE account_id = 2
AND outbox_type = 0;
```

src/post/post.repository.knex.ts
```diff
@@ -1091,4 +1091,80 @@ export class KnexPostRepository {
...
+            )
+            .join('posts', 'outboxes.post_id', 'posts.id')
+            .join('accounts', 'accounts.id', 'posts.author_id')
+            .leftJoin('users', 'users.account_id', 'accounts.id')
+            .leftJoin('sites', 'sites.id', 'users.site_id')
+            .where('outboxes.account_id', accountId)
+            .where('outboxes.outbox_type', OutboxType.Original)
+            .orderBy('outboxes.published_at', 'desc')
+            .limit(pageSize)
+            .offset(cursor ? Number.parseInt(cursor) : 0);
```
allouis: I think we'd want to use the `published_at` as the cursor here, otherwise we can potentially end up with duplicates if posts are deleted in between fetching pages. Or we could use the `outboxes.id` for pagination too

allouis (commented):
Overall looks good - I left a comment about the pagination.

The service layer isn't really doing anything - and I think it would be cool to shift the logic from the dispatcher into the service - making the dispatcher more of a wrapper - BUT I don't think converting posts -> activities belongs in the post service.

I think this is fine as-is, because we don't have a pattern for it atm - but I wonder what you think!

src/post/post.repository.knex.ts
allouis: Noice 💪 !!

src/post/post.repository.knex.ts
vershwal: Fair point!!

src/dispatchers.ts
vershwal: Updated this to use `getBySite`

src/dispatchers.ts
vershwal: Done!!

src/post/post.repository.knex.ts
vershwal: Updated this to use `published_at` for pagination

vershwal:
The service layer in this case does feel redundant. It's just there to maintain the call pattern: controller -> service -> repository.

> BUT I don't think converting posts -> activities belongs in the post service.

Yeah, I don't think so either. That was my thought while creating the helper for the post -> activity mapping, as I didn't want to include it in the post service

PR #897, opened 2025-06-19
Fixed Mentions for Ghost AP user's post

vershwal:
ref https://linear.app/ghost/issue/PROD-2066/

- `tags` can be an object when there is only one item in tag
- Handled this so that we correctly render mentions in both cases - 1. When tag is array, 2. When tag is an object

### DIFF
--- a/src/http/api/views/account.posts.view.ts
+++ b/src/http/api/views/account.posts.view.ts
@@ -437,15 +437,18 @@ export class AccountPostsView {
                         ? activity.object.tag
                         : [activity.object.tag];
 
-                    for await (const tag of tags) {
-                        if (tag.type === 'Mention') {
-                            const mention = await this.getMentionedAccount(
-                                new URL(tag.href),
-                                tag.name,
-                            );
-                            if (!isError(mention)) {
-                                mentionedAccounts.push(getValue(mention));
-                            }
+                    const mentions = tags.filter(
+                        (tag) => tag.type === 'Mention',
+                    );
+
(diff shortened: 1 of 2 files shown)


src/http/api/views/account.posts.view.ts
```diff
@@ -433,7 +433,11 @@ export class AccountPostsView {
 
                 if (activity.object.tag && activity.object.type === 'Note') {
                     const mentionedAccounts: Mention[] = [];
-                    for await (const tag of activity.object.tag) {
+                    const tags = Array.isArray(activity.object.tag)
+                        ? activity.object.tag
+                        : [activity.object.tag];
+
+                    for await (const tag of tags) {
```
sagzy: The `await` here shouldn't be necessary

sagzy: Idea: we could filter mentions before entering the loop to increase readability, e.g.:

```
  const mentions = tags.filter(tag => tag.type === 'Mention');

  for (const mention of mentions) {
      const account = await this.getMentionedAccount(
          new URL(mention.href),
          mention.name
      );

      if (!isError(account)) {
          mentionedAccounts.push(getValue(account));
      }
  }
  ```

sagzy (approved):

src/http/api/views/account.posts.view.ts
vershwal: Nice suggestion @sagzy. Did the changes!

PR #898, opened 2025-06-19
Injected PubSubEvents into the message handler

allouis:
ref https://github.com/TryGhost/ActivityPub/commit/b8de38e

During a refactor we inadvertently switched from injecting a PubSubEvents instance to an AsyncEvents instance which caused a runtime error. This adds a new dependency `pubSubEvents` which is specific to that type, and updates the handler to have it injected correctly.

### DIFF
--- a/src/events/pubsub-http.ts
+++ b/src/events/pubsub-http.ts
@@ -21,7 +21,7 @@ const IncomingMessagePayloadSchema = z.object({
 });
 
 export function createIncomingPubSubMessageHandler(
-    events: PubSubEvents,
+    pubSubEvents: PubSubEvents,
     fedify: Federation<ContextData>,
     fedifyContextFactory: FedifyContextFactory,
 ) {
@@ -68,7 +68,7 @@ export function createIncomingPubSubMessageHandler(
             await fedifyContextFactory.registerContext(
                 hostFedifyCtx,
                 async () => {
-                    await events.handleIncomingMessage(
+                    await pubSubEvents.handleIncomingMessage(
                         payload.message.data,
                         payload.message.attributes,
                     );
--- a/src/configuration/registrations.ts
+++ b/src/configuration/registrations.ts
@@ -134,6 +134,7 @@ export function registerDependencies(
 
     if (deps.globalPubSubEvents) {
         container.register('commandBus', asValue(deps.globalPubSubEvents));
+        container.register('pubSubEvents', asValue(deps.globalPubSubEvents));
     } else {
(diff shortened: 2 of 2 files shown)


src/events/pubsub-http.ts
```diff
@@ -21,7 +21,7 @@ const IncomingMessagePayloadSchema = z.object({
 });
 
 export function createIncomingPubSubMessageHandler(
-    events: PubSubEvents,
+    pubSubEvents: PubSubEvents,
```
allouis, line 24: We could have just renamed this to `commandBus` - but that's not quite right. Because the type of `commandBus` is technically just `AsyncEvents` - which doesn't confirm to the `PubSubEvents` needed here.

When the wrong type is injected, we get a runtime error which is tricky to track down. But here `pubSubEvents` is registered conditionally so we will get an error on boot if things are misconfigured.

sagzy: Yes that makes sense! Alternatively, we could register `pubSubMessageHandler` only when `deps.globalPubSubEvents` is present?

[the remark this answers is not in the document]
allouis: Good catch!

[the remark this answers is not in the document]
sagzy: +1

sagzy: We're not using `commandBus` anywhere else either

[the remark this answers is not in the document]
allouis: Fixed - come check it

sagzy: We're using `commandBus` in PostInteractionCountsService, but we expect a `PubSubEvents` there too, not `AsyncEvents`

sagzy: I don't want to block this PR - just a thought! Think we should get a fix in and your solution works 👍

sagzy (approved):

src/events/pubsub-http.ts
allouis: > Alternatively, we could register pubSubMessageHandler only when deps.globalPubSubEvents is present?

I think this means that now pubSubMessageHandler has to know how/when it's dependencies are constructed, rather than just knowing what it's dependencies are! This way - we can inline the `globalPubSubEvents` if we need, or define them differently, and `pubSubMessageHandler` continues to work as-is

> We're not using commandBus anywhere else either

It's used by the PostInteractionCountsService to dispatch the update commands - and this service only depends on the AsyncEvents interface, rather than PubSubEvents - which accounts for the difference in types between the two registrations

PR #904, opened 2025-06-20
Added reposts to outbox

vershwal:
ref https://linear.app/ghost/issue/PROD-2105/

- This adds the reposts to the outbox of Ghost's AP accounts
- Using outbox type from the outboxes table to differentiate between original posts and reposts

### DIFF
--- a/src/dispatchers.ts
+++ b/src/dispatchers.ts
@@ -25,7 +25,7 @@ import {
     buildAnnounceActivityForPost,
     buildCreateActivityAndObjectFromPost,
 } from 'helpers/activitypub/activity';
-import type { Post } from 'post/post.entity';
+import { OutboxType, type Post } from 'post/post.entity';
 import type { AccountService } from './account/account.service';
 import type { ContextData } from './app';
 import { isFollowedByDefaultSiteAccount } from './helpers/activitypub/actor';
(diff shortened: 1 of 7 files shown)


src/dispatchers.ts
```diff
@@ -890,9 +893,17 @@ export function createOutboxDispatcher(
         );
         const outboxItems = await Promise.all(
             outbox.posts.map(async (post: Post) => {
-                const { createActivity } =
-                    await buildCreateActivityAndObjectFromPost(post, ctx);
-                return createActivity;
+                if (post.author.id === siteDefaultAccount.id) {
```
allouis: This is gonna break if/when we add likes or any other non-author activity 🤔 

The `outboxes` table stores the type of activity right? I think we should be trying to use this data when reconstructing

I think the ideal state here would be to follow the same pattern as feeds - and have a view service which is interacting the with the outbox directly and mapping it to the DTO needed.

I think we can't easily do that as part of this PR - but it would be good to get it recorded somewhere...

Can you think of a way to expose the type from the outboxes table without restructuring everything?

vershwal: Added it here https://github.com/TryGhost/ActivityPub/pull/904/commits/b7e79a52694b581a7081b1bb795f9121e3693066

PR #906, opened 2025-06-20
Added fromDraft method to AccountEntity

allouis:
This will allow us to handle events when creating accounts.

### DIFF
--- a/src/account/account.entity.ts
+++ b/src/account/account.entity.ts
@@ -110,6 +110,7 @@ export class AccountEntity implements Account {
     }
 
     static fromDraft(draft: AccountDraft, id: number): AccountEntity {
+        const events: AccountEvent[] = [];
         return new AccountEntity(
             id,
             draft.uuid,
@@ -123,7 +124,7 @@ export class AccountEntity implements Account {
             draft.apFollowers,
             draft.apInbox,
             draft.isInternal,
-            [],
+            events,
         );
     }
 


src/account/account.entity.ts
```diff
@@ -109,6 +109,24 @@ export class AccountEntity implements Account {
...
+            draft.name,
+            draft.bio,
+            draft.url,
+            draft.avatarUrl,
+            draft.bannerImageUrl,
+            draft.apId,
+            draft.apFollowers,
+            draft.apInbox,
+            draft.isInternal,
+            [],
```
vershwal: While reading this, it's difficult to figure out what this [] is for. Maybe a comment here `//account events`

src/account/account.repository.knex.integration.test.ts
```diff
@@ -664,4 +664,96 @@ describe('KnexAccountRepository', () => {
...
+            }
+        }
+
+        const mockEvent = new MockAccountEvent();
+
+        fromDraftSpy.mockImplementation((...args) => {
+            const account = originalFromDraft.call(AccountEntity, ...args);
+
+            // biome-ignore lint/suspicious/noExplicitAny: We need to mock events
+            (account as any).events = [mockEvent];
```
vershwal, line 723: We can use something like this to avoid the use of `any` here:-

```
(account as unknown as { events: AccountEvent[] }).events = [mockEvent];
```

This will help us get rid of the // biome-ignore

src/account/account.repository.knex.integration.test.ts
```diff
@@ -664,4 +664,96 @@ describe('KnexAccountRepository', () => {
...
+        });
+
+        const draft = AccountEntity.draft(draftData);
+
+        await accountRepository.create(draft);
+
+        expect(emitSpy).not.toHaveBeenCalled();
+
+        emitSpy.mockRestore();
+    });
```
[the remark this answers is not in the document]
vershwal: I think coderabbitai is right here as we are checking `expect(emitSpy).not.toHaveBeenCalled();`

src/account/account.entity.ts
allouis: Great shout

src/account/account.repository.knex.integration.test.ts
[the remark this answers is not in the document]
allouis: Eh yeah - there's no initial events right now, but there will be soon.

I'll delete the test for now, and we can reintroduce it later

src/account/account.repository.knex.integration.test.ts
allouis: I did think of that too, but it feels kinda clunky

Do you think that's easier to read than the biome comment? If so I will change it

src/account/account.repository.knex.integration.test.ts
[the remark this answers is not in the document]
vershwal: Or you can just leave it like this and just change the test name. This is also a valid test

PR #907, opened 2025-06-20
Removed transactions from SiteService

allouis:
no-issue

First, we don't want to be dealing with the database at the service level, let alone transactions. Secondly - they are interacting with transactions at the repository layer in a confusing way.

Because we removed the transactions, we need to explicitly handle partial states because it's now possible for only some of the operations to have completed.

### DIFF
--- a/src/site/site.service.ts
+++ b/src/site/site.service.ts
@@ -21,15 +21,18 @@ export class SiteService {
         private ghostService: IGhostService,
     ) {}
 
-    private async createSite(host: string, conn: Knex): Promise<Site> {
-        const rows = await conn.select('*').from('sites').where({ host });
+    private async createSite(host: string): Promise<Site> {
+        const rows = await this.client
+            .select('*')
+            .from('sites')
+            .where({ host });
 
         if (rows && rows.length !== 0) {
             throw new Error(`Site already exists for ${host}`);
         }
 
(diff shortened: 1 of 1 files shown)


src/site/site.service.ts
```diff
@@ -67,17 +67,25 @@ export class SiteService {
...
-                return existingSite;
-            }
+        const existingSite = await this.getSiteByHost(host);
 
-            const newSite = await this.createSite(host, tx);
+        let site: Site;
+        if (existingSite === null) {
+            site = await this.createSite(host);
+        } else {
+            site = existingSite;
```
allouis, line 76: No early return - because now it's possible for site to have been created, but a later operation like account or user to fail.

src/site/site.service.ts
```diff
@@ -87,12 +95,11 @@ export class SiteService {
             };
 
             await this.accountService.createInternalAccount(
-                newSite,
+                site,
                 internalAccountData,
-                tx,
```
mike182uk, line 92: Have you already removed `tx` usage from `accountService.createInternalAccount`?

mike182uk (approved):

src/site/site.service.ts
allouis: This param is optional, but I will be removing it in a followup

PR #911, opened 2025-06-20
Enabled auto-merge for minor and patch versions regardless of their version number

sagzy:
ref https://linear.app/ghost/issue/PROD-2078

- our shared config only allows auto-merge on minor and patch versions if the version number is higher than or equal to 1.0.0
- this commit overrides the shared behaviour and allows auto-merging patch/minor versions even if there are on major 0

### DIFF
--- a/.github/renovate.json5
+++ b/.github/renovate.json5
@@ -2,6 +2,23 @@
   "extends": ["github>tryghost/renovate-config"],
   "packageRules": [
     {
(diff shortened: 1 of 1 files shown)


sagzy:
Closing after discussing with the team: we prefer to keep auto-merge disabled for < 1.0.0. Eventually, we'll introduce an allowlist of stable dependencies below 1.0.0, for which at least patch releases are auto-merged

PR #918, opened 2025-06-23
Added followedByMe to PostAuthor

allouis:
ref https://linear.app/ghost/issue/PROD-2097

In order to show inline follow buttons, we need to know if we're following the account.

This adds the `followedByMe` property to the PostAuthor DTO - and updates endpoints using this DTO to be conformant.

NOTE: We have not solved this for authors of posts when fetching a remote account posts, that will come in a followup PR

### DIFF
--- a/src/notification/notification.service.ts
+++ b/src/notification/notification.service.ts
@@ -156,7 +156,7 @@ export class NotificationService {
                 ).andOn('post_reposts.post_id', 'post.id');
             })
             .leftJoin('follows as follows_actor', function () {
-                this.onVal(
+                this.on(
                     'follows_actor.following_id',
                     'actor_account.id',
                 ).andOnVal(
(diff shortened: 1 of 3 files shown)


src/notification/notification.service.ts
```diff
@@ -148,6 +155,16 @@ export class NotificationService {
                     options.accountId.toString(),
                 ).andOn('post_reposts.post_id', 'post.id');
             })
+            .leftJoin('follows as follows_actor', function () {
+                this.onVal(
```
vershwal: This won't work. I tested this for notifications and `actor_followed_by_user` will always be `0` -  When using `onVal`, `'actor_account.id'` will be treated as a string. 

We should use:-

```
this.on(
                    'follows_actor.following_id',
                    'actor_account.id',
                )
```

vershwal: I also think we should add some tests that check for `"actor_followed_by_user": 1,` so that we don't miss this case.

allouis: Nice catch - thanks

PR #930, opened 2025-06-24
Updated createInternalAccount to use AccountDraft

allouis:
ref https://linear.app/ghost/issue/PROD-2077

This will cause all newly created internal accounts to emit an AccountCreatedEvent, which we can hook up to the Ghost explore service

### DIFF
--- a/src/account/account.service.ts
+++ b/src/account/account.service.ts
@@ -200,54 +200,61 @@ export class AccountService {
 
             throw new Error(`Account ${account.id} not found`);
         } catch (error) {
-            if (isDuplicateEntryError(error)) {
-                const existingAccount = await this.accountRepository.getByApId(
-                    draft.apId,
+            if (!isDuplicateEntryError(error)) {
+                throw error;
+            }
+
+            const existingAccount = await this.accountRepository.getByApId(
+                draft.apId,
+            );
+
+            if (!existingAccount) {
+                throw new Error(
+                    `Got duplicate entry error for account but account ${draft.apId} not found`,
                 );
+            }
 
-                if (!existingAccount) {
-                    throw error;
-                }
-
-                if (existingAccount.isInternal) {
-                    const returnVal = await this.getByInternalId(
-                        existingAccount.id,
-                    );
-                    if (returnVal) {
-                        return returnVal;
-                    }
-                    throw new Error(`Account ${existingAccount.id} not found`);
-                }
-
-                // If the existing account doesn't have a private key, generate one
-                // This is required for the account to sign outgoing activities after
-                // a potential migration from a different server.
-                const newKeyPair = await this.generateKeyPair();
-                await this.db('accounts')
-                    .where({ id: existingAccount.id })
-                    .update({
-                        ap_public_key: JSON.stringify(
-                            await exportJwk(newKeyPair.publicKey),
-                        ),
-                        ap_private_key: JSON.stringify(
-                            await exportJwk(newKeyPair.privateKey),
-                        ),
-                    });
-
-                await this.db('users').insert({
-                    account_id: existingAccount.id,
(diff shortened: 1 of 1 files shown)


src/account/account.service.integration.test.ts
```diff
@@ -118,7 +118,7 @@ describe('AccountService', () => {
             username: 'index',
             name: 'Test Site Title',
             bio: 'Test Site Description',
-            avatar_url: 'Test Site Icon',
+            avatar_url: 'https://www.example.com/avatars/internal-account.png',
```
allouis, line 121: This needs to be a real URL because we use the entity layer now which validates this.

src/account/account.service.integration.test.ts
```diff
@@ -173,7 +173,7 @@ describe('AccountService', () => {
                 username: username,
                 bio: internalAccountData.bio || null,
                 avatar_url: internalAccountData.avatar_url || null,
-                url: `https://${site.host}`,
+                url: `https://${site.host}/`,
```
allouis, line 176: URL's with no path have a path of `/` as according to the spec, and as implemented by the URL constructor

src/account/account.service.integration.test.ts
```diff
@@ -740,11 +740,16 @@ describe('AccountService', () => {
...
                 username: 'account1',
             });
-            await service.createInternalAccount(site, {
+            const otherSite = await fixtureManager.createSite();
+            await service.createInternalAccount(otherSite, {
                 ...internalAccountData,
                 username: 'account2',
             });
 
+            await db('users').where({ site_id: otherSite.id }).update({
```
allouis, line 749: We need to manually create the two users for a site

mike182uk: I'm not following whats happening here?  

- we are creating a new site (`otherSite`) (which will create a user)
- creating another internal account for `otherSite`
- then updating the `site_id` of both created users to `site.id`

why do we need `otherSite`?

src/account/account.service.ts
```diff
@@ -173,116 +173,82 @@ export class AccountService {
...
-                                ),
-                                ap_private_key: JSON.stringify(
-                                    await exportJwk(newKeyPair.privateKey),
-                                ),
-                            });
-
-                        existingAccount.ap_public_key = JSON.stringify(
+                // If the existing account doesn't have a private key, generate one
+                // This is required for the account to sign outgoing activities after
+                // a potential migration from a different server.
```
mike182uk: This out of date? We now always generate the key pair

src/account/account.service.ts
```diff
@@ -173,116 +173,82 @@ export class AccountService {
...
+                return returnVal;
+            }
 
-                await tx('users').insert({
-                    account_id: accountId,
-                    site_id: site.id,
-                });
+            throw new Error(`Account ${account.id} not found`);
+        } catch (error) {
+            if (isDuplicateEntryError(error)) {
```
mike182uk: We could maybe swap this around to reduce the levels of indentation:

```
if (!isDuplicateEntryError(error)) {
    throw error
}

const existingAccount...
```

src/account/account.service.integration.test.ts
allouis: We're testing what happens when we have two internal users for a site in the db. It's not possible to get into that state using our normal methods, so we create two internal users for different sites, and then move one into the first site

src/account/account.service.ts
allouis: We only generate the keypair if the account is not internal - which means it doesn't have a private key, i'll update the comment tho

src/account/account.service.ts
allouis: Yeah nice

PR #931, opened 2025-06-24
Added GhostExploreService

allouis:
This is the beginning of our GhostExploreService which will handle surfacing accounts to be shared!

The first step of that is having them follow the Ghost Explore account so we can interact with them easier

### DIFF
--- a/src/explore/ghost-explore.service.integration.test.ts
+++ b/src/explore/ghost-explore.service.integration.test.ts
@@ -2,7 +2,9 @@ import type { Logger } from '@logtape/logtape';
 import { KnexAccountRepository } from 'account/account.repository.knex';
 import type { AccountService } from 'account/account.service';
 import { AccountCreatedEvent } from 'account/events';
+import type { FedifyContextFactory } from 'activitypub/fedify-context.factory';
 import { AsyncEvents } from 'core/events';
+import { error, ok } from 'core/result';
 import type { Knex } from 'knex';
 import { createTestDb } from 'test/db';
 import { type FixtureManager, createFixtureManager } from 'test/fixtures';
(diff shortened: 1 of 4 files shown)


src/explore/ghost-explore.service.integration.test.ts
```diff
@@ -0,0 +1,264 @@
...
+    beforeEach(async () => {
+        await fixtureManager.reset();
+
+        // Clean up the database
+        await db.raw('SET FOREIGN_KEY_CHECKS = 0');
+        await db('follows').truncate();
+        await db('accounts').truncate();
+        await db('users').truncate();
+        await db('sites').truncate();
+        await db.raw('SET FOREIGN_KEY_CHECKS = 1');
```
mike182uk: `fixtureManager.reset` should be doing this already?

src/explore/ghost-explore.service.integration.test.ts
```diff
@@ -0,0 +1,264 @@
...
+
+        // Insert a site
+        const siteData = {
+            host: 'www.example.com',
+            webhook_secret: 'secret',
+        };
+        const [id] = await db('sites').insert(siteData);
+        site = { id, ...siteData };
```
mike182uk: `fixtureManager.createSite`?

src/explore/ghost-explore.service.integration.test.ts
```diff
@@ -0,0 +1,264 @@
...
+                ap_liked_url:
+                    'https://mastodon.social/users/ghostexplore/liked',
+                ap_shared_inbox_url: 'https://mastodon.social/inbox',
+                url: 'https://mastodon.social/users/ghostexplore',
+                domain: 'mastodon.social',
+            });
+
+            // Create internal account
+            const [internalAccount] =
+                await fixtureManager.createInternalAccount(site);
```
mike182uk: We could move this to a helper maybe? Looks to be repeated in a bunch of tests

src/explore/ghost-explore.service.ts
```diff
@@ -0,0 +1,64 @@
...
+                'Not following Ghost Explore account for non-internal account {apId}',
+                {
+                    apId: account?.apId.href,
+                },
+            );
+            return;
+        }
+
+        const ghostExplore = await this.accountRepository.getByApId(
+            new URL('https://mastodon.social/users/ghostexplore'),
```
mike182uk, line 63: hm, i was going suggesting moving this out to a constant, or have it be passed into the service on construction, but i guess the likelihood of this ever changing is slim right?

src/explore/ghost-explore.service.integration.test.ts
allouis: I've moved to a helper

src/explore/ghost-explore.service.integration.test.ts
allouis: Yeah you're right, can just let createInternlalAccount handle it

src/explore/ghost-explore.service.integration.test.ts
allouis: yup!

mike182uk (approved):

PR #933, opened 2025-06-24
Added alttext for images

vershwal:


vershwal:
@coderabbitai review

src/http/api/note.ts
```diff
@@ -7,6 +7,12 @@ import { postToDTO } from './helpers/post';
 
 const NoteSchema = z.object({
     content: z.string(),
+    image: z
+        .object({
+            url: z.string().url().optional(),
+            altText: z.string().optional(),
+        })
+        .optional(),
```
vershwal, line 16: I want all image related data in one object so I added this new `image` object. But I am still keeping the `imageUrl` field to keep this API backward compatible till the UI changes are done. Once the UI changes are in, I'll clean this up in a follow PR.

src/post/post.entity.ts
```diff
@@ -412,6 +412,7 @@ export class Post extends BaseEntity {
         noteContent: string,
         imageUrl?: URL,
         mentions: Mention[] = [],
+        altText?: string,
```
allouis: I think your pattern of having `{url: URL, name: string}` fits really nice here, rather than adding an extra parameter

src/post/post.service.ts
```diff
@@ -288,6 +288,7 @@ export class PostService {
         account: Account,
         content: string,
         image?: URL,
+        altText?: string,
```
allouis: Same comment as above, I think putting the image url and alt text into an object is a good pattern, and we should use it here too

allouis (changes requested):
I think just a small cleanup to params, otherwise looks great!

src/http/api/note.ts
```diff
@@ -35,17 +36,21 @@ export async function handleCreateNote(
 
     if (data.imageUrl) {
         imageUrl = new URL(data.imageUrl);
-    } else if (data.image?.url) {
+    } else if (data.image) {
         imageUrl = new URL(data.image.url);
     }
 
-    const altText = data.image?.altText ?? undefined;
+    const image: ImageAttachment | undefined = imageUrl
```
vershwal, line 43: This will be cleaned up after UI changes, we don't have to construct this here we can directly pass the image object received in the req params.

src/http/api/note.ts
```diff
@@ -2,14 +2,15 @@ import { z } from 'zod';
...
+import type { ImageAttachment } from 'post/post.entity';
 import type { PostService } from 'post/post.service';
 import { postToDTO } from './helpers/post';
 
 const NoteSchema = z.object({
     content: z.string(),
     image: z
         .object({
-            url: z.string().url().optional(),
+            url: z.string().url(),
```
allouis, line 13: Nice catch

PR #946, opened 2025-06-26
Added job to backfill follows for Ghost Explore

allouis:


.claude/settings.local.json
```diff
@@ -0,0 +1,20 @@
+{
```
mike182uk: I think this needs to be named `settings.json` if its intended to share with the rest of the team (`settings.local.json` is just _your_ project specific settings). I'd also remove some of the allowed commands for writing (`chmod`, `mkdir`) as this should be developer preference. We probs don't need the job backfill commands either

docker-compose.yml
```diff
@@ -64,6 +64,22 @@ services:
...
+    build: jobs/backfill-ghost-explore-follows
+    environment:
+      - MYSQL_USER=ghost
+      - MYSQL_PASSWORD=password
+      - MYSQL_HOST=mysql
+      - MYSQL_PORT=3306
+      - MYSQL_DATABASE=activitypub
+      - BATCH_SIZE=10
+      - MAX_CONCURRENT=5
+      - GHOST_EXPLORE_AP_ID=https://mastodon.social/users/testingshtuff
```
mike182uk, line 77: Leftover testing?

.claude/settings.local.json
allouis: ooops that's my mistake, you're right

docker-compose.yml
allouis: Kinda, I didn't want to use the real account when running locally - i think this should stay

PR #947, opened 2025-06-26
Added handler for Update(Person) activity

vershwal:
closes https://linear.app/ghost/issue/PROD-2173/

### DIFF
--- a/src/account/account.service.ts
+++ b/src/account/account.service.ts
@@ -633,6 +633,41 @@ export class AccountService {
         await this.accountRepository.save(updated);
     }
 
+    async updateAccountByApId(
+        apId: URL,
+        data: {
+            name: string | null;
+            bio: string | null;
+            username: string;
+            avatarUrl: string | null;
+            bannerImageUrl: string | null;
+            url: string | null;
+        },
+    ): Promise<Result<true, 'account-not-found'>> {
+        const profileData = {
+            name: data.name,
+            bio: data.bio,
+            username: data.username,
+            avatarUrl: data.avatarUrl ? new URL(data.avatarUrl) : null,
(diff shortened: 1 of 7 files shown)


allouis (changes requested):
We want to avoid adding specific methods like this to the repository, because it becomes easy to leak domain logic into the repo. Instead we want to stick with the `save` method!

We have the `updateProfile` method on `Account`, and that returns an updated account which can be saved with the repository, though that might need updating to handle custom_fields and url

We also have an `updateAccountProfile` on the `AccountService` which we can probably use, though we might want to add any extra one like `updateAccountProfileByApId` because we don't have access to `ctx.get('account')` in activity handlers.

The updateAccountProfile method takes the updated values, which leaves the job of pulling them out to the controller, which is usually a http handler, but in this case is the activity handler.

So we end up with
- UpdateActivityHandler responsible for parsing and extracting data from the real world, and then passing a strict/known object to the service layer
- AccountService responsible for orchestrating the use-case, coordination the Entity & Repository
- AccountEntity responsible for setting the new values, validating them, creating events, etc...
- AccountRepository responsible for persisting the entity to the db and dispatching the events

src/account/account.service.ts
```diff
@@ -633,6 +633,41 @@ export class AccountService {
...
+        apId: URL,
+        data: {
+            name: string | null;
+            bio: string | null;
+            username: string;
+            avatarUrl: string | null;
+            bannerImageUrl: string | null;
+            url: string | null;
+        },
+    ): Promise<Result<true, 'account-not-found'>> {
```
allouis, line 646: I think generally we return the updated entity from the service, but not a big deal

allouis (approved):

PR #948, opened 2025-06-27
Added /feed/notes and /feed/reader APIs

allouis:
ref https://linear.app/ghost/issue/PROD-2186
ref https://linear.app/ghost/issue/PROD-2184

We want to rename these concepts in the UI and I'd like to update the API now while we can, so we keep everything in sync.

### DIFF
--- a/features/step_definitions/feed_steps.js
+++ b/features/step_definitions/feed_steps.js
@@ -1,23 +1,8 @@
 import assert from 'node:assert';
 
-import { Then, When } from '@cucumber/cucumber';
+import { Then } from '@cucumber/cucumber';
 
 import { waitForAPObjectInFeed } from '../support/feed.js';
-import { fetchActivityPub } from '../support/request.js';
-
-When('we request the feed with the next cursor', async function () {
-    const responseJson = await this.response.clone().json();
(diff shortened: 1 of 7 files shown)


features/step_definitions/feed_steps.js
```diff
@@ -1,23 +1,8 @@
 import assert from 'node:assert';
 
-import { Then, When } from '@cucumber/cucumber';
+import { Then } from '@cucumber/cucumber';
 
 import { waitForAPObjectInFeed } from '../support/feed.js';
-import { fetchActivityPub } from '../support/request.js';
-
-When('we request the feed with the next cursor', async function () {
```
allouis, line 8: This wasn't used anywhere, and refered to a non-existent endpoint

src/app.ts
```diff
@@ -1121,6 +1121,22 @@ app.get(
         return handlerFactory('Inbox')(ctx);
     }),
 );
+app.get(
```
allouis, line 1124: We will delete the old API as part of a cleanup issue already tracked in linear

vershwal (approved):
lgtm!

PR #952, opened 2025-06-27
Added custom fields in AccountEntity

vershwal:
ref https://linear.app/ghost/issue/PROD-2173/

- Added custom fields in AccountEntity and related interfaces/objects/constructors
- Added custom fields while reading and writing to accounts table
- Updating custom fields when an Update activity is received


src/account/account.service.ts
```diff
@@ -570,7 +571,7 @@ export class AccountService {
             avatar_url: row.avatar_url,
             banner_image_url: row.banner_image_url,
             url: row.url,
-            custom_fields: JSON.parse(row.custom_fields),
```
allouis, line 573: Why did this change? Was it broken before?

vershwal:
> we've dropped JSON.parse in a few places, is that a bug that we never saw before because we don't save these or?

Yes, our integration tests found this issue once I added the customeFields. The db already returns [Object] so we don't need JSON.parse here. Added more tests to verify the same here: https://github.com/TryGhost/ActivityPub/pull/952/commits/8e8a359f590b240f9d4faf2c9bdc609583fe1f0f

PR #957, opened 2025-06-30
Prevented unecessary like & repost count updates

allouis:
ref https://linear.app/ghost/issue/PROD-2098

We think that the unecessary updates to these columns is causing
contention on the rows, and resulting in slow queries. This solution is
a little bit dirty, pun intended, but a fast way to validate the problem.


sagzy (approved):
Nice, this should solve the issue!

In terms of naming, wonder if we can do something like:
- `hasLikeCountChanged` (instead of `isLikeCountDirty`)
- and have `_newLikeCount` / `_newRepostCount` (instead of `_likeCountDirty` / `_repostCountDirty`)

I would then find the logic more readable in post repository > save() 

Just an idea, feel free to discard!

PR #980, opened 2025-07-03
Added attachment data to notifications API endpoint

minimaluminium:
ref PROD-1971

- Extended notifications API response to include image attachments
- Modified backend to retrieve and include attachment URLs and metadata
- Prepared data structure for frontend to display images in notifications

### DIFF
--- a/src/notification/__snapshots__/get-notifications-data.json
+++ b/src/notification/__snapshots__/get-notifications-data.json
@@ -74,7 +74,7 @@
         "name": "example.jpg",
         "type": "image",
         "url": "https://example.com/image.jpg",
-      }
+      },
     ],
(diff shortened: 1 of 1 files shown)


src/notification/notification.service.ts
```diff
@@ -47,6 +47,12 @@ interface BaseGetNotificationsDataResultRow {
     post_reply_count: number;
     post_repost_count: number;
     post_reposted_by_user: 0 | 1;
+    post_attachments: {
+        type: string | null;
+        mediaType: string | null;
+        name: string | null;
+        url: string;
+    }[];
```
allouis: I think we need an `| null` here as the attachments column can be null 🤔

allouis (approved):

PR #984, opened 2025-07-03
Updated fedify to 1.7.2

allouis:
This fixes a bug when handling incoming RFC 9421 signatures, rather than
responding with a 500, we 401 instead.


allouis:
https://github.com/fedify-dev/fedify/compare/1.7.0...1.7.2

mike182uk (approved):

PR #994, opened 2025-07-07
Added Fedify delivery error analysis to prevent redudant retries

mike182uk:
ref https://linear.app/ghost/issue/PROD-2210

Added error analysis when processing a message from the queue. If the error is a Fedify delivery error (classified based on the format of the message), we analyze the status code in the message and decide on a retry / report strategy based on this status code. In the future we can also analyze other types of error to decide on an applicable retry / report strategy for the given error

### DIFF
--- a/src/mq/gcloud-pubsub-push/error-utils.ts
+++ b/src/mq/gcloud-pubsub-push/error-utils.ts
@@ -13,7 +13,7 @@ export interface ErrorAnalysis {
  * Analyze an error to determine its characteristics and handling strategy
  *
  * Fedify delivery errors have the format:
- * "Failed to send activity <activityId> to <inbox> (<status> <statusText>):\n<error body>"
+ * "Failed to send activity <activity-id> to <inbox-url> (<status-code> <status-text>):\n<error body>"
  *
  * Non-Fedify delivery errors are considered application errors and are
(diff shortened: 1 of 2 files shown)


src/mq/gcloud-pubsub-push/error-utils.ts
```diff
@@ -0,0 +1,57 @@
...
+ * "Failed to send activity <activityId> to <inbox> (<status> <statusText>):\n<error body>"
+ *
+ * Non-Fedify delivery errors are considered application errors and are
+ * retryable and reportable
+ *
+ * @param error The error to analyze
+ */
+export function analyzeError(error: Error): ErrorAnalysis {
+    // Try to match Fedify delivery error format
+    const fedifyMatch = error.message.match(/\((\d{3})\s+[\w\s]+\):/);
```
allouis: Do we want to be a little more specific here? even something like `error.message.startsWith('Failed to send activity') && ...`?

mike182uk: Yeh nice one, I've made the regex more specific and added a couple of test cases to verify → https://github.com/TryGhost/ActivityPub/pull/994/commits/e1b03a164a79a805440d50996b39de6eca8f5284

PR #999, opened 2025-07-08
Added agentic AI documentation

mike182uk:
no ref

Added documentation to help AI agents contribute to the codebase


dev/ai/agent-docs
```diff
@@ -0,0 +1,13 @@
+#!/usr/bin/env bash
```
allouis: Can we just commit the symlinks, instead of having a script to create them?

mike182uk: Symlinks are committed, this was just to make it easier to add / remove others if needed 😅

mike182uk: Discussed on Slack - We don't need the script for now, have removed in https://github.com/TryGhost/ActivityPub/pull/999/commits/dcdf97e519ef834ad0c4bf5479730e3425b7c69a

PR #1004, opened 2025-07-08
Moved post deletion logic to post service

vershwal:
no issue

- We want to move the deletion logic from controller to service so that it can be reused in handlers
- No functional change

### DIFF
--- a/src/post/post.service.integration.test.ts
+++ b/src/post/post.service.integration.test.ts
@@ -910,4 +910,142 @@ describe('PostService', () => {
             expect(count).toBe(0);
         });
     });
+
+    describe('deleteByApId', () => {
+        it('should delete a post successfully', async () => {
+            const post = await fixtureManager.createPost(account);
+
+            const result = await postService.deleteByApId(post.apId, account);
+
+            if (isError(result)) {
+                throw new Error('Result should not be an error');
+            }
+
+            const success = getValue(result);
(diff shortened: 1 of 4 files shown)


src/post/post.service.integration.test.ts
```diff
@@ -910,4 +910,142 @@ describe('PostService', () => {
...
+
+    describe('deleteByApId', () => {
+        it('should delete a post successfully', async () => {
+            const post = await fixtureManager.createPost(account);
+
+            const result = await postService.deleteByApId(post.apId, account);
+
+            if (isError(result)) {
+                throw new Error('Result should not be an error');
+            }
```
sagzy: Could also be written as

```
expect(isError(result)).toBe(false);
```

src/post/post.service.ts
```diff
@@ -499,4 +501,21 @@ export class PostService {
...
+    async deleteByApId(
+        apId: URL,
+        account: Account,
+    ): Promise<Result<boolean, DeletePostError>> {
+        const postResult = await this.getByApId(apId);
+        if (isError(postResult)) {
+            return postResult;
+        }
+        const post = getValue(postResult);
+        if (post.author.id !== account.id) {
```
sagzy: Works either way, but just noting that we've changed this author check from a `uuid` check to `id` check

sagzy (approved):

PR #1011, opened 2025-07-09
Added logic to update a post

vershwal:
ref https://linear.app/ghost/issue/PROD-2287/

- These changes add the post update logic to entity and repository
- `PostUpdateParams` are the fields that can be updated
- A post update even is emitted when a post is updated

### DIFF
--- a/src/post/post.entity.ts
+++ b/src/post/post.entity.ts
@@ -64,6 +64,16 @@ export interface GhostPost {
     authors?: GhostAuthor[] | null;
 }
 
+export interface PostUpdateParams {
+    title: PostTitle | null;
+    content: string | null;
+    excerpt: PostSummary | null;
+    summary: PostSummary | null;
+    imageUrl: URL | null;
+    url: URL;
+    metadata: Metadata | null;
+}
+
 export interface PostAttachment {
     type: string | null;
     mediaType: string | null;
@@ -122,7 +132,7 @@ export type CreatePostError = 'private-content' | 'missing-content';
 export class Post extends BaseEntity {
     public readonly uuid: string;
     public readonly apId: URL;
-    public readonly url: URL;
+    private _url: URL;
     private likesToRemove: Set<number> = new Set();
     private likesToAdd: Set<number> = new Set();
     private repostsToAdd: Set<number> = new Set();
@@ -130,7 +140,13 @@ export class Post extends BaseEntity {
     private deleted = false;
     private _likeCountDirty = false;
     private _repostCountDirty = false;
-    public readonly content: string | null;
+    private _updateDirty = false;
+    private _content: string | null;
+    private _title: PostTitle | null;
+    private _excerpt: PostSummary | null;
+    private _summary: PostSummary | null;
+    private _imageUrl: URL | null;
+    private _metadata: Metadata | null;
     public readonly mentions: MentionedAccount[] = [];
 
     constructor(
@@ -139,14 +155,14 @@ export class Post extends BaseEntity {
         public readonly author: Account,
         public readonly type: CreatePostType,
         public readonly audience: Audience,
-        public readonly title: PostTitle | null,
-        public readonly excerpt: PostSummary | null,
-        public readonly summary: PostSummary | null,
+        title: PostTitle | null,
+        excerpt: PostSummary | null,
+        summary: PostSummary | null,
         content: string | null,
         url: URL | null,
-        public readonly imageUrl: URL | null,
+        imageUrl: URL | null,
         public readonly publishedAt: Date,
-        public readonly metadata: Metadata | null = null,
+        metadata: Metadata | null = null,
         private _likeCount = 0,
         private _repostCount = 0,
         public readonly replyCount = 0,
@@ -173,15 +189,48 @@ export class Post extends BaseEntity {
             this.apId = apId;
         }
         if (url === null) {
-            this.url = this.apId;
+            this._url = this.apId;
         } else {
-            this.url = url;
+            this._url = url;
         }
-        this.content = content !== null ? sanitizeHtml(content) : null;
+        this._title = title;
+        this._excerpt = excerpt;
+        this._summary = summary;
+        this._content = content !== null ? sanitizeHtml(content) : null;
+        this._imageUrl = imageUrl;
+        this._metadata = metadata;
         if (_deleted) {
             this.deleted = true;
             this.handleDeleted();
         }
+    }
+
+    get title(): PostTitle | null {
+        return this._title;
+    }
+
+    get excerpt(): PostSummary | null {
+        return this._excerpt;
+    }
+
+    get summary(): PostSummary | null {
+        return this._summary;
+    }
+
+    get content(): string | null {
+        return this._content;
+    }
+
+    get imageUrl(): URL | null {
+        return this._imageUrl;
+    }
+
+    get url(): URL {
+        return this._url;
+    }
+
+    get metadata(): Metadata | null {
+        return this._metadata;
     }
 
     get isInternal() {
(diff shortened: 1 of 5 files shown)


src/http/api/webhook.controller.ts
```diff
@@ -93,4 +93,46 @@ export class WebhookController {
...
+     *
+     * @param ctx App context instance
+     */
+    async handlePostUpdated(ctx: AppContext) {
+        let data: PostInput;
+        console.log('handlePostUpdated', await ctx.req.json());
+
+        try {
+            data = PostPublishedWebhookSchema.parse(
+                (await ctx.req.json()) as unknown,
```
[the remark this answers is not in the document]
vershwal: Removed the log

src/post/post.repository.knex.ts
```diff
@@ -150,6 +150,12 @@ export class KnexPostRepository {
         });
     }
 
+    async getByUuid(uuid: Post['uuid']): Promise<Post | null> {
+        return await this.getByQuery((qb: Knex.QueryBuilder) => {
+            return qb.where('posts.uuid', uuid);
+        });
+    }
+
```
vershwal: This query will change when we'll have the new table that maps ghost-uuid to our AP posts

src/post/post.repository.knex.ts
```diff
@@ -438,6 +464,10 @@ export class KnexPostRepository {
                 );
             }
 
+            if (wasUpdated) {
+                //TODO: Send post updated event
```
vershwal: Will add the event and sending update activity to fediverse in the next PR

src/app.ts
```diff
@@ -881,6 +881,16 @@ app.post(
     }),
 );
 
+app.post(
+    '/.ghost/activitypub/webhooks/post/updated',
```
allouis: I don't think we need this one

src/post/post.repository.knex.ts
```diff
@@ -150,6 +150,12 @@ export class KnexPostRepository {
         });
     }
 
+    async getByUuid(uuid: Post['uuid']): Promise<Post | null> {
```
allouis: I don't think we need this, we can just get by ap id instead? Once we lose the link between ghost post uuid and ap post uuid, this becomes redundant anyway

src/post/post.service.ts
```diff
@@ -541,4 +544,56 @@ export class PostService {
...
+        const postResult = await Post.createArticleFromGhostPost(
+            account,
+            ghostPost,
+        );
+        if (isError(postResult)) {
+            const error = getError(postResult);
+            switch (error) {
+                case 'missing-content':
+                case 'private-content':
+                    //Remove the post if it's private or empty
```
allouis: This logic belongs in the entity not the service

src/post/post.service.ts
```diff
@@ -541,4 +544,56 @@ export class PostService {
...
+
+    async updateGhostPostByUuid(
+        account: Account,
+        ghostPost: GhostPost,
+    ): Promise<Result<boolean, UpdatePostError>> {
+        const post = await this.postRepository.getByUuid(ghostPost.uuid);
+        if (post === null) {
+            return error('post-not-found');
+        }
+        const postResult = await Post.createArticleFromGhostPost(
```
allouis: I don't think we want to be using `createArticleFromGhostPost` like this 🤔 I would expect that we would have a method like `updateFromGhostPost` on the entity, which would handle all of the logic that is below like deleting etc...

src/post/post.repository.knex.ts
vershwal: Ghost webhook req doesn't have the ap_id, do you mean we generate ap_id using uuid and then fetch using it?

src/post/post.entity.ts
```diff
@@ -378,6 +411,116 @@ export class Post extends BaseEntity {
         );
     }
 
+    async updateArticleFromGhostPost(
```
allouis: Can we deduplicate the logic in here? it's almost identitical to the method above, I think we can probably share most/all of it

src/post/post.entity.ts
```diff
@@ -378,6 +411,116 @@ export class Post extends BaseEntity {
...
+                summary: ghostPost.custom_excerpt,
+                imageUrl: parseURL(ghostPost.feature_image),
+                url: new URL(ghostPost.url),
+                metadata: metadata,
+            };
+
+            this.update(account, params);
+        }
+
+        return new Post(
```
allouis: I'm not sure we want to be returning a new post here? That is the pattern we use in Account, and it is the pattern we want to move to for Post - but I think we need to do this all at once, rather than partially

src/post/post.repository.knex.ts
allouis: Yeah exactly, we know that the `uuid` on the ghost post isn't going to match the ap post forever, once we've implemented the unpublish/republish flow! So i'm thinking we're going to have some kind of mapping between a ghost post and an activitypub post - which is either gonna be the `id` or the `apId`

Right now we're able to generate the `apId` from the uuid, so it seems like a good starting point!

vershwal: For now I have used this method to generate the apId from the uuid :-
```
const apId = account.getApIdForPost({
            uuid: ghostPost.uuid,
            type: PostType.Article,
        });
``` 
`getApIdForPost` is deprecated but this will be removed once we have the `ghost_uuid` <-> `ap_id` mapping

src/post/post.entity.ts
vershwal: Great! I wasn't sure of this either. We don't even need the updatedPost in the webhook controller. Removed it!

src/post/post.entity.ts
vershwal: I have used the `createArticleFromGhostPost` method to create the updated post. I could have also had the common logic separated out, but I think this approach is cleaner.

src/post/post.entity.ts
```diff
@@ -199,6 +218,23 @@ export class Post extends BaseEntity {
...
     }
 
+    update(account: Account, params: PostUpdateParams) {
+        if (account.uuid !== this.author.uuid) {
+            throw new Error(
+                `Account ${account.uuid} cannot update Post ${this.uuid}`,
+            );
+        }
+        this.updated = true;
+        this.updatedParams = params;
```
allouis: We don't want to be storing the updated params as a separate object here, they should be on the entity itself.

```ts
const post: Post;

post.update(account, {title: 'new title'});

post.title // this should be 'new title';
```

This would remove the need for a `getUpdatedParams` method too

allouis: We can store a dirty flag and do the update if it's dirty, i think we do similar in account

vershwal: It's not great but I added that because the fields we want to update are readonly. Can't directly update these fields so I have used any to bypass the readonly constraint, this was also done in [handleDelete()](https://github.com/TryGhost/ActivityPub/pull/1011/commits/9737a7b672d9f2354a47271627efc7026172b22e#diff-3522562e572113f90d24c5500be2ae5bf5647a64867dfb920f6727a32a18facfL238) - don't really like it though

I have kept the interface `PostUpdateParams` makes passing these updated params between functions easier. Also keeps a check on which fields are allowed to update

Used update dirty flag to keep track of updates

src/http/api/webhook.controller.ts
```diff
@@ -93,4 +93,42 @@ export class WebhookController {
...
+
+        const account = ctx.get('account');
+
+        const result = await this.postService.updateGhostPost(account, data);
+
+        if (isError(result)) {
+            const error = getError(result);
+            switch (error) {
+                case 'post-not-found':
+                    return NotFound('Could not find post to update');
```
allouis: Do we want to create the post in this case?

src/post/post.entity.ts
```diff
@@ -199,6 +217,26 @@ export class Post extends BaseEntity {
...
+        if (account.uuid !== this.author.uuid) {
+            throw new Error(
+                `Account ${account.uuid} cannot update Post ${this.uuid}`,
+            );
+        }
+        this._updateDirty = true;
+
+        // TODO: Clean up the any type
+        // biome-ignore lint/suspicious/noExplicitAny: Legacy code needs proper typing
+        const self = this as any;
```
allouis: We shouldn't be introducing new usage of `any` here - this breaks type checking, if `PostUpdateParams` is out of sync with `Post` then we won't catch it!

allouis: I just realised we actually are out of sync! title, excerpt and summary are no longer `string` types - they use `PostTitle` and `PostSummary`

allouis: So I know this is annoying, as you need to switch these properties from being `public readonly` to having a private mutable state and then a `readonly` getter. This is part of the reason why the `Acccount` entity was refactored to be immutable, because it simplifies a lot of this.

Here is where the change happened https://github.com/TryGhost/ActivityPub/commit/2465f741cd52340431768bdf748aafef4159277d and you can see the `updateProfile` method before and after.

Before, we had underscore prefixed private properties, which we could modify in the method, and then the readonly getters to expose them on the entity.

I think you'll need to do one of these solutions here - it might be worth fixing the `Post` entity first in a separate PR and then building this on top, depending how long you think it'll take you. Alternatively we add the state&getters for these 7 properties

src/http/api/webhook.controller.ts
vershwal: Good point. Added here: https://github.com/TryGhost/ActivityPub/pull/1011/commits/7301d1306f9494d0012d15652799228ab1804be8

src/post/post.entity.ts
vershwal: It will def be worth fixing the Post entity but, for now I have added the private properties and getters here: https://github.com/TryGhost/ActivityPub/pull/1011/commits/d9bf1c707b5c331ea9ece727fef0917d38837228

PR #1014, opened 2025-07-09
Added /v1 routes

allouis:
ref https://linear.app/ghost/issue/PROD-2224

We want to make sure that we're able to make breaking changes to the API endpoints and adding the version to the path helps with that.

Note that AP endpoints are unversioned, as they adhere to the AP spec.

### DIFF
--- a/src/app.ts
+++ b/src/app.ts
@@ -53,7 +53,7 @@ import { LikeController } from 'http/api/like.controller';
 import { MediaController } from 'http/api/media.controller';
 import { NotificationController } from 'http/api/notification.controller';
 import { PostController } from 'http/api/post.controller';
-import { ReplyChainController } from 'http/api/reply-chain';
+import { ReplyChainController } from 'http/api/reply-chain.controller';
 import { SearchController } from 'http/api/search.controller';
 import type { SiteController } from 'http/api/site.controller';
 import { WebFingerController } from 'http/api/webfinger.controller';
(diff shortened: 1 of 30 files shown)


src/app.ts
```diff
@@ -881,6 +891,16 @@ app.post(
     }),
 );
 
+app.post(
```
allouis, line 894: This is manually duplicated because webhooks aren't migrated to the `@Route` decorator yet, as they have special middleware

src/app.ts
```diff
@@ -781,6 +781,16 @@ app.get(
     }),
 );
 
+app.get(
+    '/.ghost/activitypub/v1/site',
```
allouis, line 785: This is manually duplicated as this endpoint doesn't use the `@Route` decorator because it needs to come earlier in the middleware stack

src/http/api/feed.controller.ts
```diff
@@ -29,15 +29,15 @@ export class FeedController {
         private readonly postInteractionCountsService: PostInteractionCountsService,
     ) {}
 
-    @Route('GET', '/.ghost/activitypub/feed')
```
allouis, line 32: This is the old route which we've migrated away from

src/http/api/feed.controller.ts
```diff
@@ -29,15 +29,15 @@ export class FeedController {
...
     @Route('GET', '/.ghost/activitypub/feed/notes')
+    @Route('GET', '/.ghost/activitypub/v1/feed/notes')
     @RequireRoles(GhostRole.Owner, GhostRole.Administrator)
     async getNotesFeed(ctx: AppContext) {
         return this.handleGetFeed(ctx, 'Feed');
     }
 
-    @Route('GET', '/.ghost/activitypub/inbox')
```
allouis, line 39: This is the old route which we've migrated away from

sagzy (approved):
This all looks good! Definitely see the benefits on the earlier code refactoring in this PR :D

PR #1020, opened 2025-07-10
Added `updated` to fedifyObject

vershwal:
ref https://linear.app/ghost/issue/PROD-2217/

- Update activity needs the `updated` field. Mastodon Docs [here](https://docs-p.joinmastodon.org/spec/activitypub/#supported-activities-for-statuses)
- Added it as a pre requisite for sending Update activity. 
- Added `updated` to the `getFedifyObjectForPost` method which will be shared between Create and Update activity.

### DIFF
--- a/src/helpers/activitypub/__snapshots__/article-create-activity.json
+++ b/src/helpers/activitypub/__snapshots__/article-create-activity.json
@@ -42,7 +42,7 @@
     "published": "2025-01-12T10:30:00Z",
     "to": "as:Public",
     "type": "Article",
-    "updated": "2025-01-12T10:30:00Z",
+    "updated": "2025-01-15T10:30:00Z",
     "url": "https://example.com/post/post-123",
   },
   "to": "as:Public",
--- a/src/helpers/activitypub/activity.ts
+++ b/src/helpers/activitypub/activity.ts
@@ -22,7 +22,7 @@ async function getFedifyObjectForPost(
     let ccs: URL[] = [];
     const updatedAt = post.updatedAt
         ? Temporal.Instant.from(post.updatedAt.toISOString())
-        : Temporal.Now.instant();
+        : Temporal.Instant.from(post.publishedAt.toISOString());
 
     if (post.type === PostType.Note) {
         mentions = post.mentions.map(
--- a/src/helpers/activitypub/activity.unit.test.ts
+++ b/src/helpers/activitypub/activity.unit.test.ts
@@ -15,25 +15,6 @@ import {
     buildCreateActivityAndObjectFromPost,
 } from './activity';
 
-vi.mock('@js-temporal/polyfill', async () => {
-    const original = await vi.importActual<
-        typeof import('@js-temporal/polyfill')
-    >('@js-temporal/polyfill');
-    return {
-        Temporal: {
-            ...original.Temporal,
-            Now: {
-                // Return a fixed instant for deterministic testing
-                instant: vi
-                    .fn()
-                    .mockReturnValue(
-                        original.Temporal.Instant.from('2025-01-15T10:30:00Z'),
-                    ),
-            },
-        },
-    };
-});
-
 describe('Build activity', () => {
     let context: FedifyContext;
     let mockUriBuilder: UriBuilder<FedifyObject>;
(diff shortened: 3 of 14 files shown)


src/helpers/activitypub/__snapshots__/article-create-activity.json
```diff
@@ -42,6 +42,7 @@
     "published": "2025-01-12T10:30:00Z",
     "to": "as:Public",
     "type": "Article",
+    "updated": "2025-01-01T00:00:00Z",
     "url": "https://example.com/post/post-123",
```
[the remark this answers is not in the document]
vershwal: Done in df04b7cd1d53717807d3d6a4a001ed6319861c0d

src/helpers/activitypub/activity.ts
```diff
@@ -40,6 +43,7 @@ async function getFedifyObjectForPost(
             content: post.content,
             summary: post.summary,
             published: Temporal.Instant.from(post.publishedAt.toISOString()),
+            updated: updatedAt,
```
vershwal, line 46: This is the main change

mike182uk: Should we add a test for the branching behaviour? i.e it uses `post.updatedAt` if provided, otherwise defaults to new?

mike182uk: Hm, would it be better to fallback to `post.publishedAt` instead of `now` when `updatedAt` is falsy?

src/helpers/activitypub/activity.unit.test.ts
```diff
@@ -15,6 +15,25 @@ import {
...
+                // Return a fixed instant for deterministic testing
+                instant: vi
+                    .fn()
+                    .mockReturnValue(
+                        original.Temporal.Instant.from('2025-01-15T10:30:00Z'),
+                    ),
+            },
+        },
+    };
+});
```
mike182uk: I wonder if we could extract this to a helper? I think we have it in a few places now. Something like:

```ts
export const createTemporalMock = (value: string) => {
  return async () => {
    const original = await vi.importActual<typeof import('@js-temporal/polyfill')>('@js-temporal/polyfill');
    return {
      Temporal: {
        ...original.Temporal,
        Now: {
          instant: vi.fn().mockReturnValue(
            original.Temporal.Instant.from(value)
          ),
        },
      },
    };
  };
};  
```

Then use like:

```ts
import { createTemporalMock } from '...blah'

vi.mock('@js-temporal/polyfill', createTemporalMock('2025-01-15T10:30:00Z'));
```

What do you think?

src/helpers/activitypub/activity.ts
vershwal: > fallback to post.publishedAt instead of now when updatedAt is falsy

Yes I think this is better, Thanks. Fixed here: https://github.com/TryGhost/ActivityPub/pull/1020/commits/e4db39d2e034ade3ca5b89d2d1b2987f64a3f156

> Should we add a test for the branching behaviour? i.e it uses post.updatedAt if provided, otherwise defaults to new?

I think we already have this tested in multiple tests and it's no worth adding the test separately. 
1. Case 1: When `post.updatedAt` if provided : [here](https://github.com/TryGhost/ActivityPub/pull/1020/files#diff-e055bf088481f986eabe35f124bfc94c5bdf1c307cc958d27af28b0976064c11R66)
2. Case 2: Many tests where updatedAt is not provided it falls back to publishedAt. Example [here](https://github.com/TryGhost/ActivityPub/pull/1020/files#diff-62e7c43113e4ac6470cecbe199629fec14106286348f029e2200621371831a79R33)

src/helpers/activitypub/activity.unit.test.ts
vershwal: Great idea. But now that we have added fallback to `post.publishedAt` we don't need this mock 😊

src/helpers/activitypub/activity.ts
mike182uk: ok cool, as long as this gets caught if we change the behaviour 👍

src/helpers/activitypub/activity.unit.test.ts
mike182uk: Even better 🎉

mike182uk (approved):

PR #1023, opened 2025-07-10
Fixed workflow for running public image deploys

allouis:
This was not being run when tags are pushed, which means the v1.x images were not being built and pushed.

### DIFF
--- a/.github/workflows/cicd.yml
+++ b/.github/workflows/cicd.yml
@@ -12,6 +12,8 @@ on:
   push:
     branches:
       - main
+    tags:
+      - 'v*.*.*'  # Trigger on semantic version tags like v1.0.0
 
 permissions:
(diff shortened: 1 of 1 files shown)


.github/workflows/cicd.yml
```diff
@@ -157,7 +159,7 @@ jobs:
         run: yarn test
 
       - name: "Authenticate with GCP"
-        if: github.ref == 'refs/heads/main' || (github.event_name == 'pull_request' && (github.event.action == 'opened' || github.event.action == 'synchronize' || github.event.action == 'reopened' || github.event.action == 'labeled' || github.event.action == 'unlabeled'))
```
allouis, line 160: Do we need to update this step? i'm not clear if it's required for the login public registry later

PR #1036, opened 2025-07-15
Added Ghost Post Service

vershwal:
ref https://linear.app/ghost/issue/PROD-2290/

- Ghost post service handles updates related to ghost posts
- Wired up handlePostUpdated method to update ghost posts
- updateByApId in post service can update a post based on apId and provided params
- Added tests


src/ghost-post/ghost-post.service.ts
```diff
@@ -0,0 +1,98 @@
...
+import type { PostService } from 'post/post.service';
+
+export class GhostPostService {
+    constructor(
+        private readonly postService: PostService,
+        private readonly logger: Logger,
+    ) {}
+
+    async updateArticleFromGhostPost(account: Account, ghostPost: GhostPost) {
+        const apId = account.getApIdForPost({
```
vershwal: This will change when we'll have the mapping table

PR #1040, opened 2025-07-15
Added logic to delete ghost post

vershwal:
ref https://linear.app/ghost/issue/PROD-2273/

- handlePostDeleted handles post.deleted event
- This deletes a post from user's feed when a post is deleted in ghost admin


src/http/api/webhook.controller.ts
```diff
@@ -128,6 +136,22 @@ export class WebhookController {
...
+        } catch (err) {
+            if (err instanceof Error) {
+                return BadRequest(`Could not parse payload: ${err.message}`);
+            }
+            return BadRequest('Could not parse payload');
+        }
+
+        const account = ctx.get('account');
+
+        await this.ghostPostService.deleteGhostPost(account, uuid);
```
mike182uk: Do we need to check if this op was ok and return appropriate res or are we ok always returning 200?

src/ghost/ghost-post.service.ts
```diff
@@ -95,4 +95,19 @@ export class GhostPostService {
             }
         }
     }
+
+    async deleteGhostPost(account: Account, uuid: string) {
+        const apId = account.getApIdForPost({
```
mike182uk, line 109: `getApIdForPost` is[ marked as deprecated](https://github.com/TryGhost/ActivityPub/blob/main/src/account/account.entity.ts#L40-L43), should we be using an alternative?

vershwal: You are right but this will change when we'll have have the `ghost-uuid` <> `ap_id` mapping in a separate table. Those changes will come in this week itself so temporarily using this.

src/http/api/webhook.controller.ts
vershwal: I initially thought that since these are webhook URLs and no one is consuming the responses, returning 200 would be just fine. But now when I think of it, it makes sense to add appropriate responses after all. This way, we can have the responses in our logs, which could be useful for monitoring, metrics etc. Added here: https://github.com/TryGhost/ActivityPub/pull/1040/commits/0ceb53712156ace2d8376c21b4d51745d2cb7296

I'll also add these error responses for the other method(`updateArticleFromGhostPost`) separately for consistency.

src/ghost/ghost-post.service.ts
```diff
@@ -95,4 +101,23 @@ export class GhostPostService {
...
+            type: PostType.Article,
+        });
+
+        const deleteResult = await this.postService.deleteByApId(apId, account);
+        if (isError(deleteResult)) {
+            this.logger.error(
+                'Failed to delete post with apId: {apId}, error: {error}',
+                { apId, error: getError(deleteResult) },
+            );
+        }
```
mike182uk: now you are returning the result back to the controller, should the logging happen there instead of here?

vershwal: Fair point. Done ☑️

PR #1041, opened 2025-07-15
Refactored `app.ts` global dep registrations

mike182uk:
no ref

Refactored `app.ts` global dependencies registrations so that all dependencies are registered via the container

- Removed queue initialisation try/catch as this is now redundant - The constructor does not perform any logic
- Removed conditional dep initialisation / registration  - The queue, pubsub client, pubsub event classes are always registered, and usage is determined by `process.env.USE_MQ`
- Made `KnexKvStore.create` not `async` as it doesn't do any `async` operations - This enables us to construct via the container
- Removed intermediate `const`'s when resolving services to `init` them in `app.ts` - This prevents wtf moments when you name the variable the same as something in the container and the bundling blows up
- Removed conditional around push message handler route - This will only be called if the infrastructure makes a request to it

### DIFF
--- a/src/app.ts
+++ b/src/app.ts
@@ -8,6 +8,7 @@ import {
     type Context,
     Create,
     Delete,
+    type Federation,
     Follow,
     type KvStore,
     Like,
@@ -16,7 +17,6 @@ import {
     type RequestContext,
     Undo,
     Update,
-    createFederation,
 } from '@fedify/fedify';
 import { federation } from '@fedify/fedify/x/hono';
 import { serve } from '@hono/node-server';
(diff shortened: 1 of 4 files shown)


src/app.ts
```diff
@@ -272,28 +201,24 @@ if (process.env.MANUALLY_START_QUEUE === 'true') {
 }
 
 // Initialize services that need it
-const fediverseBridge = container.resolve<FediverseBridge>('fediverseBridge');
-fediverseBridge.init();
+container.resolve<FediverseBridge>('fediverseBridge').init();
```
allouis, line 204: Would be great to get these `init`s handled by the container/registration for us eventually

allouis (approved):

PR #1043, opened 2025-07-15
Added delivery failure tracking

mike182uk:
ref https://linear.app/ghost/issue/PROD-2266

When a delivery to an account fails, the system should track this using a backoff system to prevent sending activities to accounts that are permanently broken

In the case of a temporary failure, when the system successfully sends an activity to an account, any previously recorded backoffs are removed

### DIFF
--- a/src/account/account.service.ts
+++ b/src/account/account.service.ts
@@ -51,7 +51,7 @@ type RemoteAccountFetchError =
     | 'network-failure'
     | 'not-found';
 
-export const DELIVERY_FAILURE_BACKOFF_SECONDS = 60;
+export const DELIVERY_FAILURE_BACKOFF_SECONDS = 604800; // 1 week - This is temporarily high to do an initial block of broken accounts - This will be reduced to 60 seconds in the future;
 export const DELIVERY_FAILURE_BACKOFF_MULTIPLIER = 2;
 
 export class AccountService {


src/account/account.service.ts
```diff
@@ -51,6 +51,9 @@ type RemoteAccountFetchError =
     | 'network-failure'
     | 'not-found';
 
+const DELIVERY_FAILURE_BACKOFF_SECONDS = 60;
+const DELIVERY_FAILURE_BACKOFF_MULTIPLIER = 2;
```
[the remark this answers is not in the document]
mike182uk: This is actually by design - When a delivery failure occurs we can't be sure its because the receiver is having a temporary problem or not. This backoff mechanism gives the receiver a fair chance of receiving future activities if they are having temporary issues that later get resolved. The greater the period of time that delivery attempts fail, the more confident we are that this account really is broken and not just experiencing a temporary issue

src/account/account.service.ts
```diff
@@ -748,4 +751,51 @@ export class AccountService {
...
+            );
+
+            await this.db('account_delivery_backoffs').insert({
+                account_id: accountId,
+                last_failure_reason: failureReason,
+                backoff_until: backoffUntil,
+                backoff_seconds: DELIVERY_FAILURE_BACKOFF_SECONDS,
+            });
+        }
+    }
```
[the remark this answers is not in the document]
mike182uk: See https://github.com/TryGhost/ActivityPub/pull/1043#discussion_r2207423976

src/mq/gcloud-pubsub-push/mq.ts
```diff
@@ -260,6 +266,64 @@ export class GCloudPubSubPushMessageQueue implements MessageQueue {
...
+            return;
+        }
+
+        try {
+            const inboxUrl = new URL(message.inbox);
+
+            const account =
+                await this.accountService.getAccountByInboxUrl(inboxUrl);
+
+            if (account) {
```
allouis: What do you think about:
```ts
if (!account) {
  throw new Error('Account not found even though we just sent to it???');
}

await this.accountService.recordDeliverFailure();
```

src/mq/gcloud-pubsub-push/mq.ts
```diff
@@ -260,6 +266,64 @@ export class GCloudPubSubPushMessageQueue implements MessageQueue {
...
```
allouis: Same as above

src/account/account.service.ts
```diff
@@ -748,4 +751,51 @@ export class AccountService {
 
         await this.accountRepository.save(updated);
     }
+
+    async getAccountByInboxUrl(inboxUrl: URL): Promise<Account | null> {
+        return await this.accountRepository.getByInboxUrl(inboxUrl);
+    }
+
+    async recordDeliveryFailure(
+        accountId: number,
```
allouis: If this took an `inboxUrl` we wouldn't have to expose a `getAccountByInboxUrl` method

mike182uk: Ah nice, this would also mean we don't need to resolve the account in `mq.handleMessage` also

PR #1054, opened 2025-07-16
Write/Delete from mapping table on post publish and deletion

vershwal:
ref https://linear.app/ghost/issue/PROD-2294 https://linear.app/ghost/issue/PROD-2297

- Added createGhostPost logic to GhostPostService
- While creating the post we check if the post already exists. Create a new post if it doesn't exist and add it to the mapping table
- Deletes the ghost post mapping when a PostDeletedEvent is received

### DIFF
--- a/src/ghost/ghost-post.service.integration.test.ts
+++ b/src/ghost/ghost-post.service.integration.test.ts
@@ -82,7 +82,13 @@ describe('GhostPostService', () => {
             logger,
         );
 
-        ghostPostService = new GhostPostService(db, postService, logger);
+        ghostPostService = new GhostPostService(
+            db,
+            postService,
+            logger,
+            events,
+        );
+        await ghostPostService.init();
 
         // Reset the database before each test
         await fixtureManager.reset();
@@ -522,6 +528,52 @@ describe('GhostPostService', () => {
 
             expect(deleteByApIdSpy).toHaveBeenCalledWith(apId, account);
             expect(isError(deleteResult)).toBe(true);
+        });
+
+        it('should remove ghost post mapping when post is deleted', async () => {
+            const uuid = 'ee218320-b2e6-11ef-8a80-0242ac120011';
+
+            const createResult = await ghostPostService.createGhostPost(
+                account,
+                {
+                    title: 'Test Article for Mapping Deletion',
+                    uuid,
+                    html: '<p>Content that will be deleted</p>',
+                    excerpt: 'Test excerpt',
+                    custom_excerpt: null,
+                    feature_image: null,
+                    published_at: new Date().toISOString(),
+                    url: 'https://example.com/test-article-mapping-deletion',
+                    visibility: 'public',
+                    authors: [],
+                },
+            );
+
+            expect(isError(createResult)).toBe(false);
+            if (isError(createResult)) {
+                throw new Error('Failed to create initial post');
+            }
+
+            const mappingBeforeDeletion = await db('ghost_ap_post_mappings')
+                .select('*')
+                .where('ghost_uuid', uuid)
+                .first();
+
+            expect(mappingBeforeDeletion).not.toBeNull();
+            expect(mappingBeforeDeletion.ghost_uuid).toBe(uuid);
+
+            const deleteResult = await ghostPostService.deleteGhostPost(
+                account,
+                uuid,
+            );
+            expect(isError(deleteResult)).toBe(false);
+
+            const mappingAfterDeletion = await db('ghost_ap_post_mappings')
+                .select('*')
+                .where('ghost_uuid', uuid)
+                .first();
+
+            expect(mappingAfterDeletion).toBeUndefined();
         });
     });
 });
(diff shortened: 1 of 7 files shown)


src/post/post.service.ts
```diff
@@ -452,12 +452,6 @@ export class PostService {
 
         const post = getValue(postResult);
 
-        const existingPost = await this.postRepository.getByApId(post.apId);
-
-        if (existingPost) {
-            return error('post-already-exists');
```
vershwal, line 458: This check is moved to `ghost-post.service.ts` [here](https://github.com/TryGhost/ActivityPub/pull/1054/files#diff-cdf8e932faf153e8dd07b3ab83ed447c2c99d26ef904946a92340f50e16fcd85R141)

vershwal:
@coderabbitai full review

vershwal:
@coderabbitai full review

src/ghost/ghost-post.service.ts
```diff
@@ -113,4 +129,57 @@ export class GhostPostService {
 
         return await this.postService.deleteByApId(apId, account);
     }
+
+    async createGhostPost(
```
mike182uk, line 138: It looks like there is going to be multiple inserts happening (1 for the post, 1 for the mapping), we should probably use a transaction for these

src/ghost/ghost-post.service.ts
```diff
@@ -113,4 +129,57 @@ export class GhostPostService {
...
+            return postResult;
+        }
+
+        const post = getValue(postResult);
+        await this.insertGhostPostMapping(data.uuid, post.apId);
+
+        return ok(post);
+    }
+
+    private async insertGhostPostMapping(ghostUuid: string, apId: URL) {
```
mike182uk: It looks like this and `getApIdForGhostPost` is only used once, it maybe better to just inline these in `createGhostPost` (especially if using a transaction)

src/ghost/ghost-post.service.ts
```diff
@@ -113,4 +129,57 @@ export class GhostPostService {
...
+        account: Account,
+        data: GhostPost,
+    ): Promise<Result<Post, GhostPostError>> {
+        const existingPost = await this.getApIdForGhostPost(data.uuid);
+
+        if (existingPost) {
+            return error('post-already-exists');
+        }
+
+        const postResult = await this.postService.handleIncomingGhostPost(
```
mike182uk: Seeming though this is the Ghost post service, shouldn't the Ghost specific logic be here instead of in `handleIncomingGhostPost`? Do we even need `postService.handleIncomingGhostPost` anymore if we could handle everything Ghost specific here?

src/ghost/ghost-post.service.integration.test.ts
```diff
@@ -380,6 +386,90 @@ describe('GhostPostService', () => {
         });
     });
 
+    describe('createGhostPost', () => {
+        it('should create a new ghost post successfully', async () => {
```
mike182uk, line 388: Hm, I was thinking that maybe a unit test would be better for this and then a separate integration test for the mapping insertion - Really `createGhostPost` is just a wrapper around the post service create post method, and that should already have tests, so in a unit test we could just test that this dep gets called with the correct params (meaning we don't need to duplicate the testing of the db insertions etc). What do you think?

src/ghost/ghost-post.service.ts
vershwal: `getApIdForGhostPost` will be used for getting the post for update and delete. Made `insertGhostPostMapping` inline

src/ghost/ghost-post.service.ts
vershwal: You are right, `handleIncomingGhostPost` does feel redundant now. Moved that logic to `GhostPostService`. Removed `handleIncomingGhostPost`

src/ghost/ghost-post.service.integration.test.ts
vershwal: Right, these were duplicate earlier. But now after removing `handleIncomingGhostPost` and [related tests](https://github.com/TryGhost/ActivityPub/pull/1054/files#diff-c2506487a6d9669587ed71a21c410bb537537cef8d60aed36ac915b0a567fe63L144) these are not duplicate so I am gonna keep them :)

src/ghost/ghost-post.service.unit.test.ts
```diff
@@ -0,0 +1,170 @@
...
+            expect(mockQueryBuilder.insert).toHaveBeenCalledWith({
+                ghost_uuid: ghostPostData.uuid,
+                ap_id: mockPost.apId.href,
+            });
+
+            expect(mockLogger.error).not.toHaveBeenCalled();
+
+            expect(mockPostService.deleteByApId).not.toHaveBeenCalled();
+        });
+    });
```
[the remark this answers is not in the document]
vershwal: This is already tested here https://github.com/TryGhost/ActivityPub/pull/1054/files#diff-fc10f45372b9607aef286bf1a68aba8e84b42fd76905297d3ff4ce7329ff484dR438

src/ghost/ghost-post.service.ts
vershwal: Added error handling for ghost post mapping insert failures. We'll clean up the post if mapping insertion fails. Changes here: https://github.com/TryGhost/ActivityPub/pull/1054/commits/bcb40cdf704bd51c14b2c7d1e9693d67e1e9ef82

PR #1060, opened 2025-07-16
Added additional error analysis

mike182uk:
no ref

Added additional error analysis for network connectivity related errors as well as adding additional checks for SSL errors

### DIFF
--- a/src/mq/gcloud-pubsub-push/error-utils.ts
+++ b/src/mq/gcloud-pubsub-push/error-utils.ts
@@ -118,9 +118,9 @@ function isNetworkConnectivityError(error: Error, depth = 0): boolean {
 }
 
 function analyzeNetworkConnectivityError(error: Error): ErrorAnalysis {
-    // Network connectivity errors are not retryable and not reportable
+    // Network connectivity errors are retryable but not reportable
     return {
-        isRetryable: false,
+        isRetryable: true,
         isReportable: false,
     };
 }
--- a/src/mq/gcloud-pubsub-push/error-utils.ts
+++ b/src/mq/gcloud-pubsub-push/error-utils.ts
@@ -61,6 +61,14 @@ function isUpstreamSSLError(error: Error, depth = 0): boolean {
         return true;
     }
 
+    if (error.message.match(/certificate has expired/i) !== null) {
+        return true;
+    }
+
+    if (error.message.match(/self-signed certificate/i) !== null) {
+        return true;
+    }
+
     if ('cause' in error && error.cause instanceof Error) {
         return isUpstreamSSLError(error.cause, depth + 1);
     }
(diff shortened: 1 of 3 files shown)


tsconfig.json
```diff
@@ -12,7 +12,7 @@
     // "disableReferencedProjectLoad": true,             /* Reduce the number of projects loaded automatically by TypeScript. */
 
     /* Language and Environment */
-    "target": "es2018" /* Set the JavaScript language version for emitted JavaScript and include compatible library declarations. */,
+    "target": "es2023" /* Set the JavaScript language version for emitted JavaScript and include compatible library declarations. */,
```
mike182uk, line 15: I've bumped this to `es2023` as this should be fully supported in node22, which we are using. This was necessary to be able to use the `AggregateError` type in `isNetworkConnectivityError`

src/mq/gcloud-pubsub-push/error-utils.ts
```diff
@@ -76,6 +84,47 @@ function analyzeUpstreamSSLError(error: Error): ErrorAnalysis {
...
+        return false;
+    }
+
+    if (
+        error.message.match(/connect EHOSTUNREACH/i) !== null ||
+        error.message.match(/connect ETIMEDOUT/i) !== null ||
+        error.message.match(/connect ECONNREFUSED/i) !== null ||
+        error.message.match(/connect ECONNRESET/i) !== null ||
+        error.message.match(/socket hang up/i) !== null
+    ) {
```
vershwal, line 98: There can be other types of network error as well, but I think we have added he ones we see frequently on sentry. Is that correct?

src/mq/gcloud-pubsub-push/error-utils.ts
```diff
@@ -76,6 +84,47 @@ function analyzeUpstreamSSLError(error: Error): ErrorAnalysis {
...
+        return isNetworkConnectivityError(error.cause, depth + 1);
+    }
+
+    return false;
+}
+
+function analyzeNetworkConnectivityError(error: Error): ErrorAnalysis {
+    // Network connectivity errors are not retryable and not reportable
+    return {
+        isRetryable: false,
```
vershwal: Why are these not retryable?
For ex ETIMEDOUT - Could be due to temporary slowness, socket hang up - could be transient?

src/mq/gcloud-pubsub-push/error-utils.ts
mike182uk: Yep thats correct, these are the most common ones in the recent Sentry error reports. We will add more as and when needed 👍

vershwal: Nice 👌

src/mq/gcloud-pubsub-push/error-utils.ts
mike182uk: Hm, yeh i wasn't sure about this. Maybe they should be retryable but not reportable (as we can't do anything about it) 🤔

I guess my concern was if we keep getting `ETIMEDOUT` or whatever, we are still going to keep trying to send to them indefinitely

If we class the error as non retryable, the account will get a delivery backoff assigned to it, which means we temporarily stop sending future activities to it, which feels more like what we want (and solves the transient issue), but does start to feel like we are implementing our own retry mechanism 😅

vershwal (approved):

PR #1063, opened 2025-07-16
Added logic for unpublishing ghost post

vershwal:
ref https://linear.app/ghost/issue/PROD-2298 https://linear.app/ghost/issue/PROD-2311

- When a post is unpublished from Ghost admin, we want to remove it from the feeds
- Post will be deleted and when it is published again, we create a brand new post
- Using a random uuid while creating a post in AP for Ghost Article to avoid duplicate uuid while publishing the post again after deletion


vershwal:
@coderabbitai full review

vershwal:
@coderabbitai full review

src/http/api/webhook.controller.ts
```diff
@@ -141,12 +145,18 @@ export class WebhookController {
         });
     }
 
-    async handlePostDeleted(ctx: AppContext) {
+    async handlePostDeleted(ctx: AppContext, reason: DeleteReason) {
```
vershwal: I added this `reason` because for these two request only the input is different and this way we can avoid duplicating the whole controller logic

src/post/post.entity.unit.test.ts
```diff
@@ -549,7 +549,6 @@ describe('Post', () => {
         );
         const post = getValue(postResult as Ok<Post>) as Post;
 
-        expect(post.uuid).toEqual(ghostPost.uuid);
```
mike182uk, line 552: 🤔 should we maybe check that the `post.uuid` _doesn't_ equal `ghostPost.uuid`?

src/http/api/webhook.controller.ts
mike182uk: what about a common fn?

```ts
async handlePostDeleted(ctx: AppContext) {
	return this.deletePost(ctx, PostDeletedWebhookSchema)
}

async handlePostUnpublished(ctx: AppContext) {
	return this.deletePost(ctx, PostUnpublishedWebhookSchema)
}

async private deletePost(ctx: AppContext, schema: blah) {
	//...
}
```

src/post/post.entity.unit.test.ts
vershwal: Fair fair. Done 👍

src/http/api/webhook.controller.ts
vershwal: Nice one! Did this. 
Slight change - I have left the input parsing logic in the controller method and added a `deletePost` method that deletes post for a given account and uuid.

PR #1072, opened 2025-07-17
Restricted Create activity sending to external accounts

allouis:
The Create activity handler is essentially a noop when handling
activities that we already know about. This means that sending
Create activities to internal accounts is superfluous.

### DIFF
--- a/src/account/account.service.ts
+++ b/src/account/account.service.ts
@@ -752,6 +752,17 @@ export class AccountService {
         await this.accountRepository.save(updated);
     }
 
+    // TODO Move all methods below to a delivery service
+
+    async shouldDeliverActivity(inboxUrl: URL): Promise<boolean> {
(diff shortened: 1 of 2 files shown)


sagzy (approved):

src/account/account.service.ts
```diff
@@ -752,6 +752,17 @@ export class AccountService {
         await this.accountRepository.save(updated);
     }
 
+    // TODO Move all methods below to a delivery service
+
+    async shouldDeliverActivity(inboxUrl: URL): Promise<boolean> {
+        const account = await this.accountRepository.getByInboxUrl(inboxUrl);
+        if (!account) {
+            return false;
```
vershwal, line 760: If account is not found, it could be an external account that is not in our db. So we `shouldDeliverActivity` to it? 🤔

vershwal: ooh, there is a condition added below to search only for internal accounts. All good 👍

PR #1073, opened 2025-07-17
Added reads from mapping table

vershwal:
ref https://linear.app/ghost/issue/PROD-2296/

- Added logic to read from the new `ghost_ap_post_mappings` table for updating and deleting posts
- The `ghost_ap_post_mappings` table has been backfilled with data on prod. Commit here: https://github.com/TryGhost/ActivityPub/commit/ba12b87504138343800c79550c697a14d4c97576
- Cleaned up some tests

### DIFF
--- a/src/ghost/ghost-post.service.integration.test.ts
+++ b/src/ghost/ghost-post.service.integration.test.ts
@@ -7,7 +7,7 @@ import { AsyncEvents } from 'core/events';
 import { getError, getValue, isError } from 'core/result';
 import type { Knex } from 'knex';
 import { ModerationService } from 'moderation/moderation.service';
-import { Post, PostType } from 'post/post.entity';
+import { Post } from 'post/post.entity';
 import { KnexPostRepository } from 'post/post.repository.knex';
 import { PostService } from 'post/post.service';
 import type { ImageStorageService } from 'storage/image-storage.service';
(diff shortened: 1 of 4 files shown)


src/ghost/ghost-post.service.integration.test.ts
```diff
@@ -160,42 +160,6 @@ describe('GhostPostService', () => {
             );
         });
 
-        it('should create a new post when the post does not exist', async () => {
```
vershwal, line 163: Tested [here](https://github.com/TryGhost/ActivityPub/pull/1073/files#diff-5c81e24007c2cfda5c5b7b00fc25c87538a30c855661cbe153797a7b8cc6c4a2R181) in unit tests. And we already have integration tests for creating new post so this feels redundant.

src/ghost/ghost-post.service.integration.test.ts
```diff
@@ -507,27 +473,6 @@ describe('GhostPostService', () => {
             expect(Post.isDeleted(deletedPost!)).toBe(true);
         });
 
-        it('should return an error when deletion fails', async () => {
```
vershwal, line 510: Same for this. This test doesn't have much to be an integration test. Makes more sense to be tested [here](https://github.com/TryGhost/ActivityPub/pull/1073/files#diff-5c81e24007c2cfda5c5b7b00fc25c87538a30c855661cbe153797a7b8cc6c4a2R279)

src/ghost/ghost-post.service.ts
```diff
@@ -44,10 +45,25 @@ export class GhostPostService {
     }
 
     async updateArticleFromGhostPost(account: Account, ghostPost: GhostPost) {
```
vershwal, line 47: This method is a little mess and it's on my radar to clean this up but I want to keep this PR focused on reading from the new table. Will clean this in a separate PR

src/ghost/ghost-post.service.integration.test.ts
sagzy: Yes, as long as we do have one integration test that tests that creating a new post works - then I think it's ok to test the failure cases in unit tests

src/ghost/ghost-post.service.ts
```diff
@@ -44,10 +45,25 @@ export class GhostPostService {
...
+                account,
+                ghostPost,
+            );
+            if (isError(newPostResult)) {
+                this.logger.error(
+                    'Failed to create new post with uuid: {uuid}, error: {error}',
+                    { uuid: ghostPost.uuid, error: getError(newPostResult) },
+                );
+            }
+            return;
```
mike182uk, line 64: Are we ok to fail silently here?

vershwal: No. You are right, I have this on my list to clean up this `updateArticleFromGhostPost` method. This method doesn't return the errors back to the controller and we always return a 200 for updates. I'll clean this up in a separate PR.

PR #1075, opened 2025-07-17
Set Ghost (Pro) flag when creating a new site

sagzy:
ref https://linear.app/ghost/issue/PROD-2284

- context: as we open up the ActivityPub infra to self-hosters, we want to be able to monitor usage
- logic: when a site is created, we check for the request IP address. If the IP address matches one of the Ghost (Pro) IP address, we set the `"ghost_pro"` flag in the `"sites"` table to true

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -30,6 +30,7 @@ services:
       - ACTIVITYPUB_COLLECTION_PAGE_SIZE=20
       - LOCAL_STORAGE_PATH=/opt/activitypub/content
       # - LOCAL_STORAGE_HOSTING_URL=https://<tailscale-url>/.ghost/activitypub/local-storage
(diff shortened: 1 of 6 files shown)


docker-compose.yml
```diff
@@ -30,6 +30,7 @@ services:
       - ACTIVITYPUB_COLLECTION_PAGE_SIZE=20
       - LOCAL_STORAGE_PATH=/opt/activitypub/content
       # - LOCAL_STORAGE_HOSTING_URL=https://<tailscale-url>/.ghost/activitypub/local-storage
+      # - GHOST_PRO_IP_ADDRESSES=100.83.192.90,192.168.65.1
```
sagzy, line 33: The `GHOST_PRO_IP_ADDRESSES ` env variable has been set on staging and prod

allouis (approved):

PR #1088, opened 2025-07-22
Ran multitenant activitypub service for cucumber

allouis:

In order to be able to test the changes made for disabling activity sending to internal accounts, we need to have a test environment which support running multiple sites on the same service.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -222,15 +222,13 @@ services:
       - GCS_LOCAL_STORAGE_HOSTING_URL=https://self.test/.ghost/activitypub/gcs
       - TAGS
       - URL_GHOST_ACTIVITY_PUB=https://self.test
-      - URL_EXTERNAL_ACTIVITY_PUB=http://fake-external-activitypub.test
+      - URL_EXTERNAL_ACTIVITY_PUB=https://fake-external-activitypub.test
     command: /opt/activitypub/node_modules/.bin/cucumber-js
     depends_on:
       caddy-testing:
         condition: service_started
-      fake-gcs:
+      activitypub-testing:
         condition: service_healthy
-      fake-external-activitypub:
-        condition: service_started
 
   mysql-testing:
     networks:
@@ -278,6 +276,7 @@ services:
           - alice.test
           - bob.test
           - carol.test
+          - fake-external-activitypub.test
     image: caddy:2
     ports:
       - "443:443"
@@ -292,6 +291,8 @@ services:
         condition: service_started
       ghost-wiremock:
         condition: service_started
+      external-wiremock:
+        condition: service_started
 
   ghost-wiremock:
     networks:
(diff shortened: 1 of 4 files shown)


docker-compose.yml
```diff
@@ -216,22 +216,21 @@ services:
       - MYSQL_PORT=3306
       - MYSQL_DATABASE=activitypub
       - NODE_ENV=testing
+      - NODE_TLS_REJECT_UNAUTHORIZED=0
```
allouis, line 219: We are using self-signed certs from Caddy

features/step_definitions/index.js
```diff
@@ -33,100 +37,87 @@ BeforeAll(async () => {
...
-    ghostActivityPub.register(
-        {
-            method: 'GET',
-            endpoint: '/ghost/.well-known/jwks.json',
-        },
-        {
-            status: 200,
-            body: {
-                keys: [jwk],
+    await Promise.all([
```
allouis, line 40: We need to await wiremock registrations, otherwise we can make the request before wiremock is configured

features/step_definitions/index.js
```diff
@@ -33,100 +37,87 @@ BeforeAll(async () => {
...
     await resetDatabase();
-
-    const [siteId] = await getClient()('sites').insert({
-        host: new URL(process.env.URL_GHOST_ACTIVITY_PUB).host,
-        webhook_secret: getWebhookSecret(),
-    });
-
-    this.SITE_ID = siteId;
+    await Promise.all([
+        fetchActivityPub('https://self.test/.ghost/activitypub/site'),
```
allouis, line 110: This forces initialisation of the internal accounts before each test suite

features/step_definitions/index.js
```diff
@@ -33,100 +37,87 @@ BeforeAll(async () => {
...
+Before(async function setupState() {
+    const res = await fetch('https://self.test/.ghost/activitypub/users/index');
+    const actor = await res.json();
+
+    this.activities = {};
+    this.objects = {};
+    this.actors = {
+        Us: actor,
+    };
+    this.actors.Us.handle = '@index@self.test';
```
allouis, line 126: This is required by some tests to exist, and we're using the real AP Actor now, so we have to add this property manually

features/support/db.js
```diff
@@ -36,6 +36,5 @@ export async function reset() {
     await db('outboxes').truncate();
     await db('accounts').truncate();
     await db('users').truncate();
-    await db('sites').truncate();
```
allouis, line 39: Keeping this around because we have 4 sites to setup and don't want to slow tests down

features/support/wiremock.js
```diff
@@ -6,25 +6,25 @@ let externalActivityPub;
 /** @type {WireMock} */
 let ghostActivityPub;
 
-export function getExternalActivityPub() {
+export function getExternalWiremock() {
     if (!externalActivityPub) {
-        externalActivityPub = new WireMock(
-            process.env.URL_EXTERNAL_ACTIVITY_PUB,
-        );
+        externalActivityPub = new WireMock('http://fake-external-activitypub');
```
allouis: The wiremock hostnames now differ from the URLs because they are proxied to via Caddy - I've hardcoded for now

PR #1094, opened 2025-07-23
Added fix for post repository setting negative `reply_count` during a race condition

mike182uk:
ref https://linear.app/ghost/issue/PROD-2238

The issue seems to be occurring as part of a race condition when the system is trying to handle multiple deletes for the same post. This change adds in some defensive SQL to ensure this value can not go below zero

### DIFF
--- a/src/post/post.repository.knex.ts
+++ b/src/post/post.repository.knex.ts
@@ -271,7 +271,9 @@ export class KnexPostRepository {
                     if (post.inReplyTo) {
                         await transaction('posts')
                             .update({
-                                reply_count: this.db.raw('reply_count - 1'),
+                                reply_count: this.db.raw(
+                                    'CASE WHEN reply_count > 0 THEN reply_count - 1 ELSE 0 END',
+                                ),
                             })
                             .where({ id: post.inReplyTo });
                     }
--- a/src/post/post.repository.knex.integration.test.ts
+++ b/src/post/post.repository.knex.integration.test.ts
@@ -283,96 +283,67 @@ describe('KnexPostRepository', () => {
             expect(postRowAfterDelete.reply_count).toBe(0);
         });
 
-        it('negative reply count error repro', async () => {
-            // This simulates what happens in production when the same reply is
-            // deleted multiple times. This can occur when a delete is sent to
-            // many internal accounts at once:
-            // - foo.com is followed by bar.com, baz.com and qux.com
-            // - foo.com deletes a post
-            // - bar.com, baz.com and qux.com all receive the delete and try to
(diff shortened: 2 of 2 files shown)


src/post/post.repository.knex.ts
```diff
@@ -271,7 +271,9 @@ export class KnexPostRepository {
                     if (post.inReplyTo) {
                         await transaction('posts')
                             .update({
-                                reply_count: this.db.raw('reply_count - 1'),
+                                reply_count: this.db.raw(
+                                    'CASE WHEN reply_count > 0 THEN reply_count - 1 ELSE 0 END',
```
mike182uk: I did try using `GREATEST(reply_count - 1, 0)` here instead, but this still caused the issue. I wasn't totally sure why this was the case, but Claude had an explanation:

```
> so why didn't the GREATEST(reply_count - 1, 0) not work here?


⏺ Great question! The issue with GREATEST(reply_count - 1, 0) is about MySQL's order of operations.

  When MySQL evaluates:
  GREATEST(reply_count - 1, 0)

  It follows these steps:
  1. First, it evaluates the expression reply_count - 1
  2. Then, it passes that result to GREATEST()

  The problem occurs in step 1. When reply_count is 0 and it's an UNSIGNED integer column, MySQL tries to compute 0 - 1, which would be -1. But since the column is UNSIGNED, it can't store
  negative values, so MySQL throws the error before it even gets to the GREATEST() function.

  It's like trying to do this in code:
  unsigned int x = 0;
  unsigned int y = x - 1;  // ERROR HERE - can't store -1 in unsigned
  int result = Math.max(y, 0);  // Never reaches this line

  With CASE WHEN reply_count > 0 THEN reply_count - 1 ELSE 0 END, we:
  1. First check if reply_count > 0
  2. Only perform the subtraction if it's safe to do so
  3. Otherwise return 0

  This avoids ever trying to compute 0 - 1 for an unsigned integer, preventing the error entirely.

  The CASE WHEN approach is evaluating the condition first and choosing which branch to execute, rather than evaluating all expressions and then choosing the result.
```

src/post/post.repository.knex.ts
```diff
@@ -262,36 +262,44 @@ export class KnexPostRepository {
...
-                        await transaction('posts')
-                            .update({
-                                reply_count: this.db.raw('reply_count - 1'),
-                            })
-                            .where({ id: post.inReplyTo });
+                        .where({
+                            id: post.id,
+                            deleted_at: null,
+                        })
+                        .then((result) => result > 0);
```
allouis, line 277: Nice, now we have an actual boolean

allouis (approved):

PR #1095, opened 2025-07-23
Added tests for internal account activity delivery

allouis:

This adds tests for the existing functionality of deliverying activities to internal accounts.

I've broken the commits down to show how I approached writing these tests


allouis:
I'm sure there's some deduplication we can do here, but first we need to standardise our response values from endpoints etc...

sagzy (approved):
LGTM, besides the one typo spotted by the rabbit

PR #1097, opened 2025-07-23
Stopped sending activities to internal accounts

allouis:

Our domain layer handles all of the state updates for the the Create, Like, Announce, Delete & Undo activitites, which means we don't need to send them to the inboxes of internal accounts, as the state is already handled by the initial API request to create the activity.

We have tested this code in production gated to https://fabien.ghost.io

### DIFF
--- a/src/mq/gcloud-pubsub-push/mq.ts
+++ b/src/mq/gcloud-pubsub-push/mq.ts
@@ -155,10 +155,7 @@ export class GCloudPubSubPushMessageQueue implements MessageQueue {
                                 inboxUrl,
                             );
 
-                        if (
-                            !shouldDeliver &&
-                            message.baseUrl === 'https://fabien.ghost.io'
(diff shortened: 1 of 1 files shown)


allouis:
The full changes to `mq.ts`: https://github.com/TryGhost/ActivityPub/compare/b322e78c7839a602bcaae5b8800ce8a508a0d293...internal-account-activity-delivery#diff-e65a56fd3d198c7efbe714e0e57f6e07161f9901cec10115d0c9a4628ba41eea

The full changes to `account.service.ts`: https://github.com/TryGhost/ActivityPub/compare/b322e78c7839a602bcaae5b8800ce8a508a0d293...internal-account-activity-delivery#diff-7d84197d78536e8e5e6e64512cb6d3865b8cd897eea4c34da4631cd075176f08

vershwal (approved):

PR #1103, opened 2025-07-24
Ensured calculation of a negative reply count does not happen

mike182uk:
ref https://linear.app/ghost/issue/PROD-2332

It was possible to get a negative reply count if the post reply count had been updated  via another process whilst the post was being updated


src/post/post.repository.knex.ts
```diff
@@ -403,7 +403,7 @@ export class KnexPostRepository {
                             .update({
                                 repost_count: post.isInternal
                                     ? transaction.raw(
-                                          `repost_count + ${insertedRepostsCount - removedRepostsCount}`,
+                                          `GREATEST(0, CAST(repost_count AS SIGNED) + ${insertedRepostsCount - removedRepostsCount})`,
```
mike182uk, line 436: Why use `CAST`? From the AI:

```
 In MySQL, the repost_count column is defined as BIGINT UNSIGNED. When MySQL performs arithmetic operations on unsigned integers, it tries to keep the result unsigned as well.

  Here's what happens:

  1. Without CAST:
    - repost_count is 1 (unsigned)
    - We try to calculate: 1 + (-2)
    - MySQL attempts to calculate this as unsigned arithmetic
    - Since unsigned integers can't be negative, MySQL throws an error: "BIGINT UNSIGNED value is out of range"
  2. With CAST AS SIGNED:
    - repost_count is cast to a signed integer first
    - The calculation becomes: 1 + (-2) = -1 (as a signed integer)
    - Then GREATEST(0, -1) evaluates to 0
    - The final result 0 is valid for an unsigned column

  The CAST is necessary because MySQL's type system is strict about unsigned arithmetic. It won't allow intermediate calculations that would result in negative values, even if the final
  result (after applying GREATEST) would be non-negative.
```

mike182uk:
This is now out of date so closing

PR #1119, opened 2025-07-28
Added debug logging for `repost_count`

mike182uk:
ref https://linear.app/ghost/issue/PROD-2332

Added debug logging for `repost_count` so we can work out why the value is going negative

### DIFF
--- a/src/post/post.repository.knex.ts
+++ b/src/post/post.repository.knex.ts
@@ -409,7 +409,7 @@ export class KnexPostRepository {
                     );
 
                     if (insertedRepostsCount - removedRepostsCount !== 0) {
-                        this.logger.debug(
+                        this.logger.info(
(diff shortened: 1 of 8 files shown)


src/post/post.repository.knex.ts
```diff
@@ -385,20 +387,47 @@ export class KnexPostRepository {
...
+                    if (repostsToRemove.length > 0) {
+                        const res = await this.removeReposts(
+                            post,
+                            repostsToRemove,
+                            transaction,
+                        );
+
+                        if (res !== 0) {
+                            removedRepostsDb = res;
+                        }
```
mike182uk, line 401: this is so we can see if `this.removeReposts` is returning `0`

allouis (approved):
This looks good, but we're gonna have to enable debug logs get these to show up!

PR #1132, opened 2025-07-30
Added handling for a site being disabled / enabled

mike182uk:
ref https://linear.app/ghost/issue/PROD-2351

When a site is disabled (via `DELETE` request to `/v1/sites`) the `site` record will be removed from the database. This will cascade and remove the `user` associated with the site, the feed entries associated with the user and the notifications associated with the user. This will leave the `account` in-tact. If the site is later re-enabled, a new site & user record is generated. The user record will be linked to the previously left over account. If there was a previously left over account, it will retain its keypair in case other servers have cached this

### DIFF
--- a/features/step_definitions/index.js
+++ b/features/step_definitions/index.js
@@ -77,7 +77,7 @@ BeforeAll(async function setupWiremock() {
     ]);
 });
 
-BeforeAll(async function setupSelfSite() {
+async function setupSelfSite() {
     const res = await fetchActivityPub(
         'https://self.test/.ghost/activitypub/v1/site',
     );
@@ -91,9 +91,7 @@ BeforeAll(async function setupSelfSite() {
     await getClient()('sites')
         .update('webhook_secret', getWebhookSecret())
         .where('id', '=', json.id);
-
-    this.SITE_ID = json.id;
-});
+}
 
 BeforeAll(async function setupLocalSites() {
     await Promise.all([
@@ -107,7 +105,7 @@ Before(async function reset() {
     await resetWiremock();
     await resetDatabase();
     await Promise.all([
-        fetchActivityPub('https://self.test/.ghost/activitypub/v1/site'),
+        setupSelfSite(),
         fetchActivityPub('https://alice.test/.ghost/activitypub/v1/site'),
         fetchActivityPub('https://bob.test/.ghost/activitypub/v1/site'),
         fetchActivityPub('https://carol.test/.ghost/activitypub/v1/site'),
--- a/features/step_definitions/site_steps.js
+++ b/features/step_definitions/site_steps.js
@@ -3,10 +3,8 @@ import { Given, When } from '@cucumber/cucumber';
 import { getClient } from '../support/db.js';
 import { fetchActivityPub } from '../support/request.js';
 
-Given('there is no entry in the sites table', async function () {
+Given('there is no entry in the sites table', async () => {
     await getClient()('sites').del();
-
-    this.SITE_ID = null;
 });
 
 When('we request the site endpoint', async function () {


features/step_definitions/activitypub_steps.js
```diff
@@ -166,6 +166,45 @@ When('we request the outbox', async function () {
...
+            headers: {
+                Accept: 'application/ld+json',
+            },
+        },
+    );
+});
+
+When('we request the webfinger', async function () {
+    this.response = await fetchActivityPub(
+        'https://self.test/.well-known/webfinger?resource=acct:test@self.test',
```
allouis: should this be `index@self.test`?

features/disable-site.feature
```diff
@@ -0,0 +1,47 @@
+Feature: Disabling a site
+
+  Scenario: Disabling a site prevents access to the outbox
+    When an authenticated "delete" request is made to "/.ghost/activitypub/v1/site"
```
allouis: IMO this should be something like

```
Given a site is disabled
When we request the outbox
Then the request is rejected with a 403
```

src/http/api/site.controller.ts
```diff
@@ -41,9 +41,33 @@ export class SiteController {
...
+            return new Response(JSON.stringify({ error: 'No Host header' }), {
+                status: 401,
+            });
+        }
+
+        try {
+            const result = await this.siteService.disableSiteForHost(host);
+
+            return new Response(null, {
+                status: result ? 200 : 404,
```
allouis: I think an `if` block is preferable here but nabd

src/account/account.service.ts
```diff
@@ -235,7 +235,15 @@ export class AccountService {
             // a potential migration from a different server.
             const newKeyPair = await this.generateKeyPair();
             await this.db('accounts')
-                .where({ id: existingAccount.id })
+                .where({
+                    id: existingAccount.id,
+                })
+                // If the account already has a private key, we shouldn't replace
+                // the keypair in case other servers have cached the keypair
+                .whereNull('ap_private_key')
```
allouis: `generateKeyPair` is a fairly costly function, it is quite slow! It might be worth checking first and conditionally generating it

allouis (approved):

features/disable-site.feature
mike182uk: Yep agree 👍

mike182uk: Fixed in https://github.com/TryGhost/ActivityPub/pull/1132/commits/8386c231fd858d8da683cb0dce549b538796aeba

src/http/api/site.controller.ts
mike182uk: Fixed in https://github.com/TryGhost/ActivityPub/pull/1132/commits/29f5c15faf245fef5dbf61e64b566115bc40fb3a

features/step_definitions/activitypub_steps.js
mike182uk: Fixed in https://github.com/TryGhost/ActivityPub/pull/1132/commits/2bb1a03ab432e93f86684a4e59c6741da7eb70c0

src/account/account.service.ts
mike182uk: Updated in https://github.com/TryGhost/ActivityPub/pull/1132/commits/e63d1d63c59dae3d308c93498153902f777ad736

features/disable-site.feature
```diff
@@ -0,0 +1,40 @@
...
+    Then the request is rejected with a 403
+
+  Scenario: Disabled site can not be followed
+    Given an Actor "Person(Alice)"
+    Given a "Follow(Us)" Activity "Follow" by "Alice"
+    When we disable the site
+    And "Alice" sends "Follow" to the Inbox
+    Then the request is rejected with a 403
+
+  Scenario: Disabled site can not be re-enabled
```
allouis: Small typo - I think this should be `can be re-enabled`

src/account/account.service.ts
```diff
@@ -230,20 +230,33 @@ export class AccountService {
...
+            const hasPrivateKey =
+                [null, ''].includes(
+                    (
+                        await this.db('accounts')
+                            .select('ap_private_key')
+                            .where({
+                                id: existingAccount.id,
+                            })
+                            .first()
+                    )?.ap_private_key,
```
allouis: Can also just cast to boolean here

```ts

const hasPrivateKey = !!(await this.db('accounts')
    .select('ap_private_key')
    .where({
        id: existingAccount.id,
    })
    .first()?
    .ap_private_key)
```

features/disable-site.feature
mike182uk: Fixed in https://github.com/TryGhost/ActivityPub/pull/1132/commits/b5e24e03f333430aa6be645482c561829bc1519f

src/account/account.service.ts
mike182uk: Yeh this is a little less verbose

Updated in https://github.com/TryGhost/ActivityPub/pull/1132/commits/5f0ebecb0e7de600618d1209b8f2834f027d89fa

PR #1146, opened 2025-08-04
Ignored preloaded documents from the KvStore

allouis:
We don't want to ever read or store these documents in our store, Fedify already has them hardcoded in memory, but due to a bug it will sometimes attempt to read from the cache before reading from memory.


allouis:
Superceded by https://github.com/TryGhost/ActivityPub/pull/1151 which includes https://github.com/fedify-dev/fedify/pull/352

PR #1154, opened 2025-08-06
Improved 403 responses to distinguish between blocks

allouis:
ref https://linear.app/ghost/issue/PROD-2402

The "Account Suspended" screen currently shows on any 403, which is incorrect because we use a 403 status code for missing roles as well as for sites which have not been configured in the database. Added a body with a code will allow us to distinguish between the errors.


src/http/middleware/role-guard.ts
```diff
@@ -51,10 +51,18 @@ export function createRoleMiddleware(jwksCache: KvStore) {
...
         const host = request.header('host');
         if (!host) {
-            ctx.get('logger').info('No Host header');
-            return new Response('No Host header', {
-                status: 401,
-            });
+            ctx.get('logger').error('No Host header');
+            return new Response(
+                JSON.stringify({
+                    error: 'Unauthorized',
```
mike182uk, line 57: Do we want a `code` for every error?

PR #1161, opened 2025-08-06
Added ADR for frontend versioning strategy

allouis:
**TL;DR** Instead of the frontend version being determined by Ghost version, causing mismatches, the ActivityPub backend will have an endpoint that tells the frontend loader which CDN bundle URL to use. This guarantees the frontend and backend versions are always compatible - each backend version declares its matching frontend version. Simple fix, minimal code changes, solves all the deployment scenarios.

### DIFF
--- a/adr/frontend-versioning.md
+++ b/adr/frontend-versioning.md
@@ -1,0 +1,74 @@
+# ActivityPub Service Frontend Versioning via Backend Configuration
+
+## Status
+
+Proposed
+
+## Context
+
+We have an ActivityPub service with multiple deployment scenarios that creates version compatibility challenges:
+
+**Current Architecture:**
+- ActivityPub frontend is dynamically imported as a JS bundle into Ghost Admin
+- Frontend bundle is served from jsDelivr CDN
+- Backend API has versioned endpoints (`/activitypub/v1/*`)
+
+**Deployment Scenarios:**
+1. **Ghost Pro hosted**: Users always run the latest backend and frontend versions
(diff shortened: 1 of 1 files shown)


kevinansfield:
At first read-through this sounds like a really nice approach! A couple of questions...

1. How are upgrades to the backend with breaking changes handled for already-running clients?
    - E.g. do all backend responses include the compatible client version so clients can react accordingly with an auto-reload or notification when they get a version-incompatible response/error?
3. Are there any dependencies between the client and the client's parent app? If so, how do those get handled?
    - E.g. for Koenig there's a fair amount of data and functions passed through from the host app (ghost) to the client app (editor) so there's a coupled dependency across both apps for certain types of changes. Does ActivityPub have anything similar?

allouis:
Great Q's!

1. Yeah, I think we need to include a header in the response that the client can do something with, either a forced reload, or a notification. I think we can just respond with the required semver for the client 🤔
2. The client app depends on the admin-x-framework code, which handles things like allowing us to read/update route params, I don't think there is much we can do here, other than not making breaking changes in the framework. We also have a dependency in Admin on the notifications count endpoint, which will need to be treated as a special case and not have breaking changes made to it.

kevinansfield:
> The client app depends on the admin-x-framework code

Does that get bundled with the AP client build or does it depend on an already-loaded version provided by the host app?

allouis:
Good point, that is bundled with the AP client!

allouis:
There are two linked PR's which show a proof of concept for this implementation

PR #1172, opened 2025-08-07
Used atomic query for updating repost count outside of manual updates

mike182uk:
ref https://linear.app/ghost/issue/PROD-2422

When concurrent `announce` / `undo` (`announce`) activities are processed for external posts, the repost count could become negative due to using stale in-memory values

 Example:
 - Post does not exist in the db
 - We receive an `announce` and an `undo` for the post very close together
 - Both handlers try to create the post (as it does not exist)
 - One handler succeeds in creating the post and gets the result (a post with a repost count of 0)
 - The other handler fails gracefully (as we ignore duplicate inserts) and gets the same post, with a repost count of 0
 - `announce` handler adds repost and saves: 0 + 1 = 1
 - `undo` handler removes repost (which exists because `announce` has already saved), but calculates: 0 - 1 = -1

To resolve we use atomic SQL operations (repost_count + delta) for external posts when doing add/remove operations, preventing stale reads. Manual count updates via `setRepostCount()` are still preserved when the dirty flag is set

### DIFF
--- a/src/post/post.repository.knex.ts
+++ b/src/post/post.repository.knex.ts
@@ -410,6 +410,13 @@ export class KnexPostRepository {
                     );
 
                     if (insertedRepostsCount - removedRepostsCount !== 0) {
+                        // For external posts with manual count updates we need to
+                        // preserve the manual count while applying the delta.
+                        // For all other cases we should use atomic SQL to
+                        // prevent race conditions with concurrent updates
+                        const shouldUseAtomicUpdate =
+                            post.isInternal || !post.isRepostCountDirty;
+
                         this.logger.info(
                             `Updating repost count for post ${post.id}`,
                             {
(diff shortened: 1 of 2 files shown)


src/post/post.repository.knex.ts
```diff
@@ -410,6 +410,13 @@ export class KnexPostRepository {
                     );
 
                     if (insertedRepostsCount - removedRepostsCount !== 0) {
+                        // For external posts with manual count updates we need to
+                        // preserve the manual count while applying the delta.
+                        // For all other cases we should use atomic SQL to
+                        // prevent race conditions with concurrent updates
+                        const shouldUseAtomicUpdate =
+                            post.isInternal || !post.isRepostCountDirty;
```
sagzy, line 418: Shouldn't this be an AND, instead of an OR?

PR #1215, opened 2025-08-18
Updated Redis to use a different port than the default 6379

sagzy:
no issue

- Using a different port for Redis than the default 6379 helps to avoid conflicts with other running Redis instances (e.g. in Ghost)


mike182uk:
Superseded by https://github.com/TryGhost/ActivityPub/pull/1206/

PR #1295, opened 2025-09-11
Added job to fix reply counts

mike182uk:
ref https://linear.app/ghost/issue/PROD-2420

Added one-off job that will fix reply counts on posts that were created before 24th July 2025 where a it was possible for a race condition to occur which would result in the `reply_count` field on the post being incorrect

### DIFF
--- a/jobs/fix-reply-counts/index.ts
+++ b/jobs/fix-reply-counts/index.ts
@@ -17,9 +17,9 @@ export async function findPostsToFix(
     pool: mysql.Pool,
     cutoffDate: string,
     batchSize: number,
+    fixZeroOnly: boolean = false,
 ): Promise<PostToFix[]> {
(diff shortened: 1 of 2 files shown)


jobs/fix-reply-counts/index.ts
```diff
@@ -0,0 +1,157 @@
...
+        } else {
+            await new Promise((resolve) => setTimeout(resolve, delayMs));
+        }
+    }
+
+    return totalFixed;
+}
+
+async function main() {
+    const connection = await mysql.createConnection(
```
allouis: We are going to make 100 concurrent updates on this connection, should we use a pool instead?

mike182uk: Yeh a pool with give us true parallelisation which will make execution quicker - I've set the limit to `10` so we don't hammer the db whilst this is running. Do you think this is ok or too conservative?

PR #1311, opened 2025-09-18
Adding ADRs with Claudette

sagzy:


adr/0003-result-pattern.md
```diff
@@ -0,0 +1,86 @@
...
+## Context
+
+TypeScript's exception-based error handling creates implicit control flow and lacks type safety for error cases. The ActivityPub server requires explicit, type-safe error handling for network failures, validation errors, and business rule violations.
+
+## Decision
+
+Adopt the Result Pattern for explicit error handling in the business logic layer.
+
+```typescript
+type Result<T, E> = { ok: true; value: T } | { ok: false; error: E };
```
allouis, line 18: This is wrong 😛 `type Result <T, E> = [null, T] | [E, null]`

Issue #1230, opened 2025-08-24
Suspected incorrect key algorithm on createInternalAccount


mqdeandrade:
CAVEAT: I may be misunderstanding how dependencies are initialised, so apologies if that is the case

[`createInternalAccount`](https://github.com/TryGhost/ActivityPub/blob/d6c2203bbc30bc68ed2f648c616e1318fa2b58c5/src/account/account.service.ts#L177C11-L177C32) seems to be called when first enabling Network from the Ghost Admin settings
It in turn calls `generateKeyPair` which has a [default](https://github.com/TryGhost/ActivityPub/blob/d6c2203bbc30bc68ed2f648c616e1318fa2b58c5/src/account/account.service.ts#L67) of [generateCryptoKeyPair from fedify](https://github.com/fedify-dev/fedify/blob/9dd4221c859c46d638dab38e419aa2654b729fac/packages/fedify/src/sig/key.ts#L63)
`generateKeyPair` is called without arguments, and looking at the implementation in fedify, calling without arguments will create a `RSASSA-PKCS1-v1_5` key.
I can confirm that this behaviour happens in practice, as my ActivityPub logs show `No algorithm specified.  Using RSASSA-PKCS1-v1_5 by default, but it is recommended to specify the algorithm explicitly as the parameter will be required in the future.` when enabling Network in settings.

I logged in to the DB container and confirmed that the private key for my account is`"alg": "RS256"`
When trying to follow another account, I get the following error:
```
activitypub-1  | 12:51:22.026 WRN fedify·federation·outbox: No supported key found to create a proof for the activity 'http://miggl.es/.ghost/activitypub/follow/f0d92341-1865-4037-9e65-0c581da29d88'.  The activity will be sent without a proof.  In order to create a proof, at least one Ed25519 key must be provided.
activitypub-1  | 12:51:22.026 INF activitypub: KnexKvStore: Get key _fedify,httpMessageSignaturesSpec,https://www.404media.co
activitypub-1  | 12:51:22.584 ERR fedify·federation·outbox: Failed to send activity 'http://miggl.es/.ghost/activitypub/follow/f0d92341-1865-4037-9e65-0c581da29d88' to 'https://www.404media.co/.ghost/activitypub/inbox/index' (401 'Unauthorized'):
activitypub-1  | 'The signer and the actor do not match.'
[...]

PR #1316, opened 2025-09-22
Fix key attachment bug and add dual-key support

schuettc:
# Fix key attachment bug and add dual-key support for ActivityPub federation

## Description

This PR addresses a bug that prevents the Ghost ActivityPub implementation from federating with other ActivityPub servers. The issue stems from cryptographic keys not being properly attached to the Person object, causing authentication failures when other servers attempt to verify signatures.

Additionally, this PR enhances compatibility across the Fediverse by implementing dual-key support (both Ed25519 and RSA), ensuring broader interoperability with different ActivityPub implementations.

### Current Behavior
The current implementation uses `publicKeys` (plural) property which is not recognized by Fedify:

```typescript
publicKeys: (await ctx.getActorKeyPairs(identifier)).map(
    (key) => key.cryptographicKey,
),
```

This results in:
- Keys being generated and stored but not properly attached to the Person object
- Authentication failures when other servers attempt to verify HTTP signatures
- Follow requests failing silently
- Posts not federating to other servers
- WebFinger discovery working but actual federation failing

### Root Cause
According to the [Fedify documentation](https://fedify.dev/manual/actor), the Person class expects:
- `publicKey` (singular) - Contains a `CryptographicKey` instance for HTTP Signatures
- `assertionMethods` (plural) - An array of `Multikey` instances for Object Integrity Proofs

The current code uses `publicKeys` (plural) which is not a valid Person property in Fedify.

### 1. Fixed Key Attachment
Updated `actorDispatcher` to use the correct Fedify properties:

```typescript
// Get key pairs for proper attachment
const keyPairs = await ctx.getActorKeyPairs(identifier);

const person = new Person({
    // ... other properties ...

    // FIX: Use correct Fedify properties for keys
    publicKey: keyPairs[0]?.cryptographicKey, // For HTTP Signatures
    assertionMethods: keyPairs.map((kp) => kp.multikey), // For Object Integrity Proofs
});
```

### 2. Dual-Key Support
Implemented generation of both Ed25519 and RSA keys for maximum compatibility:

```typescript
const ed25519Keys = await generateCryptoKeyPair('Ed25519');
const rsaKeys = await generateCryptoKeyPair('RSASSA-PKCS1-v1_5');
return [ed25519Keys, rsaKeys];
```

### 3. Backwards Compatibility
The implementation maintains full backwards compatibility:
- Existing single-key accounts continue to work
- New accounts automatically get dual keys
- No database migration required
- The `keypairDispatcher` handles both formats transparently

### Files Modified (6 files total)
1. **src/dispatchers.ts**
   - Fixed `actorDispatcher` to attach keys using correct properties
   - Updated `keypairDispatcher` to handle dual-key JSON format

2. **src/account/account.service.ts**
   - Modified key generation to create both Ed25519 and RSA keys
   - Updated account creation to handle dual keys
   - Fixed duplicate account handling for dual keys

3. **src/account/account.repository.knex.ts**
   - Added support for storing dual keys in JSON array format
   - Maintains backwards compatibility with single-key format

4. **src/account/account.service.unit.test.ts**
   - Updated type signature to match new key array format

5. **src/test/account-entity-test-helpers.ts**
   - Updated test helpers to work with key arrays

6. **src/test/crypto-key-pair.ts**
   - Modified test key generation to create dual keys

### Automated Tests
All tests pass with minimal modifications to support key arrays:
- ✅ Type checking (`yarn test:types`)
- ✅ Unit tests (`yarn test:unit`)
- ✅ Integration tests (`yarn test:integration`)

Modified test files to support dual-key arrays:
- `src/test/crypto-key-pair.ts` - Generate dual keys for tests
- `src/test/account-entity-test-helpers.ts` - Use first key from array
- `src/account/account.service.unit.test.ts` - Updated type signature

### Production Verification
These changes are running in production at https://subaud.io:

1. **Keys properly attached**:
```bash
# Verify publicKey exists
curl -s -H "Accept: application/activity+json" \
  "https://subaud.io/.ghost/activitypub/users/index" | \
  jq '.publicKey.publicKeyPem'
# Returns: Valid PEM-formatted public key

# Verify dual keys in assertionMethod
curl -s -H "Accept: application/activity+json" \
  "https://subaud.io/.ghost/activitypub/users/index" | \
  jq '.assertionMethod | length'
# Returns: 2 (Ed25519 + RSA)
```

2. **WebFinger discovery working**:
```bash
curl "https://subaud.io/.well-known/webfinger?resource=acct:index@subaud.io"
# Returns valid WebFinger response with ActivityPub links
```

3. **Dual-key types confirmed**:
   - Ed25519 key: Multibase identifier starts with `z6Mk`
   - RSA key: Longer multibase identifier starts with `zggh`

4. **Federation capability** - The site is discoverable and keys are properly formatted for federation. 

## Issues

Resolves: https://github.com/TryGhost/ActivityPub/issues/1230

### DIFF
--- a/src/account/account.repository.knex.ts
+++ b/src/account/account.repository.knex.ts
@@ -59,10 +59,24 @@ export class KnexAccountRepository {
                 ap_outbox_url: draft.apOutbox?.href ?? null,
                 ap_following_url: draft.apFollowing?.href ?? null,
                 ap_liked_url: draft.apLiked?.href ?? null,
-                ap_public_key: JSON.stringify(
-                    await exportJwk(draft.apPublicKey),
-                ),
-                ap_private_key: draft.apPrivateKey
+                ap_public_key: (draft as any).dualKeyPairs
+                    ? JSON.stringify({
+                          keys: await Promise.all(
+                              (draft as any).dualKeyPairs.map((kp: { publicKey: CryptoKey, privateKey: CryptoKey }) =>
+                                  exportJwk(kp.publicKey)
+                              )
+                          )
+                      })
+                    : JSON.stringify(await exportJwk(draft.apPublicKey)),
+                ap_private_key: (draft as any).dualKeyPairs
+                    ? JSON.stringify({
+                          keys: await Promise.all(
+                              (draft as any).dualKeyPairs.map((kp: { publicKey: CryptoKey, privateKey: CryptoKey }) =>
+                                  exportJwk(kp.privateKey)
+                              )
+                          )
+                      })
+                    : draft.apPrivateKey
                     ? JSON.stringify(await exportJwk(draft.apPrivateKey))
                     : null,
                 custom_fields: draft.customFields
--- a/src/account/account.service.ts
+++ b/src/account/account.service.ts
@@ -64,7 +64,12 @@ export class AccountService {
         private readonly events: AsyncEvents,
         private readonly accountRepository: KnexAccountRepository,
         private readonly fedifyContextFactory: FedifyContextFactory,
-        private readonly generateKeyPair: () => Promise<CryptoKeyPair> = generateCryptoKeyPair,
+        private readonly generateKeyPair: () => Promise<{ publicKey: CryptoKey, privateKey: CryptoKey }[]> = async () => {
+            // Generate both Ed25519 and RSA for maximum compatibility
+            const ed25519Keys = await generateCryptoKeyPair('Ed25519');
+            const rsaKeys = await generateCryptoKeyPair('RSASSA-PKCS1-v1_5');
+            return [ed25519Keys, rsaKeys];
+        },
     ) {}
 
     /**
@@ -178,7 +183,7 @@ export class AccountService {
         site: Site,
         internalAccountData: InternalAccountData,
     ): Promise<AccountType> {
-        const keyPair = await this.generateKeyPair();
+        const keyPairs = await this.generateKeyPair();
 
         const normalizedHost = site.host.replace(/^www\./, '');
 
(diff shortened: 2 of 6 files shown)


allouis:
@schuettc I'm not seeing that public keys are not attached to the Person objects 🤔 

```
curl -H "Accept: application/activity+json" \
        "https://activitypub.ghost.org/.ghost/activitypub/users/index" | jq '.publicKey'
```

You're right that the plural is deprecated by fedify, but it does still work and is correctly attached - is there a specific bug you're running into without these changes? If we can get replication steps that would be awesome

src/account/account.repository.knex.ts
```diff
@@ -59,10 +59,24 @@ export class KnexAccountRepository {
                 ap_outbox_url: draft.apOutbox?.href ?? null,
                 ap_following_url: draft.apFollowing?.href ?? null,
                 ap_liked_url: draft.apLiked?.href ?? null,
-                ap_public_key: JSON.stringify(
-                    await exportJwk(draft.apPublicKey),
-                ),
-                ap_private_key: draft.apPrivateKey
+                ap_public_key: (draft as any).dualKeyPairs
```
allouis, line 62: We shouldn't be casting to `any` here, the draft type should be updated

src/account/account.repository.knex.ts
```diff
@@ -59,10 +59,24 @@ export class KnexAccountRepository {
...
+                    ? JSON.stringify({
+                          keys: await Promise.all(
```
allouis, line 64: I don't love the idea of storing two different data types in the same column here.

I think we need a discussion about either:

1. Migrating all columns to store multiple keys
2. Adding new columns for the new keys

src/account/account.service.ts
```diff
@@ -64,7 +64,12 @@ export class AccountService {
         private readonly events: AsyncEvents,
         private readonly accountRepository: KnexAccountRepository,
         private readonly fedifyContextFactory: FedifyContextFactory,
-        private readonly generateKeyPair: () => Promise<CryptoKeyPair> = generateCryptoKeyPair,
+        private readonly generateKeyPair: () => Promise<{ publicKey: CryptoKey, privateKey: CryptoKey }[]> = async () => {
```
allouis, line 67: This method name needs updating to reflect the new return type (singular -> plural).

Why have we stopped using the `CryptoKeyPair` type here?

PR #1325, opened 2025-09-23
Fixed flakey test for unread notifications

sagzy:
closes https://linear.app/ghost/issue/PROD-2663

- The notification count test scenario was flaky when running the entire cucumber test suite, returning a count of 6 instead of 4, sometimes
- We’re not testing the exact count of unread notifications anymore, but rather whether new notifications are marked as unread / visiting the notifications marks them as read - which is closer to the main use case for the notification badge

### DIFF
--- a/features/notifications.feature
+++ b/features/notifications.feature
@@ -7,22 +7,14 @@ Feature: Notifications
     When an authenticated request is made to "/.ghost/activitypub/v1/notifications?limit=200"
     Then the request is rejected with a 400
 
-  Scenario: Requests for unread notifications count
+  Scenario: Getting unread notifications count
     Given we are following "Alice"
     And we are not following "Bob"
+    And we reset unread notifications count
     When we get a like notification from "Alice"
     And we get a like notification from "Bob"
     And we get a reply notification from "Alice"
     And we get a reply notification from "Bob"
     Then the unread notifications count is 4
-
-  Scenario: Reset unread notifications count
-    Given we are following "Alice"
-    And we are not following "Bob"
-    And we get a like notification from "Alice"
-    And we get a like notification from "Bob"
-    And we get a reply notification from "Alice"
-    And we get a reply notification from "Bob"
-    And the unread notifications count is 4
     When we reset unread notifications count
     Then the unread notifications count is 0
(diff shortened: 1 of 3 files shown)


features/notifications.feature
```diff
@@ -7,22 +7,18 @@ Feature: Notifications
...
     And we are not following "Bob"
     When we get a like notification from "Alice"
-    And we get a like notification from "Bob"
-    And we get a reply notification from "Alice"
     And we get a reply notification from "Bob"
-    Then the unread notifications count is 4
+    Then we have unread notifications
 
-  Scenario: Reset unread notifications count
+  Scenario: Visiting the notifications page marks notifications as read
```
mike182uk, line 17: I don't think we should talk about "pages" here - I know this aligns better with the action the user is taking, but its not a pattern we follow in any other scenarios and is potentially confusing / misleading. What about just: 

`Retrieving notifications marks them all as read`

features/support/notifications.js
```diff
@@ -68,23 +67,59 @@ export async function waitForUnreadNotifications(
...
+        await new Promise((resolve) => setTimeout(resolve, options.delay));
+    }
+
+    return await waitForUnreadNotifications({
+        retryCount: options.retryCount + 1,
+        delay: options.delay + 500,
+    });
+}
+
+export async function waitForZeroUnreadNotifications(
```
mike182uk, line 92: I'm not following how this is any different to `waitForUnreadNotifications(0)`?

features/notifications.feature
```diff
@@ -7,22 +7,18 @@ Feature: Notifications
...
-  Scenario: Requests for unread notifications count
+  Scenario: New notifications are marked as unread
     Given we are following "Alice"
     And we are not following "Bob"
     When we get a like notification from "Alice"
-    And we get a like notification from "Bob"
-    And we get a reply notification from "Alice"
     And we get a reply notification from "Bob"
-    Then the unread notifications count is 4
+    Then we have unread notifications
```
mike182uk, line 15: 🤔 My concern is here we are just masking a problem? It is valid that we do expect a specific amount of notifications, and if there are more or less, than something is not correct somewhere

sagzy:
Closing, as the solution doesn't feel 100% right

PR #1328, opened 2025-09-25
Fix federation error handling to prevent 500 responses

betschki:
When USE_MQ=false, federation happens synchronously during the HTTP request. If delivery to any follower fails (e.g., deleted accounts, down servers), it causes the entire request to return 500 even though the action itself succeeded.


sagzy:
Hey @betschki,

Sorry we've been quiet here. We've been making some progress on this issue internally, with a more comprehensive solution. It's not quite ready just yet, but getting there!

We'll update you once the fix is out, I'm closing this PR in the meantime 🙏

PR #1377, opened 2025-10-14
Added bluesky handle confirmation endpoint

mike182uk:
no ref

We found situations where the computed Bluesky handle did not match what Bridgy had actually used for the account - This is because Bridgy has some conditional logic based on the domain setup. We do not want to emulate this logic so we have instead changed our approach to require handle confirmation. The client will request handle confirmation in which we will look up the Bluesky account that should have been created for the account and use the actual handle that was assigned


docs/bluesky-sharing.md
```diff
@@ -0,0 +1,127 @@
...
+      Bluesky->>ActivityPub: Return actor
+      ActivityPub->>Client: 200 OK (enabled: true, handleConfirmed: true, handle: "@test.example.com.ap.brid.gy")
+
+      Note over Client,Bluesky: Disabling Bluesky Sharing
+
+      Client->>ActivityPub: POST /.ghost/activitypub/v1/actions/bluesky/disable
+      ActivityPub->>Bridgy Fed: Send "stop" DM
+      ActivityPub->>Bridgy Fed: Send unfollow activity
+      ActivityPub->>Client: 204 No Content
+```
```
sagzy, line 128: This doc is 💯! 

I'm curious: is that written by you directly, or did you somehow manage to get AI to write concisely, without the usual noise?

src/integration/bluesky.service.ts
```diff
@@ -23,26 +33,33 @@ export class BlueskyService {
...
+                'bluesky_integration_account_handles',
+            )
+                .where('account_id', account.id)
+                .first();
+
+            return {
+                enabled: true,
+                handleConfirmed: existing?.confirmed || false,
+                handle: existing?.handle || null,
+            };
```
sagzy, line 60: nitpick: `this.isEnabledForAccount()` and then `const existing = ...` both do the same database query. It may be worth simplifying to:

```
const existing = await this.db('bluesky_integration_account_handles')
    .where('account_id', account.id)
    .first();

if (existing) {
    this.logger.info(
        `Bluesky integration already enabled for account {id}`,
        { id: account.id },
    );

   return {
        enabled: true,
        handleConfirmed: existing.confirmed,
        handle: existing.handle,
    };
}
```

This also removes the need for optional chaining in the returned object

src/integration/bluesky.utils.ts
```diff
@@ -0,0 +1,38 @@
...
+): string | null {
+    // Find Bridgy Fed actors
+    const bridgyActors = actors.filter((actor) => {
+        if (!actor.labels) return false;
+
+        return actor.labels.some((label) => label.val === BRIDGY_FED_LABEL);
+    });
+
+    // Find actors with a handle ending in .{domain}.ap.brid.gy
+    const expectedHandleSuffix = `.${domain}.ap.brid.gy`;
```
sagzy: question: I'm curious if this handles sites on www as expected?

For a site hosted on www, e.g. www.sambinder.com, `accounts.apId.hostname` is `www.sambinder.com`

However, as Bluesky Fed strips www (sometimes), the search results may return a handle without www, e.g. `art.sambinder.com.ap.brid.gy` - which won't match `.www.sambinder.com.ap.brid.gy` suffix?

https://public.api.bsky.app/xrpc/app.bsky.actor.searchActors?q=www.sambinder.com

sagzy (commented):
Code looks really good! I left one nitpick and one question on what might be a bug

docs/bluesky-sharing.md
mike182uk: Yeh this was mostly hand written - I did start out with asking Claude to document the feature, but what it created was far too verbose and focused too much on the implementation rather than the API contract

src/integration/bluesky.service.ts
mike182uk: Yeh good catch, this is much cleaner

src/integration/bluesky.utils.ts
mike182uk: Yeh good catch, we need to normalize the domain first before it can be used to match on 👍

src/integration/bluesky.service.ts
mike182uk: Updated in https://github.com/TryGhost/ActivityPub/pull/1377/commits/262153c174dcbcb7b253d533353f81fbb54fbc93

src/integration/bluesky.utils.ts
mike182uk: Fixed in https://github.com/TryGhost/ActivityPub/pull/1377/commits/935440263d23f20886566b5422e5c0bdac5caddc

docs/bluesky-sharing.md
sagzy: Yeah same experience - quick to write docs, but quality is meh / way too much noise

PR #1385, opened 2025-10-20
Extracted helper to map feed results to post DTOs

sagzy:
ref https://linear.app/ghost/issue/BER-2908/add-dummy-api-endpoint-to-start-vertical-integration

- in the context of introducing the global feed, extracted the mapping from a feed result to a post DTO, so that it can reused


src/http/api/helpers/feed.ts
```diff
@@ -0,0 +1,63 @@
...
+                      mediaType: attachment.mediaType ?? '',
+                      name: attachment.name ?? '',
+                      url: attachment.url,
+                  }))
+                : [],
+            author: {
+                id: result.author_id.toString(),
+                handle: getAccountHandle(
+                    result.author_url ? new URL(result.author_url).host : '',
+                    result.author_username,
```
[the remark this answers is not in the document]
sagzy: Yeah good one; made that safer in 06c3790f9dd2e00a2498112a6d12578c6aa11601

src/http/api/helpers/feed.ts
```diff
@@ -0,0 +1,63 @@
...
+            repostedByMe: result.post_reposted_by_user === 1,
+            repostedBy: result.reposter_id
+                ? {
+                      id: result.reposter_id.toString(),
+                      handle: getAccountHandle(
+                          result.reposter_url
+                              ? new URL(result.reposter_url).host
+                              : '',
+                          result.reposter_username,
+                      ),
```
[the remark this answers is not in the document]
sagzy: Yeah good one; made that safer in 06c3790f9dd2e00a2498112a6d12578c6aa11601

src/http/api/feed.controller.ts
```diff
@@ -63,62 +62,7 @@ export class FeedController {
             cursor,
         });
 
-        const posts: PostDTO[] = results.map((result) => {
```
mike182uk, line 66: How come you decided to extract this out to a helper in a different file, rather than keep co-located in this class (i.e via a private method). Its unlikely to be used anywhere else outside this controller right?

sagzy: Yeah that's fair :) Only reason was to follow the existing pattern of https://github.com/TryGhost/ActivityPub/blob/main/src/http/api/helpers/post.ts - but I like the cohesion order argument better, i.e. keep things that are coupled next to each other, whenever possible

mike182uk: Ah yeh, i think thats because `postToDTO` is used across multiple controllers

sagzy: Updated in 1ccbda5d738a7f703bfec5688e3eafae95a0aab6

PR #1394, opened 2025-10-21
Added API endpoint for the global feed

sagzy:
ref https://linear.app/ghost/issue/BER-2908

- the global feed currently returns all long-form posts (Articles) from all users

### DIFF
--- a/features/global-feed.feature
+++ b/features/global-feed.feature
@@ -1,4 +1,3 @@
-@global-feed
 Feature: Global Feed
 
   Background:
--- a/features/step_definitions/global_feed_steps.js
+++ b/features/step_definitions/global_feed_steps.js
@@ -14,32 +14,6 @@ Then(
     },
 );
 
-Then(
-    'the article {string} is not in our global feed',
-    async function (articleName) {
-        const article = this.objects[articleName];
-
-        try {
-            await waitForAPObjectInGlobalFeed(article.id);
-            assert.fail(
-                `Expected article ${article.id} to be not be found in the global feed`,
-            );
-        } catch (error) {
-            assert.equal(
-                error.message,
-                `Max retries reached when waiting on item ${article.id} in the global feed`,
-            );
-        }
-    },
-);
-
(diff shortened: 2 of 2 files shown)


src/http/api/feed.controller.ts
```diff
@@ -98,6 +98,53 @@ export class FeedController {
...
+
+        return new Response(
+            JSON.stringify({
+                posts,
+                next: nextCursor,
+            }),
+            {
+                status: 200,
+            },
+        );
```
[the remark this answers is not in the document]
sagzy: This is intentional

src/feed/feed.service.ts
```diff
@@ -121,6 +121,24 @@ export class FeedService {
...
+            return null;
+        }
+
+        const { account_id: accountId } = await this.db('users')
+            .where('id', globalFeedUserId)
+            .select('account_id')
+            .first();
+
+        return accountId;
+    }
```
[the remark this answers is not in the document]
sagzy: Thanks! Fixed

features/step_definitions/global_feed_steps.js
```diff
@@ -0,0 +1,57 @@
...
+        }
+    },
+);
+
+Then('the note {string} is in our global feed', async function (noteName) {
+    const note = this.objects[noteName];
+
+    const found = await waitForAPObjectInGlobalFeed(note.id);
+    assert(found);
+});
```
mike182uk: I don't think this is used?

features/global-feed.feature
```diff
@@ -0,0 +1,15 @@
+@global-feed
```
mike182uk: This doesn't look to be used for anything?

features/step_definitions/global_feed_steps.js
sagzy: Good catch, leftover indeed!

features/global-feed.feature
sagzy: Good catch, leftover indeed!

PR #1500, opened 2026-01-12
Removed unnecessary database queries in dispatchers

sagzy:
ref https://linear.app/ghost/issue/BER-3012

- in [dispatchers.ts](https://github.com/TryGhost/ActivityPub/blob/main/src/dispatchers.ts), we currently make 3 database query to load the host site, account and user
- however, in most cases, this host data already exists in the Hono request context, via the host-data-context middleware: https://github.com/TryGhost/ActivityPub/blob/main/src/http/middleware/host-data-context.ts
- with this change, we pass the host data to the Fedify context instead of making additional database queries in dispatchers
- we have also added the host data database query to `ensureCorrectContext` to make sure the host data is available to all dispatchers code
- the one exception is `keypairDispatcher`, for which we still do an additional database query. The reason is that we don't want to expose the public/private key pair to the Account entity, as there is a risk we might inadvertently expose the private key


src/dispatchers.ts
```diff
@@ -41,42 +40,36 @@ import { lookupActor, lookupObject } from '@/lookup-helpers';
...
-            outbox: new URL(account.ap_outbox_url),
-            following: new URL(account.ap_following_url),
-            followers: new URL(account.ap_followers_url),
-            liked: new URL(account.ap_liked_url),
-            url: new URL(account.url || account.ap_id),
+            inbox: new URL(hostAccount.apInbox ?? ''),
+            outbox: new URL(hostAccount.apOutbox ?? ''),
+            following: new URL(hostAccount.apFollowing ?? ''),
+            followers: new URL(hostAccount.apFollowers ?? ''),
+            liked: new URL(hostAccount.apLiked ?? ''),
```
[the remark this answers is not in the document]
sagzy: fixed

src/account/account.repository.knex.ts
```diff
@@ -372,6 +373,33 @@ export class KnexAccountRepository {
         return this.mapRowToAccountEntity(accountRow);
     }
 
+    async getKeyPair(
```
mike182uk, line 375: I think we should push the "validation" / business logic to the service method and keep the repository focused on data access:

- Repo returns `{ publicKey: string | null, privateKey: string | null } | null`
- `null` if account doesn't exist, nullable fields if keys aren't present

Then service layer deals with wrapping this in the result type and applies the "must be internal account" check

What do you think?

src/account/account.service.ts
```diff
@@ -731,6 +731,22 @@ export class AccountService {
         return await this.accountRepository.getById(id);
     }
 
+    /**
+     * Get the public/private key pair for an account
+     *
+     * @param accountId id of the account
+     */
```
mike182uk: I don't think there is a need for this, code is self documenting?

src/dispatchers.ts
```diff
@@ -41,42 +40,40 @@ import { lookupActor, lookupObject } from '@/lookup-helpers';
...
-            url: new URL(account.url || account.ap_id),
+            inbox: hostAccount.apInbox ? new URL(hostAccount.apInbox) : null,
+            outbox: hostAccount.apOutbox ? new URL(hostAccount.apOutbox) : null,
+            following: hostAccount.apFollowing
+                ? new URL(hostAccount.apFollowing)
+                : null,
+            followers: hostAccount.apFollowers
+                ? new URL(hostAccount.apFollowers)
+                : null,
+            liked: hostAccount.apLiked ? new URL(hostAccount.apLiked) : null,
```
mike182uk: huh, noticed we do this elsewhere, but  aren't these properties already URLs in the entity? (this still works fine as will be cast to a string first)

src/app.ts
```diff
@@ -248,6 +253,18 @@ function ensureCorrectContext<B, R>(
...
+            const hostDataLoader = container.resolve<HostDataContextLoader>(
+                'hostDataContextLoader',
+            );
+            const result = await hostDataLoader.loadDataForHost(ctx.host);
+
+            if (!isError(result)) {
+                const { site, account } = getValue(result);
+                ctx.data.hostSite = site;
+                ctx.data.hostAccount = account;
+            }
```
mike182uk, line 283: Should we log the error just to help out with debugging if we need to?

src/app.ts
```diff
@@ -291,7 +312,7 @@ inboxListener
     .on(
         Follow,
         ensureCorrectContext(
-            spanWrapper((ctx: Context<ContextData>, activity: Follow) => {
+            spanWrapper((ctx: Context<FedifyContextData>, activity: Follow) => {
```
mike182uk: eek, wouldn't it have been better to use the type aliases we already had? (`FedifyContext`)

sagzy: Yeh you're right, more concise to use the alias! I've reverted the renaming in this PR, will do that separately in a follow-up tidying PR

src/account/account.repository.knex.ts
sagzy: Good one - I went with putting everything in one place as it was a pretty minimal method, but you're right that we should keep repositories focused on data retrieval/update. 

Updated in 1782397c34bc29e1dbdb19c493d74c3f98f808bb

src/app.ts
sagzy: Good call - added more logs to the error paths in 05fd29e0d3e58ed10b399321c40273f6d2f084d6

src/account/account.service.ts
sagzy: Removed

src/dispatchers.ts
sagzy: Good catch, unnecessary indeed - removed in 5ceda141663f7309d3785c0c071cbd2a00acb080

src/app.ts
```diff
@@ -249,6 +254,39 @@ function ensureCorrectContext<B, R>(
             ctx.data.logger = globalLogging;
         }
 
+        const hostDataLoader = container.resolve<HostDataContextLoader>(
```
sagzy, line 257: @mike182uk calling out one change since last time you looked here:

I've removed `if (!ctx.data.hostSite || !ctx.data.hostAccount) {...}` wrapper here for two reasons:
1. `ensureCorrectContext` is only called for dispatchers that don't use the request context, so we are not making a duplicate database query
2. [Cucumber test](https://github.com/TryGhost/ActivityPub/blob/5fc8d9a270be38fa006531eb5ea6626e3363e768/features/handle-group-announce-create.feature#L10) was failing, with ctx.site / ctx.account set to another account than the one expected. I first thought that it might be a test setup issue, but I couldn't validate that there wasn't a bug locally as we have a single site. If we end up caching the wrong site data, that's not great - so I preferred to play it safe and have the database query run here systematically

sagzy:
Had a chat with @mike182uk - we both don't feel super confident about this changeset as it touches the Fedify Context. We're going to trade a bit of performance gains for more safety and make the change in the dispatcher code directly. 

Closing this PR and will open another one

PR #1630, opened 2026-03-03
Cached `importJwk` and data host data to improve cpu usage

rmgpinto:
ref https://linear.app/ghost/issue/BER-3241

### DIFF
--- a/src/dispatchers.ts
+++ b/src/dispatchers.ts
@@ -104,8 +104,22 @@ export const actorDispatcher = (hostDataContextLoader: HostDataContextLoader) =>
 export const keypairDispatcher = (
     accountService: AccountService,
     hostDataContextLoader: HostDataContextLoader,
-) =>
-    async function keypairDispatcher(ctx: FedifyContext, identifier: string) {
+) => {
+    const MAX_CACHE_SIZE = 1000;
+    const KEY_TTL_MS = 30 * 60 * 1000; // 30 minutes
+    const cryptoKeyCache = new Map<
+        number,
+        {
+            publicKey: CryptoKey;
+            privateKey: CryptoKey;
+            createdAt: number;
+        }
+    >();
+
+    return async function keypairDispatcher(
+        ctx: FedifyContext,
+        identifier: string,
+    ) {
         const hostData = await hostDataContextLoader.loadDataForHost(ctx.host);
 
         if (isError(hostData)) {
@@ -145,6 +159,14 @@ export const keypairDispatcher = (
 
         const { account } = getValue(hostData);
 
+        const cached = cryptoKeyCache.get(account.id);
+        if (cached) {
+            if (Date.now() - cached.createdAt < KEY_TTL_MS) {
+                return [cached];
+            }
+            cryptoKeyCache.delete(account.id);
+        }
+
         const keyPair = await accountService.getKeyPair(account.id);
 
         if (isError(keyPair)) {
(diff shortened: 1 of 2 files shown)


ErisDS:
### 🤖 Velo CI Failure Analysis
**Classification:** 🟠 SOFT FAIL
- Workflow: `CICD`
- Failed Step: `Run E2E Tests`
- Run: [View failed run](https://github.com/TryGhost/ActivityPub/actions/runs/22615371888)
**What failed:** Test assertion failure in follow_steps.js
**Why:** The root cause of the failure is a test assertion failure in the follow_steps.js file. This is a code issue, as the test is failing due to an issue with the application code, not an infrastructure problem.
**Action:**
The developer should review the follow_steps.js file, identify the failing assertion, and fix the underlying code issue causing the test to fail.

src/http/host-data-context-loader.ts
```diff
@@ -3,23 +3,60 @@ import type { Knex } from 'knex';
...
+        if (cached && cached.expiry > now) {
+            return cached.result;
+        }
+
+        const result = await this.queryDataForHost(host);
+
+        if (!isError(result)) {
+            if (this.cache.size >= HostDataContextLoader.MAX_CACHE_SIZE) {
+                this.cache.delete(this.cache.keys().next().value!);
+            }
```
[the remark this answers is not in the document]
sagzy: suggestion: @rmgpinto think this is good feedback, I'd also suggest deleting the expired key as soon as we know it has expired (e.g. on line 42), so that we don't evict more keys than necessary

src/dispatchers.ts
```diff
@@ -104,8 +104,22 @@ export const actorDispatcher = (hostDataContextLoader: HostDataContextLoader) =>
 export const keypairDispatcher = (
     accountService: AccountService,
     hostDataContextLoader: HostDataContextLoader,
-) =>
-    async function keypairDispatcher(ctx: FedifyContext, identifier: string) {
+) => {
+    const MAX_CACHE_SIZE = 1000;
+    const KEY_TTL_MS = 30 * 60 * 1000; // 30 minutes
```
sagzy, line 109: question: What's the reasoning behind 30 min? It that an arbitrary limit?

src/http/host-data-context-loader.ts
```diff
@@ -3,23 +3,60 @@ import type { Knex } from 'knex';
...
+
+            this.cache.set(host, {
+                result,
+                expiry: now + HostDataContextLoader.CACHE_TTL_MS,
+            });
+        }
+
+        return result;
+    }
```
sagzy, line 57: Think we should cover the cache read/delete/set behaviour in automated tests, e.g. first call makes the database query and set cache, second one reads from cache, one after x min deletes key and re-fetches

sagzy (commented):
Generally adding a caching layer for these repeated queries makes sense, but I do have a few concerns around stale cache entries:
- What happens if the key pair is rotated (e.g. Ghost site is migrated to a new domain)? Will outbound ActivityPub signatures fail for x minutes?
- What happens when a site is deleted (e.g. Ghost user disables Network in Admin)? Do we continue serving stale site/account information for x minutes?

We can lower the risk by making the TTL smaller, but then the caching benefits also become smaller.

I'm curious what @mike182uk thinks too!

sagzy:
Closing, as we're intending to merge in the immediate future

PR #1645, opened 2026-03-09
Added timeout to queue outbound requests to minimize in-flight requests

rmgpinto:
ref https://linear.app/ghost/issue/BER-3241

- Adds a 10s timeout to all outbound requests. The timeout is more than enough for servers to process the request and triggers a fail-fast mechanism that frees up resources.

### DIFF
--- a/src/mq/gcloud-pubsub-push/error-utils.ts
+++ b/src/mq/gcloud-pubsub-push/error-utils.ts
@@ -103,7 +103,8 @@ function isNetworkConnectivityError(error: Error, depth = 0): boolean {
         error.message.match(/(read|connect) ECONNRESET/i) !== null ||
         error.message.match(/socket hang up/i) !== null ||
         error.message.match(/Connect Timeout Error/i) !== null ||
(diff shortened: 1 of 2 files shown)


mike182uk:
@rmgpinto Instead of monkey patching `fetch`, would this achieve what you are trying to do: https://nodejs.org/docs/latest-v22.x/api/globals.html#custom-dispatcher

Not sure on specific usage, but agent is suggesting something like:

```
import { Agent, setGlobalDispatcher } from 'undici';

setGlobalDispatcher(new Agent({
    connectTimeout: 5_000,
    headersTimeout: 10_000,
    bodyTimeout: 10_000,
}));
```

This might also require some tweaking in the error utils to catch whatever errors are thrown by `undici` (e.g `Connect Timeout Error`?)

Ideally we would be able to pass a customised version of `fetch` to `Fedify` but im not seeing an option for that

sagzy:
Closing, as we're not planning on merging this in the immediate future

PR #1673, opened 2026-03-19
Added ADR-0013 to document event driven federation

mike182uk:
ref https://linear.app/ghost/issue/BER-2715

Added ADR-0013 to document event driven federation which is a prerequisite to resolving https://linear.app/ghost/issue/BER-2715


sagzy:
@mike182uk apologies, I might have missed the ping on that ADR PR review. I just had a read now, looks good to me 👍 Going to merge, and we can make amends as needed

PR #1836, opened 2026-05-26
Removed `ghost_uuid` ownership verification on site creation

mike182uk:
The verification added in https://github.com/TryGhost/ActivityPub/pull/1661 fetched the previous owner's admin API to check whether it still claimed the UUID before allowing reassignment. In practice this blocks legitimate domain changes: when the old hostname is decommissioned (DNS NXDOMAIN, connection refused, etc.), `isHostReachableError` treats the failure as unverifiable and refuses the registration, which contradicts the PR's own stated intent of allowing reassignment when the old host is unreachable.

Revert to the pre-https://github.com/TryGhost/ActivityPub/pull/1661 behaviour from https://github.com/TryGhost/ActivityPub/pull/1441: on conflict, nullify the previous row's `ghost_uuid` in the same transaction and insert the new row. No HTTP call to the old host.

### DIFF
--- a/src/site/site.service.integration.test.ts
+++ b/src/site/site.service.integration.test.ts
@@ -210,32 +210,15 @@ describe('SiteService', () => {
         expect(siteA.host).toBe('domain-a.tld');
         expect(siteA.ghost_uuid).toBe(ghostUUID);
 
-        // When domain-b.tld initialises with the same ghost_uuid,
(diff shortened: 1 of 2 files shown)


mike182uk:
Superseded by https://github.com/TryGhost/ActivityPub/pull/1837

PR #1897, opened 2026-06-17
Fixed malformed headers emitting junk trace IDs

miloquinn:
- Validate the W3C `traceparent` header before returning Cloud Logging trace context, so malformed or spoofed headers no longer emit junk trace IDs.
- Parsing is delegated to `parseTraceParent` from `@opentelemetry/core` (already a direct dependency) rather than a hand-rolled parser. This enforces the spec rules — hex format, reserved version `ff`, all-zero trace/span IDs, no extra fields on version `00` — and correctly parses future-version headers that carry extra fields (the spec says receivers should read the first four fields).
- `getTraceContext` is now a thin mapping: null-object shape on missing/invalid input, and `traceFlags & TraceFlags.SAMPLED` → `sampled`.
- Add unit coverage: valid sampled/unsampled headers, a future-version forward-compat case, and a matrix of malformed headers (bad hex, wrong lengths, all-zero IDs, reserved version `ff`, wrong field count).

### DIFF
--- a/src/helpers/context-header.ts
+++ b/src/helpers/context-header.ts
@@ -9,7 +9,20 @@ export function getTraceContext(traceContext: string | undefined) {
         return { traceId: null, spanId: null, sampled: null };
     }
 
-    const [_version, traceId, spanId, flags] = parts;
+    const [version, traceId, spanId, flags] = parts;
+
+    const isValidTraceContext =
+        /^[0-9a-f]{2}$/.test(version) &&
+        /^[0-9a-f]{32}$/.test(traceId) &&
+        traceId !== '00000000000000000000000000000000' &&
+        /^[0-9a-f]{16}$/.test(spanId) &&
+        spanId !== '0000000000000000' &&
+        /^[0-9a-f]{2}$/.test(flags);
+
+    if (!isValidTraceContext) {
+        return { traceId: null, spanId: null, sampled: null };
+    }
+
     const sampled = (Number.parseInt(flags, 16) & 0x1) === 1;
     return { traceId, spanId, sampled };
 }
--- a/src/helpers/context-header.ts
+++ b/src/helpers/context-header.ts
@@ -1,32 +1,23 @@
+import { TraceFlags } from '@opentelemetry/api';
+import { parseTraceParent } from '@opentelemetry/core';
+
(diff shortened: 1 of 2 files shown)


src/helpers/context-header.ts
```diff
@@ -9,7 +9,20 @@ export function getTraceContext(traceContext: string | undefined) {
...
 
-    const [_version, traceId, spanId, flags] = parts;
+    const [version, traceId, spanId, flags] = parts;
+
+    const isValidTraceContext =
+        /^[0-9a-f]{2}$/.test(version) &&
+        /^[0-9a-f]{32}$/.test(traceId) &&
+        traceId !== '00000000000000000000000000000000' &&
+        /^[0-9a-f]{16}$/.test(spanId) &&
+        spanId !== '0000000000000000' &&
```
sagzy: nit: the 32/16 zeros are hard to read, should we simplify with

```
!/^0+$/.test(traceId) &&
!/^0+$/.test(spanId)
```

as the preceding regex already checks for hex of the right length?

src/helpers/context-header.ts
```diff
@@ -9,7 +9,20 @@ export function getTraceContext(traceContext: string | undefined) {
...
+        /^[0-9a-f]{2}$/.test(flags);
```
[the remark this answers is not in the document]
sagzy: This comment from Rabbit looks valid / worth addressing

src/helpers/context-header.ts
```diff
@@ -9,7 +9,20 @@ export function getTraceContext(traceContext: string | undefined) {
         return { traceId: null, spanId: null, sampled: null };
     }
 
-    const [_version, traceId, spanId, flags] = parts;
+    const [version, traceId, spanId, flags] = parts;
+
+    const isValidTraceContext =
```
sagzy: nit: would be worth adding an inline comment here to explain why we're doing this

e.g. "Reject malformed/spoofed traceparent so we don't emit junk trace IDs to Cloud Logging"

sagzy:
Hey @miloquinn, thanks for your contribution 🙏 

I've left some minor comments, this looks good overall! Can you ping me once you've gone through addressing those?

PR #2024, opened 2026-07-21
Fixed WebFinger 404s caused by stale www/non-www sibling sites

mike182uk:
ref https://linear.app/ghost/issue/ONC-1905/social-web-not-working

When a publication's domain moves between the www and non-www variant, a new site registration is created and the old one is left behind. WebFinger resolution picked a single site row - the bare host if present, otherwise the www variant - so a stale bare-host row would answer for the whole domain: queries for the live account's username returned 404 while the stale row's old username kept resolving. The failure surfaces when a customer changes their handle

Resolution now checks both host variants and returns the first account that answers to the requested username:

| Case                         | Before        | After         |
| ---------------------------- | ------------- | ------------- |
| Single site row, match       | 200           | 200           |
| Single site row, no match    | 404           | 404           |
| Both rows, bare host matches | 200 from bare | 200 from bare |
| Both rows, only www matches  | 404           | 200 from www  |
| Both rows, neither matches   | 404           | 404           |
| No site rows                 | fall through  | fall through  |

Only the "only www matches" case changes, so the fix can only turn 404s into matches - no existing 200 response changes shape. The preserved fall-through for unknown domains is what the custom domain validation flow depends on. The same handling applies to the request-host branch used during that validation

### DIFF
--- a/src/http/api/webfinger.controller.ts
+++ b/src/http/api/webfinger.controller.ts
@@ -64,25 +64,19 @@ export class WebFingerController {
             return this.createWebfingerResponse(customDomainAccount);
         }
 
-        const site =
-            (await this.siteService.getSiteByHost(normalizedResourceHost)) ||
-            (await this.siteService.getSiteByHost(
-                `www.${normalizedResourceHost}`,
-            ));
+        const resourceLookup = await this.findAccountForHandleHost(
+            resourceUsername,
+            normalizedResourceHost,
+        );
 
-        if (site) {
-            const account = await this.getAccountBySiteOrNull(site);
+        if (resourceLookup.account) {
+            return this.createWebfingerResponse(resourceLookup.account);
+        }
 
-            if (
-                !account ||
-                !this.resourceUsernameMatchesAccount(resourceUsername, account)
-            ) {
-                return new Response(null, {
-                    status: 404,
-                });
-            }
-
-            return this.createWebfingerResponse(account);
+        if (resourceLookup.siteFound) {
+            return new Response(null, {
+                status: 404,
+            });
         }
 
         const requestHost = ctx.req.header('host')?.split(':')[0];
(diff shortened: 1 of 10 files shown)


src/http/api/webfinger.controller.ts
```diff
@@ -124,6 +107,44 @@ export class WebFingerController {
...
+            if (
+                account &&
+                this.resourceUsernameMatchesAccount(resourceUsername, account)
+            ) {
+                return { siteFound, account };
+            }
+        }
+
+        return { siteFound, account: null };
+    }
```
sagzy, line 146: Could we simplify the returned value to `account: Account | null`? I'm not seeing what value we get from returning `siteFound`. After simplifying, the caller logic then becomes:

```
const account = await this.findAccountForHandleHost(...);

if (!account) {
    return new Response(null, {
        status: 404,
    });
}

return this.createWebfingerResponse(
    account,
    normalizedResourceHost,
);
```

src/http/api/webfinger.controller.ts
```diff
@@ -124,6 +107,44 @@ export class WebFingerController {
...
+        for (const host of [handleHost, `www.${handleHost}`]) {
+            const site = await this.siteService.getSiteByHost(host);
+
+            if (!site) {
+                continue;
+            }
+
+            siteFound = true;
+
+            const account = await this.getAccountBySiteOrNull(site);
```
sagzy: Pre-existing, but this method swallows errors with a try/catch, which means that a transient DB error could lead to a different result? 

Example scenario:
- Live bare site (example.com) + stale www site (www.example.com)
- `accountRepository.getBySite` for the bare site throws a transient MySQL connection error, the catch in `getAccountBySiteOrNull` returns null, the loop continues to `www.example.com`
- WebFinger returns 200 with the stale www account

Using `this.accountRepository.getBySite` directly makes things more predictable, so would prefer that personally

src/http/api/webfinger.controller.ts
mike182uk: `siteFound` differentiates "there is a site for this host but no matching account" (404) from "no site for this host, fall through to the custom-domain logic" (the request-host branch). I think if we simplified, a lookup for something like: `acct:alice@custom.example.com` that `validateWebfingerHost` makes would `404` instead of resolving via the request-host (4 tests fail with the simplification). The request-host call site _does_ ignore `siteFound` so the simplification works there, but the helper serves both places its used

src/http/api/webfinger.controller.ts
mike182uk: 🤔 `getBySite` throws a plain error when user / account not found, which means we would need to work out what the plain error was for (user / account not found VS mysql transient error)

mike182uk: @sagzy What do you think of https://github.com/TryGhost/ActivityPub/pull/2024/commits/10445f4ebc30d583123bb3da175c8f47d9f63977

The blast radius is a bit bigger than wanted, but I think the design is better (`getBySite` acts the same way as the other `getBy` methods, controllers now depend on ctx to get the account rather than querying for it which was the original intention behind the ctx. This also saves a query per controller)

sagzy: @mike182uk really nice 💯

src/http/api/webfinger.controller.ts
sagzy: Yes you're right, missed that the handling was different between the two lookups. All good then!

sagzy (approved):

PR #2063, opened 2026-07-30
🎨 Cleaned up activitypub migrations image

acburdine:
ref https://linear.app/ghost/issue/PLA-278/optimize-activitypub-migrations-docker-build
- switch base to alpine
- copy migrate binary from migrate image rather than curl from github

related to #2062 - shrinks the migration image from 228MB => 82.4MB on disk.

### DIFF
--- a/migrate/Dockerfile
+++ b/migrate/Dockerfile
@@ -1,18 +1,13 @@
-FROM docker.io/library/debian:bookworm-slim@sha256:7b140f374b289a7c2befc338f42ebe6441b7ea838a042bbd5acbfca6ec875818
-ARG TARGETARCH
+FROM docker.io/library/alpine:3.24@sha256:28bd5fe8b56d1bd048e5babf5b10710ebe0bae67db86916198a6eec434943f8b
 
-RUN apt-get update -y && \
-    apt-get install -y curl && \
-    rm -rf /var/lib/apt/lists/* && \
-    curl -fsSL -O https://github.com/golang-migrate/migrate/releases/download/v4.17.1/migrate.linux-${TARGETARCH}.tar.gz && \
-    curl -fsSL -O https://github.com/golang-migrate/migrate/releases/download/v4.17.1/sha256sum.txt && \
-    grep " migrate.linux-${TARGETARCH}.tar.gz$" sha256sum.txt | sha256sum -c - && \
-    tar xvzf migrate.linux-${TARGETARCH}.tar.gz migrate && \
-    rm migrate.linux-${TARGETARCH}.tar.gz sha256sum.txt && \
-    mv migrate /usr/bin/
+# Binary is copied from golang-migrate's published image rather than their
+# release tarball so Renovate tracks the digest. Not used as a base: it sets
+# ENTRYPOINT ["migrate"] and runs on a different version of Alpine.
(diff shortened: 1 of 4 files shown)


migrate/Dockerfile
```diff
@@ -1,18 +1,13 @@
-FROM docker.io/library/debian:bookworm-slim@sha256:7b140f374b289a7c2befc338f42ebe6441b7ea838a042bbd5acbfca6ec875818
-ARG TARGETARCH
+FROM docker.io/library/alpine:3.24@sha256:28bd5fe8b56d1bd048e5babf5b10710ebe0bae67db86916198a6eec434943f8b
```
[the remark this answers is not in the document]
acburdine: this image is only set up to run migrations for activitypub, and exits once migrations are complete. It doesn't allow any connectivity in and adding a non-root user would complicate setups where a mysql db socket is bind-mounted into the container, so we're ignoring this for now.

migrate/bin/up
```diff
@@ -1,4 +1,4 @@
-#!/usr/bin/env bash
+#!/bin/sh
 
 if [ -n "$MYSQL_DB" ]; then
     echo "Running migrate -path migrations -database $MYSQL_DB up"
```
sagzy:
```suggestion
    echo "Running migrate -path migrations -database <REDACTED> up"
```

Pre-existing, just noticing now: we're echo-ing the full DNS with the password in it in this line

migrate/bin/create
```diff
@@ -1,4 +1,4 @@
-#!/usr/bin/env bash
+#!/bin/sh
 
 echo "Running migrate create -ext sql -dir migrations -seq $1"
 migrate create -ext sql -dir migrations -seq $1
```
sagzy:
```suggestion
migrate create -ext sql -dir migrations -seq "$1"
```

pre-existing nit: $1 was unquoted and there is a trailing space

migrate/Dockerfile
```diff
@@ -1,18 +1,13 @@
...
-    rm -rf /var/lib/apt/lists/* && \
-    curl -fsSL -O https://github.com/golang-migrate/migrate/releases/download/v4.17.1/migrate.linux-${TARGETARCH}.tar.gz && \
-    curl -fsSL -O https://github.com/golang-migrate/migrate/releases/download/v4.17.1/sha256sum.txt && \
-    grep " migrate.linux-${TARGETARCH}.tar.gz$" sha256sum.txt | sha256sum -c - && \
-    tar xvzf migrate.linux-${TARGETARCH}.tar.gz migrate && \
-    rm migrate.linux-${TARGETARCH}.tar.gz sha256sum.txt && \
-    mv migrate /usr/bin/
+# Binary is copied from golang-migrate's published image rather than their
+# release tarball so Renovate tracks the digest. Not used as a base: it sets
+# ENTRYPOINT ["migrate"] and sits on Alpine 3.19.
```
sagzy: nit: "Alpine 3.19" might silently go stale with the Renovate automerges

sagzy (approved):
nice one @acburdine, left a few nits/suggestions

Also README needs a refresher as it references `bin/bash` in the [Migrations](https://github.com/tryghost/activityPub/#migrations) section

migrate/bin/up
acburdine: fixed 👍🏻

migrate/Dockerfile
acburdine: fixed 👍🏻

migrate/bin/create
acburdine: fixed 👍🏻

PR #2148, opened 2026-09-01
Let Docker majors automerge, except for production runtimes

sagzy:
We automerge major updates everywhere else. npm dependencies and GitHub Actions both come through `github>tryghost/renovate-config`, which turns automerge on for every update type:

```json5
// Automerge all dependency updates (major, minor, and patch)
{ "matchDepTypes": ["dependencies", "devDependencies", …, "action"],
  "automerge": true, "automergeType": "pr" }
```

Docker was the one exception, because `docker:disableMajor` switched its majors off entirely.

That exception just cost us a red `main`. The Pub/Sub emulator sat on `google-cloud-cli` 499.0.0 for 22 months, because every release of a calendar-versioned image reads as a major and Renovate never proposed one. Google eventually pruned the image from gcr.io and CI stopped starting at all.

## What changes

Dropping `docker:disableMajor` brings Docker in line with everything else. Three images get held back for review instead:

| Group | Covers | Held for review |
|---|---|---|
| MySQL | `mysql` image + `mysql2` | Majors |
| Redis | `redis` image + `ioredis` | Majors |
| Node.js | `node` image + `.nvmrc` | Every update |

The split is about how a bad bump shows up:

```mermaid
flowchart TD
    A[Docker major available] --> B{Backs a production service?}
    B -->|"No — wiremock, caddy,<br/>jaeger, Pub/Sub emulator"| C[Automerge]
    B -->|"Yes — MySQL, Redis, Node"| D[Group with its client]
    D --> E[Review by hand]
    C --> F{Bad bump?}
    F -->|CI goes red| G[We find out right away]
    E --> H{Bad bump?}
    H -->|CI stays green| I[Drift is invisible —<br/>this is what review catches]
```

When test scaffolding breaks, CI goes red and we know right away, so those majors are safe to automerge. MySQL, Redis and Node back services whose production versions live outside this repo, so a bump here keeps CI green while quietly testing against a runtime we don't run. That's the failure mode worth a human look, and it's the same reasoning already behind the existing Node rule.

Grouping each server with its client means the pair moves in one PR rather than two that automerge independently. That also covers the near-miss in #2076, where an `ioredis` v6 major was queued to automerge on its own — it could have re-paired `@fedify/redis` with a client outside its peer range, which pnpm only warns about.

## Notes

- Minors and patches are unaffected and still automerge
- Fedify is untouched and still requires review on every update — I confirmed against Renovate's own PR bodies, which report `Automerge: Disabled by config` on #2134 versus `Automerge: Enabled` on a normal dependency PR
- `enabled: true` on the Node rule is gone, since it only existed to override `docker:disableMajor`
- Validated with `renovate-config-validator`, and the resolved rules are unchanged apart from the additions above
- Also rewrote the comments in this file so they explain the reasoning rather than restate the config

### DIFF
--- a/.github/renovate.json5
+++ b/.github/renovate.json5
@@ -2,14 +2,14 @@
   extends: [
     'github>tryghost/renovate-config',
     'docker:pinDigests',
-    'docker:disableMajor',
   ],
-  // Automerge during EU office hours only (Monday - Thursday, 07:00 - 16:59 UTC)
-  // Why? Merging to main ships to production, so we want to avoid automerging when no engineer is online
+  // Only automerge during EU office hours (Monday–Thursday, 07:00–16:59 UTC).
+  // Merging to main ships straight to production, so we don't want anything
+  // landing while there's no engineer around to pick it up.
   //
-  // Notes:
-  // - this schedule doesn't restrict when the PRs are opened by Renovate, only when they are merged
-  // - automerge schedule works only if we're using Renovate automerge over platform automerge (Github)
+  // Two things to know. The schedule controls when PRs merge, not when
+  // Renovate opens them. And it only works while we use Renovate's own
+  // automerge rather than GitHub's, which is what platformAutomerge turns off.
   platformAutomerge: false,
   automergeSchedule: [
     '* 7-16 * * 1,2,3,4',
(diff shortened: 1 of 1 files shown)


sagzy:
The immediate CI fix is in #2149 — that one unblocks `main`, this one stops the pin going stale again.

rmgpinto (approved):

Issue #2118, opened 2026-08-20
Custom handle changes don't propagate to other Ghost instances


betschki:
## Summary

When an account sets a custom handle, that handle appears to be visible only on the instance that hosts it. Other Ghost instances keep displaying the handle derived from the actor's hostname and never pick up the change — not on first contact, and not after the remote actor sends an `Update`.

This isn't visible on Ghost Pro, where the accounts involved are all on the same instance. It shows up as soon as two Ghost instances are involved.

## Reproduction

`@john@john.onolan.org` changed his handle to `@john@onolan.org`.

The actor document is unchanged by the switch — same `id`, same `preferredUsername`, no `movedTo`:

```bash
curl -sL -H 'Accept: application/activity+json' \
  https://john.onolan.org/.ghost/activitypub/users/index
# id                = https://john.onolan.org/.ghost/activitypub/users/index
# preferredUsername = john
```

The new handle exists only in the WebFinger `subject`:

```bash
curl -sL 'https://john.onolan.org/.well-known/webfinger?resource=acct:john@john.onolan.org'
# {"subject":"acct:john@onolan.org", ...}
```

**Expected:** a remote Ghost instance displays `@john@onolan.org`.
**Actual:** it displays `@john@john.onolan.org`, indefinitely.

## Other implementations handle this

Three unrelated Mastodon instances resolved the change correctly, and dropped the old form entirely:

```
mastodon.social   acct=john@onolan.org      john@john.onolan.org -> Record not found
mstdn.social      acct=john@onolan.org      john@john.onolan.org -> Record not found
fosstodon.org     acct=john@onolan.org      john@john.onolan.org -> Record not found
```

```bash
curl -s 'https://mastodon.social/api/v1/accounts/lookup?acct=john@onolan.org'
```

So the information is on the wire in a standard field and is discoverable by any server that reads it. It just isn't pushed — there's no activity to wait for.

## Environment

- `ghcr.io/tryghost/activitypub:edge`, image `sha-eeacfc7` (`eeacfc75740c803fdfdc94a9f7c3477dc28fbb6b`, built 2026-08-19)
- Self-hosted, MySQL, Pub/Sub MQ
- Also checked against `main` at `eb5b7fe`; the commits between the two are dependency bumps only

PR #2163, opened 2026-09-03
Fixed remote custom handle discovery via WebFinger subject

louisghost:
## Summary

Remote Ghost instances ignored alternate WebFinger domains and rendered handles from the actor host, so a publisher on a custom domain showed up elsewhere as `@john@john.onolan.org` instead of `@john@onolan.org` (#2118). This resolves the WebFinger `subject` host when an external account is ingested or refreshed, stores it on the row, and reads the stored value on every display surface — feeds, posts, profiles and follower lists.

**A claimed handle is verified before it is stored.** Reading the subject from the actor's own server proves nothing about the domain it names, so any instance could serve an actor at `evil.example`, answer its WebFinger with `acct:index@victim-site.com`, and render as a real Ghost site in feeds, notifications and search on every receiving instance. A single Follow or Like is enough to get the row written. A differing domain is now only accepted once that domain's own WebFinger independently points back at the same actor, which is what Mastodon requires before trusting a subject from another host. For #2118 the second lookup at `onolan.org` redirects to `john.onolan.org` and returns the matching self link, so it passes; the spoofing case fails because `victim-site.com` answers with its own actor.

The subject's local-part has to match the actor's username too. Handles are rendered from `preferredUsername`, so an actor named `alice` advertising `acct:bob@custom.example` — with `custom.example` genuinely confirming `bob` — would otherwise be displayed as `@alice@custom.example`, a handle nobody verified, which on a shared custom-handle host can belong to somebody else and takes their slot in the `UNIQUE (username, webfinger_host_hash)` index. Both lookups carry a 5s timeout; without one, a remote that accepts the connection and never sends headers stalls the caller for undici's default header timeout.

**Resolution happens on write paths only.** An earlier revision of this PR backfilled from `getByApId` / `ensureByApId`. Those run for every inbound Create, Like, Announce and Follow, so that cost an outbound request per known actor, repeated indefinitely for any actor whose lookup never succeeded. Resolution now happens when an external account is created, and on actor `Update` — which is what Ghost federates when a publisher sets a custom domain. A known account is one indexed read again. The refresh writes only the `webfinger_host` column rather than saving the whole row, so a slow lookup can no longer overwrite profile fields another instance updated meanwhile.

Populating this column reached further than display, so three things came with it:

- `getByWebfingerHandle` had no internal-only filter, and the column had only ever been set on accounts we host. Writing it to external rows made our own `/.well-known/webfinger` answer for accounts hosted elsewhere, and made `createInternalAccount` reject a publisher moving to Ghost(Pro) on a domain some remote row already held. That lookup now joins `users`, so it only resolves accounts we host.
- `NULL` already means "use the actor host" to every reader of the column. Writing the actor host back into it as a "we checked this" marker would give the column two meanings and put every external account into the `UNIQUE (username, webfinger_host_hash)` index, so only verified custom hosts are stored.
- Domain blocks are recorded against whichever host the reader was shown, so a block taken from a surface displaying a custom handle has to match an account under either host. Matching `domain_hash` alone meant a block could hide an account from Explore while its posts still arrived in the feed and the follow survived — worse than not blocking. Every comparison now goes through `src/moderation/domain-blocks.ts`: feed fan-out, `canInteractWithAccount`, Explore, Recommendations, Blocks, search, profile `domainBlockedByMe`, the follower and following lists, and the block-time teardown of follows, feed entries and notifications. Blocking by the displayed handle domain keeps working; that arrived with #1821 and is covered by tests.

Two Blocks-list correctness gaps showed up once that either-host matching landed:

- `getBlockedAccounts` used a join with no `blocker_id` scope (pre-existing on `main`) and then, after matching both hashes, could return **two rows for one account** when the viewer had blocked both of its domains. It now asks via an `EXISTS` subquery scoped to the viewer, so the answer is a boolean and the list stays one entry per blocked account.
- Unknown actors on the follower/following fallback path can redirect to a canonical id on another host; `domainBlockedByMe` now checks both the listed URL host and the resolved actor host, consistent with the rendered handle.

Known limitation: remote accounts that set a custom domain before the sending instance runs this code keep `webfinger_host = NULL` until their next profile edit, which then federates an `Update` and corrects them. There is no backfill migration, so #2118 stays visibly unfixed for already-known follows until then — worth a conscious call before this ships.

## Test plan

- [ ] Unit and integration: reciprocal verification (accepts a confirming domain, rejects a spoofed one, rejects a mismatched local-part, tolerates timeouts), internal-only `getByWebfingerHandle`, domain blocks matching either host on both the feed and Explore, stored-host display on profile views, Blocks list stays one entry when both hosts are blocked and ignores another viewer's domain block
- [ ] Self-hosted: receive activity from an account with a custom WebFinger domain and confirm the feed, its profile, and follower lists all show the subject host rather than the actor host
- [ ] Self-hosted: serve an actor whose WebFinger subject claims a domain it does not control, and confirm the handle falls back to the actor host
- [ ] Block a remote account's displayed handle domain and confirm its posts stop arriving in the feed, the follow is torn down, and it disappears from Explore and search
- [ ] Block both the actor host and the custom handle for the same account and confirm the blocked-accounts list shows it once
- [ ] Change a publisher's custom domain and confirm the federated actor `Update` refreshes the stored host on the receiving instance
- [ ] Accounts without a custom domain still render `@user@actor-host` as before

### DIFF
--- a/src/account/account.service.ts
+++ b/src/account/account.service.ts
@@ -37,6 +37,7 @@ import { isHandle } from '@/helpers/activitypub/actor';
 import {
     lookupObject as lookupActivityPubObject,
     lookupActorProfile,
+    resolveExternalWebfingerHost,
 } from '@/lookup-helpers';
 
 interface GetFollowingAccountsOptions {
@@ -110,7 +111,7 @@ export class AccountService {
     async getByApId(id: URL): Promise<Account | null> {
         const account = await this.accountRepository.getByApId(id);
         if (account) {
-            return account;
+            return this.refreshExternalWebfingerHostIfMissing(account);
         }
 
         const context = this.fedifyContextFactory.getFedifyContext();
@@ -133,6 +134,10 @@ export class AccountService {
         }
 
         const data = await mapActorToExternalAccountData(potentialActor);
+        data.webfinger_host = await this.resolveWebfingerHostForExternal(
+            data.username,
+            new URL(data.ap_id),
+        );
 
         await this.createExternalAccount(data);
 
@@ -144,7 +149,9 @@ export class AccountService {
     ): Promise<Result<Account, RemoteAccountFetchError>> {
         const account = await this.accountRepository.getByApId(id);
         if (account) {
-            return ok(account);
+            return ok(
+                await this.refreshExternalWebfingerHostIfMissing(account),
+            );
         }
 
         const context = this.fedifyContextFactory.getFedifyContext();
@@ -188,6 +195,11 @@ export class AccountService {
         } catch (_err) {
             return error('invalid-data');
         }
+
+        data.webfinger_host = await this.resolveWebfingerHostForExternal(
+            data.username,
+            new URL(data.ap_id),
+        );
 
         await this.createExternalAccount(data);
 
@@ -423,6 +435,89 @@ export class AccountService {
     }
 
     /**
+     * Resolve the WebFinger handle host for a remote actor.
+     *
+     * Returns the custom subject host when present. When WebFinger confirms the
+     * actor host is canonical, returns that host so callers can persist it and
+     * skip re-resolving on later ensureByApId calls. Returns null only when
+     * lookup fails (caller should leave any existing value unchanged).
+     */
+    private async resolveWebfingerHostForExternal(
+        username: string,
+        apId: URL,
+    ): Promise<string | null> {
+        const resolution = await resolveExternalWebfingerHost(username, apId);
+        if (resolution.type === 'custom') {
+            return resolution.host;
+        }
+        if (resolution.type === 'default') {
+            return normalizeWebfingerHost(apId.host);
+        }
+        return null;
+    }
+
+    /**
+     * One-shot backfill for external accounts that predate remote WebFinger
+     * persistence. Once a host is stored (custom or confirmed default), further
+     * ensureByApId calls skip the lookup. custom→custom changes still arrive via
+     * actor Update → {@link refreshExternalWebfingerHost}.
+     */
+    private async refreshExternalWebfingerHostIfMissing(
+        account: Account,
+    ): Promise<Account> {
+        if (account.isInternal || account.webfingerHost !== null) {
+            return account;
+        }
+
+        return this.refreshExternalWebfingerHost(account);
+    }
+
+    /**
+     * Re-resolve and persist the WebFinger subject host for an external account.
+     * Lookup failures leave the stored value unchanged.
+     */
+    async refreshExternalWebfingerHost(account: Account): Promise<Account> {
+        if (account.isInternal) {
+            return account;
+        }
+
+        const nextHost = await this.resolveWebfingerHostForExternal(
+            account.username,
+            account.apId,
+        );
+
+        if (nextHost === null || nextHost === account.webfingerHost) {
+            return account;
+        }
+
+        const actorHost = normalizeWebfingerHost(account.apId.host);
+        if (nextHost !== actorHost) {
+            const conflict =
+                await this.accountRepository.hasWebfingerHandleConflict(
+                    account.username,
+                    nextHost,
+                    account.id,
+                );
+
+            if (conflict) {
+                return account;
+            }
+        }
+
+        const updated = account.setWebfingerHost(nextHost);
+        try {
+            await this.accountRepository.save(updated);
+        } catch (err) {
+            if (isDuplicateEntryError(err)) {
+                return account;
+            }
+            throw err;
+        }
+
+        return updated;
+    }
+
+    /**
      * Create an internal account
      *
      * An internal account is an account that is linked to a user
@@ -562,7 +657,7 @@ export class AccountService {
             uuid: randomUUID(),
             ap_private_key: null,
             domain: new URL(accountData.ap_id).host,
-            webfinger_host: null,
+            webfinger_host: accountData.webfinger_host ?? null,
         };
 
         try {
@@ -579,10 +674,20 @@ export class AccountService {
                     ])
                     .first<AccountType>();
 
-                if (!existingAccount) {
-                    throw error;
+                if (existingAccount) {
+                    return existingAccount;
                 }
-                return existingAccount;
+
+                // Likely a webfinger_host uniqueness conflict — retry without
+                // the custom host rather than failing account creation.
+                if (dataToInsert.webfinger_host) {
+                    return this.createExternalAccount({
+                        ...accountData,
+                        webfinger_host: null,
+                    });
+                }
+
+                throw error;
             }
             throw error;
         }
@@ -930,6 +1035,10 @@ export class AccountService {
         const updated = account.updateProfile(profileData);
 
         await this.accountRepository.save(updated);
+
+        if (!account.isInternal) {
+            await this.refreshExternalWebfingerHost(updated);
+        }
 
         return ok(true);
     }
--- a/src/account/account.service.ts
+++ b/src/account/account.service.ts
@@ -533,7 +533,7 @@ export class AccountService {
 
             if (conflict) {
                 getLogger(['activitypub']).warn(
-                    'WebFinger handle @{username}@{host} for {apId} is already held by another account, leaving host unchanged',
+                    'WebFinger handle @{username}@{host} for {apId} is already held by another account, falling back to actor host',
                     {
                         username: account.username,
                         host: nextHost,
@@ -541,10 +541,9 @@ export class AccountService {
                     },
                 );
 
-                // After a rename the previous host was verified for a different
-                // local-part. Refuse to keep it attached to the new username
-                // when the newly resolved host cannot be claimed.
-                if (options.usernameChanged && account.webfingerHost !== null) {
+                // WebFinger no longer advertises the stored host (or never did
+                // for this local-part). Keeping it would show a stale handle.
+                if (account.webfingerHost !== null) {
                     await this.accountRepository.updateWebfingerHost(
                         account.id,
                         null,
@@ -562,6 +561,22 @@ export class AccountService {
             );
         } catch (err) {
             if (isDuplicateEntryError(err)) {
+                getLogger(['activitypub']).warn(
+                    'WebFinger handle @{username}@{host} for {apId} conflicted on write, falling back to actor host',
+                    {
+                        username: account.username,
+                        host: nextHost,
+                        apId: account.apId.href,
+                    },
+                );
+
+                if (account.webfingerHost !== null && nextHost !== null) {
+                    await this.accountRepository.updateWebfingerHost(
+                        account.id,
+                        null,
+                    );
+                }
+
                 return;
             }
             throw err;
--- a/src/account/account.service.ts
+++ b/src/account/account.service.ts
@@ -393,7 +393,7 @@ export class AccountService {
     ): Promise<Result<Account, WebfingerHostError>> {
         if (host === null) {
             const updated = account.setWebfingerHost(null);
-            await this.accountRepository.save(updated);
+            await this.persistWebfingerHostChange(updated);
             return ok(updated);
         }
 
@@ -406,7 +406,7 @@ export class AccountService {
 
         if (normalizedHost === fallbackHost) {
             const updated = account.setWebfingerHost(null);
-            await this.accountRepository.save(updated);
+            await this.persistWebfingerHostChange(updated);
             return ok(updated);
         }
 
@@ -421,7 +421,7 @@ export class AccountService {
 
         const updated = account.setWebfingerHost(normalizedHost);
         try {
-            await this.accountRepository.save(updated);
+            await this.persistWebfingerHostChange(updated);
         } catch (err) {
             if (isDuplicateEntryError(err)) {
                 return error({ type: 'conflict', host: normalizedHost });
@@ -431,6 +431,20 @@ export class AccountService {
         }
 
         return ok(updated);
+    }
+
+    /**
+     * Persist an intentional WebFinger host change.
+     *
+     * `save()` no longer writes `webfinger_host`, so the column is updated
+     * separately — the same seam external refresh uses.
+     */
+    private async persistWebfingerHostChange(account: Account): Promise<void> {
+        await this.accountRepository.save(account);
+        await this.accountRepository.updateWebfingerHost(
+            account.id,
+            account.webfingerHost,
+        );
     }
 
     /**
--- a/src/account/account.service.ts
+++ b/src/account/account.service.ts
@@ -6,6 +6,7 @@ import {
     isActor,
     lookupObject as lookupFedifyObject,
 } from '@fedify/vocab';
+import { getLogger } from '@logtape/logtape';
 import type { Knex } from 'knex';
 
 import { type Account, AccountEntity } from '@/account/account.entity';
@@ -37,7 +38,7 @@ import { isHandle } from '@/helpers/activitypub/actor';
 import {
     lookupObject as lookupActivityPubObject,
     lookupActorProfile,
-    resolveExternalWebfingerHost,
+    resolveCustomWebfingerHost,
 } from '@/lookup-helpers';
 
 interface GetFollowingAccountsOptions {
@@ -111,7 +112,7 @@ export class AccountService {
     async getByApId(id: URL): Promise<Account | null> {
         const account = await this.accountRepository.getByApId(id);
         if (account) {
-            return this.refreshExternalWebfingerHostIfMissing(account);
+            return account;
         }
 
         const context = this.fedifyContextFactory.getFedifyContext();
@@ -134,7 +135,7 @@ export class AccountService {
         }
 
         const data = await mapActorToExternalAccountData(potentialActor);
-        data.webfinger_host = await this.resolveWebfingerHostForExternal(
+        data.webfinger_host = await this.resolveCustomWebfingerHostForExternal(
             data.username,
             new URL(data.ap_id),
         );
@@ -149,9 +150,7 @@ export class AccountService {
     ): Promise<Result<Account, RemoteAccountFetchError>> {
         const account = await this.accountRepository.getByApId(id);
         if (account) {
-            return ok(
-                await this.refreshExternalWebfingerHostIfMissing(account),
-            );
+            return ok(account);
         }
 
         const context = this.fedifyContextFactory.getFedifyContext();
@@ -196,7 +195,7 @@ export class AccountService {
             return error('invalid-data');
         }
 
-        data.webfinger_host = await this.resolveWebfingerHostForExternal(
+        data.webfinger_host = await this.resolveCustomWebfingerHostForExternal(
             data.username,
             new URL(data.ap_id),
         );
@@ -435,63 +434,69 @@ export class AccountService {
     }
 
     /**
-     * Resolve the WebFinger handle host for a remote actor.
+     * Resolve the custom WebFinger handle host to store for a remote actor.
      *
-     * Returns the custom subject host when present. When WebFinger confirms the
-     * actor host is canonical, returns that host so callers can persist it and
-     * skip re-resolving on later ensureByApId calls. Returns null only when
-     * lookup fails (caller should leave any existing value unchanged).
+     * Only a verified custom host is ever returned. A null `webfinger_host`
+     * already means "use the actor host" for every reader of the column, so the
+     * actor host is never written back into it.
      */
-    private async resolveWebfingerHostForExternal(
+    private async resolveCustomWebfingerHostForExternal(
         username: string,
         apId: URL,
     ): Promise<string | null> {
-        const resolution = await resolveExternalWebfingerHost(username, apId);
-        if (resolution.type === 'custom') {
-            return resolution.host;
+        const resolution = await resolveCustomWebfingerHost(username, apId);
+
+        if (resolution.type !== 'custom') {
+            return null;
         }
-        if (resolution.type === 'default') {
-            return normalizeWebfingerHost(apId.host);
+
+        const conflict =
+            await this.accountRepository.hasWebfingerHandleConflict(
+                username,
+                resolution.host,
+                0,
+            );
+
+        if (conflict) {
+            getLogger(['activitypub']).warn(
+                'WebFinger handle @{username}@{host} for {apId} is already held by another account, storing default host',
+                { username, host: resolution.host, apId: apId.href },
+            );
+
+            return null;
         }
-        return null;
+
+        return resolution.host;
     }
 
     /**
-     * One-shot backfill for external accounts that predate remote WebFinger
-     * persistence. Once a host is stored (custom or confirmed default), further
-     * ensureByApId calls skip the lookup. custom→custom changes still arrive via
-     * actor Update → {@link refreshExternalWebfingerHost}.
+     * Re-resolve and persist the custom WebFinger host for an external account.
+     *
+     * Only the `webfinger_host` column is written, so this cannot clobber
+     * profile fields updated concurrently. A failed lookup leaves the stored
+     * value unchanged; a successful lookup that finds no custom host clears it.
      */
-    private async refreshExternalWebfingerHostIfMissing(
-        account: Account,
-    ): Promise<Account> {
-        if (account.isInternal || account.webfingerHost !== null) {
-            return account;
-        }
-
-        return this.refreshExternalWebfingerHost(account);
-    }
-
-    /**
-     * Re-resolve and persist the WebFinger subject host for an external account.
-     * Lookup failures leave the stored value unchanged.
-     */
-    async refreshExternalWebfingerHost(account: Account): Promise<Account> {
+    async refreshExternalWebfingerHost(account: Account): Promise<void> {
         if (account.isInternal) {
-            return account;
+            return;
         }
 
-        const nextHost = await this.resolveWebfingerHostForExternal(
+        const resolution = await resolveCustomWebfingerHost(
             account.username,
             account.apId,
         );
 
-        if (nextHost === null || nextHost === account.webfingerHost) {
-            return account;
+        if (resolution.type === 'unavailable') {
+            return;
         }
 
-        const actorHost = normalizeWebfingerHost(account.apId.host);
-        if (nextHost !== actorHost) {
+        const nextHost = resolution.type === 'custom' ? resolution.host : null;
+
+        if (nextHost === account.webfingerHost) {
+            return;
+        }
+
+        if (nextHost !== null) {
             const conflict =
                 await this.accountRepository.hasWebfingerHandleConflict(
                     account.username,
@@ -500,21 +505,30 @@ export class AccountService {
                 );
 
             if (conflict) {
-                return account;
+                getLogger(['activitypub']).warn(
+                    'WebFinger handle @{username}@{host} for {apId} is already held by another account, leaving host unchanged',
+                    {
+                        username: account.username,
+                        host: nextHost,
+                        apId: account.apId.href,
+                    },
+                );
+
+                return;
             }
         }
 
-        const updated = account.setWebfingerHost(nextHost);
         try {
-            await this.accountRepository.save(updated);
+            await this.accountRepository.updateWebfingerHost(
+                account.id,
+                nextHost,
+            );
         } catch (err) {
             if (isDuplicateEntryError(err)) {
-                return account;
+                return;
             }
             throw err;
         }
-
-        return updated;
     }
 
     /**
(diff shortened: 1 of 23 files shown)


src/lookup-helpers.ts
```diff
@@ -115,6 +128,114 @@ export async function lookupActorProfile(
...
+                normalizedApId.host;
+
+            if (normalizedSelf.href !== normalizedApId.href) {
+                return { type: 'unavailable' };
+            }
+        }
+
+        if (
+            typeof webfingerData.subject !== 'string' ||
+            !webfingerData.subject.startsWith('acct:')
```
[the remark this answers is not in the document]
louisghost: In practice WebFinger subjects are emitted as lowercase `acct:` - including by Ghost and Mastodon

src/lookup-helpers.ts
```diff
@@ -115,6 +128,114 @@ export async function lookupActorProfile(
...
+        const subjectHost = normalizeWebfingerHost(subjectParts[1]);
+        if (!subjectHost) {
+            return { type: 'default' };
+        }
+
+        if (subjectHost === actorHost) {
+            return { type: 'default' };
+        }
+
+        return { type: 'custom', host: subjectHost };
```
sagzy: **Blocking: the subject host is trusted without verifying it at that domain**

This reads the WebFinger `subject` from the actor's own server and, if the host differs from the actor host, returns it as a custom handle. The self-link check above confirms the actor server is talking about the right actor, but it does not confirm that the *claimed domain* agrees. Any server can put whatever it likes in `subject`.

Concrete example. An attacker runs `evil.example` and serves:

```json
// GET https://evil.example/users/index
{ "id": "https://evil.example/users/index", "preferredUsername": "index", ... }

// GET https://evil.example/.well-known/webfinger?resource=acct:index@evil.example
{
  "subject": "acct:index@victim-site.com",
  "links": [{ "rel": "self", "type": "application/activity+json",
              "href": "https://evil.example/users/index" }]
}
```

The self link equals the actor id, so this returns `{ type: 'custom', host: 'victim-site.com' }`. The attacker only needs to send one Follow or Like to any account on this instance; `ensureByApId` creates the row with `webfinger_host = 'victim-site.com'`. From then on feeds, notifications and search render the attacker as `@index@victim-site.com`, the handle of a real Ghost site.

This is why Mastodon does a second lookup before trusting a differing subject domain (`ActivityPub::FetchRemoteActorService#verified_webfinger?`). In pseudocode:

```ts
const first = await lookupWebFinger(`acct:${username}@${actorHost}`);
const [subjUser, subjHost] = parseAcct(first.subject);

if (subjHost === actorHost) {
    return selfLink(first) === actorId ? 'default' : 'unavailable';
}

// The subject claims a different domain. Ask THAT domain.
const second = await lookupWebFinger(`acct:${subjUser}@${subjHost}`);
const [confirmedUser, confirmedHost] = parseAcct(second.subject);

if (confirmedUser !== subjUser || confirmedHost !== subjHost) return 'unavailable';
if (selfLink(second) !== actorId) return 'unavailable';

return { type: 'custom', host: subjHost };
```

In the example, the second lookup goes to `victim-site.com`, which returns the real site's actor. The self link does not match the attacker's id, so the claim is rejected. For the real case in the issue, `https://onolan.org/.well-known/webfinger?resource=acct:john@onolan.org` redirects to `john.onolan.org` and returns the matching self link, so it passes. Fedify's `lookupWebFinger` follows redirects, so this works out of the box.

src/account/account.service.ts
```diff
@@ -133,6 +134,10 @@ export class AccountService {
         }
 
         const data = await mapActorToExternalAccountData(potentialActor);
+        data.webfinger_host = await this.resolveWebfingerHostForExternal(
```
sagzy: **Blocking: persisting a host on external rows makes our own WebFinger endpoint answer for remote accounts**

This is the first place an external row gets a non-NULL `webfinger_host`. Before this PR that column was only ever set on internal accounts (the custom handle domain feature), and two readers depend on that invariant even though neither is in this diff:

- `src/http/api/webfinger.controller.ts:58` calls `accountRepository.getByWebfingerHandle(username, host)` *before* checking sites, with no internal-only filter. After the backfill, this request on any Ghost host:

  ```
  GET /.well-known/webfinger?resource=acct:john@onolan.org
  ```

  returns a 200 JRD whose self link points at `https://john.onolan.org/.ghost/activitypub/users/index`, a server we do not run. Before the PR it returned 404. Combined with the spoofing comment in `lookup-helpers.ts`, our endpoint would answer `acct:index@victim-site.com` with the attacker's actor id, served from our own domain.

- `createInternalAccount` (`account.service.ts:554`) uses the same lookup to reject "already claimed" handles. A publisher who used to federate from `www.blog.example` gets `webfinger_host = 'blog.example'` stored here, because `normalizeWebfingerHost` strips `www.`. When they later move to Ghost(Pro) as `blog.example`, site creation throws:

  ```
  WebFinger handle @index@blog.example is already claimed
  ```

Whatever else changes, both lookups need to be scoped to internal accounts, for example by joining `users` in `getByWebfingerHandle` or adding an internal-only variant. Note that `getByWebfingerHandle` also OR-matches the last path segment of `ap_id`, which is `index` for every Ghost actor, so the exposure is wider than the username column alone.

src/account/account.service.ts
```diff
@@ -422,6 +434,89 @@ export class AccountService {
...
+    private async resolveWebfingerHostForExternal(
+        username: string,
+        apId: URL,
+    ): Promise<string | null> {
+        const resolution = await resolveExternalWebfingerHost(username, apId);
+        if (resolution.type === 'custom') {
+            return resolution.host;
+        }
+        if (resolution.type === 'default') {
+            return normalizeWebfingerHost(apId.host);
```
sagzy: **Blocking: storing the actor's own host overloads a column whose NULL already means "default"**

The `default` branch persists `normalizeWebfingerHost(apId.host)` so later calls can skip the lookup. But NULL already means "use the actor host" everywhere this column is read, and the existing internal-account code deliberately stores NULL in that case:

```ts
// setWebfingerHost, existing code in this file
if (normalizedHost === fallbackHost) {
    const updated = account.setWebfingerHost(null);
    ...
}
```

Giving the column two meanings changes behaviour for every consumer that assumed "non-NULL means custom": the WebFinger controller and onboarding check, `hasWebfingerHandleConflict`, the `UNIQUE (username, webfinger_host_hash)` index, and the `COALESCE(webfinger_host, domain)` views. It also creates new unique-index collisions: two rows for one actor seen as `https://www.example.com/...` and `https://example.com/...` both normalise to `example.com` and collide, and stale rows whose remote handle was later reassigned (Threads, Bridgy Fed) hold the slot forever.

src/account/account.service.ts
```diff
@@ -422,6 +434,89 @@ export class AccountService {
...
 
+    /**
+     * Resolve the WebFinger handle host for a remote actor.
+     *
+     * Returns the custom subject host when present. When WebFinger confirms the
+     * actor host is canonical, returns that host so callers can persist it and
+     * skip re-resolving on later ensureByApId calls. Returns null only when
+     * lookup fails (caller should leave any existing value unchanged).
+     */
+    private async resolveWebfingerHostForExternal(
```
sagzy: **Blocking: domain blocks stop matching accounts once this stores a host (bites in Explore / Recommendations / Blocks)**

`src/http/api/views/explore.view.ts:19`, `recommendations.view.ts:78` and `blocks.view.ts:11` join domain blocks on `COALESCE(accounts.webfinger_host_hash, accounts.domain_hash)`. But blocks are written using the actor hostname (`account.repository.knex.ts:171`, `domain: event.getDomain().hostname`), and the feed and moderation filters compare against `accounts.domain_hash`. The two sides only agreed because external rows never had a `webfinger_host`.

Walkthrough:

1. A user blocks the domain from John's profile page. Stored row: `domain_blocks.domain = 'john.onolan.org'`.
2. The backfill stores `webfinger_host = 'onolan.org'` on John's row.
3. Explore compares `SHA2('onolan.org')` against `SHA2('john.onolan.org')`, finds no match, and the blocked account reappears in Explore and Recommendations while the feed still hides it.

The same mismatch hits any "default" actor whose host starts with `www.`. Storing only custom hosts narrows this to genuine custom domains, and those three views should probably key on `domain_hash` like the other filters.

src/account/account.service.ts
```diff
@@ -110,7 +111,7 @@ export class AccountService {
     async getByApId(id: URL): Promise<Account | null> {
         const account = await this.accountRepository.getByApId(id);
         if (account) {
-            return account;
+            return this.refreshExternalWebfingerHostIfMissing(account);
```
sagzy: **Should fix: an untimed network request on the inbox hot path, retried forever**

`getByApId` and `ensureByApId` run for every inbound Create, Like, Announce and Follow (`dispatchers.ts:250`, `follow.handler.ts:48`, `post.service.ts:136` once per mention). Before this PR a known account was one indexed read. Now, while `webfinger_host` is NULL, each call does DNS resolution, a private-address check and an HTTPS GET to the remote before the activity is processed. Nearly every existing external row is NULL, so this is one outbound request per known actor across the whole fleet.

Two things make it worse than a one-off cost:

1. When the lookup does not succeed, nothing is persisted, so the same actor is re-fetched on every activity indefinitely. That includes any actor whose self link is not byte-equal to its id (Bridgy Fed, Lemmy communities), any host without WebFinger, any host with a port, and any row that lost the duplicate-key retry in `createExternalAccount`. A chatty account on a dead instance costs one timed-out fetch per inbound message, forever.

2. There is no timeout. `getWebFingerLookupOptions()` passes only `allowPrivateAddress`, while the existing WebFinger fetch in `validateWebfingerHost` (line ~319 in this file) already sets one:

   ```ts
   signal: AbortSignal.timeout(5000),
   ```

   Fedify's `lookupWebFinger` accepts the same option:

   ```ts
   function getWebFingerLookupOptions() {
       return {
           allowPrivateAddress: ...,
           signal: AbortSignal.timeout(5000),
       };
   }
   ```

   Without it, a remote that accepts the TCP connection and never sends headers blocks the Pub/Sub handler for undici's default header timeout, the push times out, and redelivery repeats the same lookup.

src/account/account.service.ts
```diff
@@ -422,6 +434,89 @@ export class AccountService {
...
+                );
+
+            if (conflict) {
+                return account;
+            }
+        }
+
+        const updated = account.setWebfingerHost(nextHost);
+        try {
+            await this.accountRepository.save(updated);
```
sagzy: **Should fix: this `save()` overwrites concurrent profile updates**

`refreshExternalWebfingerHost` loads the entity, awaits the network call, then saves. `accountRepository.save` is an unconditional full-row UPDATE of `name`, `bio`, `username`, `avatar_url`, `banner_image_url`, `custom_fields` and `webfinger_host` with no version check (`account.repository.knex.ts:105`). `setWebfingerHost` just spreads `...this`, so every profile field in `updated` is the pre-network snapshot.

Timeline on two Cloud Run instances:

```
t0  instance A: Create from actor X   → ensureByApId loads row (name = "Old")
t1  instance B: Update from actor X   → updateAccountByApId saves name = "New"
t2  instance A: WebFinger returns     → save(updated) writes name = "Old" back
```

The read path performed no writes before this PR, so the race is new. Removing the lazy backfill from the read path removes the race. If a write must stay here, update only the `webfinger_host` column rather than the whole row.

src/account/account.service.ts
```diff
@@ -579,10 +674,20 @@ export class AccountService {
...
-                    throw error;
+                if (existingAccount) {
+                    return existingAccount;
                 }
-                return existingAccount;
+
+                // Likely a webfinger_host uniqueness conflict — retry without
+                // the custom host rather than failing account creation.
+                if (dataToInsert.webfinger_host) {
+                    return this.createExternalAccount({
```
sagzy, line 760: **Should fix: the duplicate-key retry silently drops the host and hides conflicts**

On any duplicate-entry error not explained by `ap_id`, this recurses with `webfinger_host: null`. The comment says "Likely" because the catch cannot tell which unique index fired. Nothing is logged, there is no conflict check on the create path, and the branch has no test.

Consequence: when two actors resolve to the same `(username, host)`, the first to insert keeps the handle regardless of which one actually answers it. The second is stored with NULL and re-enters the backfill loop on every activity. It also hides exactly the spoofing case, where a fake row already holds the slot and the legitimate account loses its custom handle without a trace.

The create path already knows the host before inserting, so a pre-check reads more clearly than catch-and-recurse:

```ts
const conflict = await this.accountRepository.hasWebfingerHandleConflict(
    data.username, host, 0,
);
if (conflict) {
    logger.warn('WebFinger handle already held, storing default', { apId, host });
    host = null;
}
```

Once the second-lookup verification exists, a verified custom host should arguably win the conflict, because the row it collides with is the one that fails re-verification.

src/http/api/views/account.follows.view.ts
```diff
@@ -406,12 +407,22 @@ export class AccountFollowsView {
                         continue;
                     }
 
+                    const actorId = new URL(followsActor.id);
+                    const resolution = await resolveExternalWebfingerHost(
```
sagzy: **Should fix: one more untimed request per unknown follower, inside an API request**

This loop runs inside `handleGetAccountFollows` in the account controller. For every item not already in `accounts` it already does a sequential `lookupObject`, and now also awaits `resolveExternalWebfingerHost`, with no timeout, and the result is never persisted. The remote `CollectionPage` size is passed through unsliced, so the page is as big as the remote decides.

One follower on a host whose WebFinger hangs stalls the entire response, because the `try/catch` only skips the item after the `await` returns. The admin followers list then times out instead of rendering the other items, and the same cost repeats on every page view.

src/http/api/views/account.view.ts
```diff
@@ -235,20 +239,22 @@ export class AccountView {
...
-
-            if (externalAccount) {
-                ({ followedByMe, followsMe, blockedByMe, domainBlockedByMe } =
-                    await this.getRequestUserContextData(
-                        context.requestUserAccount.id,
-                        externalAccount.id,
-                        new URL(externalAccount.ap_id).hostname,
-                    ));
-            }
+        const storedAccount = await this.db('accounts')
```
sagzy, line 239: **Should fix: this query can never match, so the stored-host branch below is dead**

`viewByApIdRemote` is only reached when `viewByApId` already queried `accounts` by the same `ap_id_hash` and found nothing (line 167). So `storedAccount` is always null here, the "prefer a previously persisted host" branch on line 286 never runs, and neither does the `followedByMe` block it guards. What the change actually does is add one DB query per anonymous remote profile view and then always fall through to a live WebFinger request whose result is not persisted.

This is also a place where the view re-implements service logic and skips the service's conflict checks, so the profile page can render a handle the service refused to store. Views should read `webfinger_host` from the row and fall back to the actor host, which is what `viewById` and `viewByHandle` already do via `getAccountHandleHost`.

sagzy (changes requested):

sagzy:
@louisghost I had a bit of back and forth with my agent when reviewing this PR and decided to post comments directly through that chat to keep most of the context in the review comments. So apologies for the Claude-ish language, but hopefully comments are readable and have sufficient context - if not, let me know!

src/lookup-helpers.ts
louisghost: Correct, and fixed in 05b96e7 — thank you, this was the most important comment on the PR.

Implemented close to your pseudocode: a differing subject domain triggers a second lookup against that domain, whose subject has to round-trip to the same `acct:` and whose self link has to equal the original actor id. `onolan.org` passes via the redirect to `john.onolan.org`; the `victim-site.com` case fails because that host answers with the real site's actor.

Two details on top of the sketch:

- Actor ids are compared through a `canonicalActorId` normalisation rather than byte-equality, so a trailing slash or a differing host case in the second response doesn't reject a legitimate claim.
- The return type distinguishes "no custom host" from "lookup failed" (`default` vs `unavailable`). Collapsing those means a transient WebFinger failure clears a correct stored host, which is the regression this PR is meant to prevent.

Both legs carry `AbortSignal.timeout(5000)`. Accept, reject and timeout cases are covered in `lookup-helpers.unit.test.ts`.

src/account/account.service.ts
louisghost: Correct, and fixed in 05b96e7. `getByWebfingerHandle` now `innerJoin`s `users`, so it only resolves accounts we host. That covers both readers you flagged — the WebFinger controller and the `createInternalAccount` claim check — without touching either call site.

The `index` path-segment OR-match you pointed out is precisely why fixing it in the query beat filtering per caller: every Ghost actor shares that segment, so any missed call site would have been broadly exposed.

`account.repository.knex.integration.test.ts` now asserts that an external row carrying a matching `webfinger_host` is not returned.

src/account/account.service.ts
louisghost: Correct, and fixed in 05b96e7. Only verified custom hosts are stored now — the actor host is never written back as a "we checked this" marker, so `NULL` keeps its single meaning of "use the actor host" and external accounts stay out of the `UNIQUE (username, webfinger_host_hash)` index. That removes the `www.example.com` / `example.com` collision and the stale-slot problem along with it.

The trade-off is that a "no custom host" result is no longer cached, so it re-resolves on the next actor `Update`. That was only worth caching because resolution used to sit on the read path; now that it doesn't, the saved lookup isn't worth giving the column two meanings.

src/account/account.service.ts
louisghost: Diagnosis correct and fixed in 05b96e7 — but deliberately not via the remedy in the last line, because keying those views on `domain_hash` breaks a shipped feature.

Blocks are recorded against whichever host the reader was shown. Someone who blocks a domain from a profile page displaying a custom handle gets `domain_blocks.domain = 'onolan.org'`, so keying Explore, Recommendations and Blocks on `domain_hash` alone makes *that* block stop matching — the mirror image of the bug you described. That's blocking-by-displayed-handle-domain from #1821; collapsing to `domain_hash` failed three existing tests, in `explore.view.integration.test.ts`, `recommendations.view.integration.test.ts` and `blocks.view.integration.test.ts`.

The joins now match *either* host, through a shared `domainBlockMatchesAccount` helper in `src/moderation/domain-blocks.ts` so the three views can't drift apart again. Both directions are pinned: your walkthrough (block by actor host, then store a custom host) by a new test in `explore.view.integration.test.ts`, and the #1821 direction by the existing three.

src/account/account.service.ts
louisghost: Correct, and fixed in 05b96e7. The lazy backfill is gone from `getByApId` and `ensureByApId`, so a known account is one indexed read again and the "re-fetched forever" case you describe can't arise on the inbox path at all.

Resolution now happens when an external account is created, and on actor `Update` — the only federation signal we get for a handle-domain change, since the custom host lives in WebFinger rather than on the actor document. `getWebFingerLookupOptions()` sets `AbortSignal.timeout(5000)`, so both the original lookup and the new verification leg are bounded.

One consequence worth surfacing rather than burying: remote accounts that set a custom domain before the sending instance runs this code keep `webfinger_host = NULL` until their next profile edit, which then federates an `Update` and corrects them. There's no backfill migration — noted in the PR description.

src/account/account.service.ts
louisghost: Correct, and fixed in 05b96e7 — both ways you suggested, since they're complementary. The read-path write is gone, and the refresh that remains writes only the `webfinger_host` column through a new `accountRepository.updateWebfingerHost`, so it can no longer carry a pre-network snapshot of `name`, `bio` and friends into an `UPDATE`. A failed lookup leaves the stored value untouched.

A later commit also stops `save()` from writing `webfinger_host` at all, so the reverse race is closed too: a concurrent profile save that loaded the row before the refresh completed can no longer write a stale null back over the host that `updateWebfingerHost` just stored. Intentional host changes (external refresh and internal `setWebfingerHost`) go through that column-only writer.

src/account/account.service.ts
louisghost: Correct, and fixed in 05b96e7. Both write paths pre-check with `hasWebfingerHandleConflict` and log a warning naming the handle and the actor id before falling back, so a lost claim is traceable instead of silent. The refresh path leaves the existing host unchanged on conflict rather than clearing it.

The `catch` in `createExternalAccount` still exists, but it now covers only the narrow race — handle free at check time, taken before the insert — and logs when it drops the host.

On your closing suggestion that a verified host should win the conflict: I left "first writer keeps the handle" in place. Eviction would mean writing to a *different* account's row from an inbound-activity path, and now that a custom host has to be confirmed by the claimed domain, the spoofing case can't take the slot to begin with; what's left is mostly genuine duplicates. Happy to change it if you'd rather verified-wins, but it seemed worth keeping that out of this PR.

src/http/api/views/account.follows.view.ts
louisghost: Correct, and fixed in 05b96e7. The per-follower `resolveExternalWebfingerHost` call is gone. Unknown actors render from `actorId.host` and pick up the custom host once `ensureByApId` ingests them, which is the right place for a network call and a persisted result.

Accounts already in `accounts` were never affected here — that path reads `webfinger_host` from the row — so removing the lookup only costs a correct handle on the first render of an actor we've never seen.

src/http/api/views/account.view.ts
louisghost: This is the one comment I don't think holds, and I made the PR worse by acting on it — calling that out explicitly since the diff moved twice here.

`viewByApId` doesn't query `accounts` directly. It goes through `getAccountByQuery`, which inner-joins `users`:

```ts
return this.db('accounts')
    .innerJoin('users', 'users.account_id', 'accounts.id')
    .select(...baseSelect, ...countSelects)
    .where(query)
    .first();
```

So it only ever matches accounts we host, and **every** remote profile view falls through to `viewByApIdRemote` — including accounts we ingested long ago and already hold a verified `webfinger_host` for. `storedAccount` is normally non-null there, so neither it nor the `followedByMe` block it guards was dead. The existing test `'should be able to view an external account by its AP ID'` returns null only because `lookupObject` is mocked out, which is easy to misread as the row not existing.

I took the comment at face value in 05b96e7 and deleted the stored-host read. The result was that known remote profiles paid a WebFinger round trip on every page load — two, given the verification added in this PR — and fell back to the actor host whenever either leg failed. Feeds and posts join `author_webfinger_host` from the database and were unaffected, so the same account could render as `@john@onolan.org` in a feed and `@john@john.onolan.org` on the profile that feed links to, depending on whether a lookup happened to succeed.

b045b19 restores it and does what your last paragraph asked, which was right: the row lookup selects `username` and `webfinger_host`, and the handle comes from `getAccountHandleHost` when the row exists. It also moved out of the `requestUserAccount` branch, since the handle needs the row whether or not the viewer is signed in. The stored `username` is preferred over the actor's `preferredUsername` so the handle matches every other surface byte for byte.

A later commit also drops the live WebFinger fallback for actors we've never stored — they render from the actor host until ingest or an actor `Update` persists a verified custom host, matching the follower-list path. Tests cover both: stored host for a known remote, and actor-host fallback when there is no row.

src/lookup-helpers.ts
louisghost: Following up on my reply above, because the verification chain I described had a hole — and it came from the pseudocode, so it's worth flagging rather than quietly fixing. Closed in 31d4ca3.

The sketch parses `subjUser` out of the first subject, looks up `acct:${subjUser}@${subjHost}`, and checks `confirmedUser !== subjUser`. Every comparison is against the subject's own local-part; nothing ties it back to the username the first lookup was made with. That's fine in Mastodon, which displays the subject's local-part, but we render the handle from the actor's `preferredUsername`, so we were verifying one handle and displaying another.

Concretely: an actor with `preferredUsername: alice` answers its WebFinger with `subject: acct:bob@custom.example`, `custom.example` confirms `bob` points back at that actor, and we store `custom.example` and render `@alice@custom.example` — a handle nobody verified and which need not resolve at all. On a shared custom handle host it can be somebody else's, and via `UNIQUE (username, webfinger_host_hash)` the row also takes the real Alice's slot under the first-writer-wins behaviour discussed on the other thread.

`resolveCustomWebfingerHost` now rejects a subject whose local-part differs from the username it was called with, before spending the confirming lookup. Keeping `preferredUsername` as the single source for the name seemed better than switching to the subject's local-part, since every other surface already renders from the stored username. Test covers it.

src/account/account.service.ts
louisghost: Correcting my reply above: scoping this to the three views was wrong, and your follow-up review was right that the rest of the enforcement path still keyed on `domain_hash` alone. Fixed properly in 31d4ca3.

Every comparison now goes through `src/moderation/domain-blocks.ts`, which grew two more helpers — `accountMatchesDomain` for the queries that start from a blocked domain, and `isAccountDomainBlocked` for views that already hold the reader's blocked domains. Converted along with the original three views: feed fan-out in `filterUsersForPost`, `canInteractWithAccount`, account search, the profile's own `domainBlockedByMe`, the follower and following lists, and the block-time teardown of follows, feed entries and notifications.

I also moved `getRequestUserContextData` onto the account id rather than a hostname handed in by the caller, since passing one host to a check that has to consider two is how this drifts again. That drops an argument.

Both directions are now pinned in `moderation.service.integration.test.ts`: an account blocked by its actor host stays filtered after a custom host is stored, and one blocked by its custom host is filtered too.

For the record, my earlier point stands on the narrow question — keying these on `domain_hash` alone would have broken #1821 — but it was the wrong conclusion to draw from it. The right fix was to make every surface match both, not to fix the three that had a test failing.

louisghost:
Two more findings addressed, in e6ccc9f and e049e5d. They look related but are separate bugs, and only the first belongs to this PR.

**Follower lists reported moderation state for the wrong identity** (e6ccc9f). For an actor not yet in `accounts`, the handle was rendered from the id on the document `lookupObject` returned, while the domain block was looked up using the host of the URL the remote collection listed it under. Those agree until the fetch redirects to a canonical actor id on another domain, and then the row shows one identity and the block state describes another.

I did not simply switch the check to the resolved host, because that trades one wrong answer for another: a reader who blocked the domain the collection listed would then be told the account is unblocked. Neither host is authoritative and there is no stored row to arbitrate, so a block on either counts, which is the rule the rest of `domain-blocks.ts` already applies to the actor host and the custom handle host. `isAccountDomainBlocked` is now variadic over the hosts an account is known under.

One detail: the comparison uses `hostname`, not the `host` the handle is built from, because `domain_blocks` rows are written from `event.getDomain().hostname` and never carry a port. Matching on `host` would miss every block against an actor served on a non-default port.

**The blocked accounts list read other people's domain blocks** (e049e5d). `getBlockedAccounts` joined `domain_blocks` on the domain alone, with nothing constraining `blocker_id`, so any account's block on that domain set `blocked_domain` and the viewer was told `domainBlockedByMe: true` for a domain they had never blocked — and could infer from that that somebody else on the instance had.

This one predates the PR; the join on `main` has the same gap, with a `COALESCE` where the shared helper now sits. What this PR changed is that matching two hashes instead of one `COALESCE` value widens it for accounts carrying a custom host, which is presumably how it surfaced. `explore.view.ts` and `recommendations.view.ts` already scoped their equivalent joins by `blocker_id`; this one had been missed. It is a separate commit so it can be cherry-picked to `main` on its own if you would rather not carry it here.

Both have tests. `isAccountDomainBlocked` also picked up unit coverage in `src/moderation/domain-blocks.unit.test.ts`, which needs no database.

louisghost:
@coderabbitai full review

louisghost:
Correction to my earlier note on `e049e5d`: that commit scoped `getBlockedAccounts` with a `blocker_id` join condition, but matching both hashes still fan-out when the viewer had blocked both of an account's domains — two join rows, two cards.

`85df8f1` replaces that join with an `EXISTS` subquery scoped to the viewer. Same properties (viewer-only, either-host match), without the duplicate. The PR body is updated to match.

louisghost:
@coderabbitai full review

louisghost:
@coderabbitai full review

src/http/api/views/account.view.ts
```diff
@@ -269,11 +276,42 @@ export class AccountView {
...
+
+            const username = actor.preferredUsername?.toString();
+            if (!actor.id || !username) {
+                return getHandle(actor);
+            }
+
+            const resolution = await resolveCustomWebfingerHost(
+                username,
+                actor.id,
+            );
```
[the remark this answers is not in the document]
louisghost: Fixed in [035d6be](https://github.com/TryGhost/ActivityPub/pull/2163/commits/035d6be83b3f5f4a17240c6d912415b31186ba78)

louisghost:
@coderabbitai full review

louisghost:
> @louisghost I had a bit of back and forth with my agent when reviewing this PR and decided to post comments directly through that chat to keep most of the context in the review comments. So apologies for the Claude-ish language, but hopefully comments are readable and have sufficient context - if not, let me know!

@sagzy Thank you very much! I let my agent reply to your comments just so that it can have the full context on a re-review. I initially had a hard time discerning through the review noise as I anticipated a simple bug fix, and tried to keep the PR lean, but after your review my understanding improved and the PR turned out to be quite sensitive.

I went through a series of hardening commits, but I guess we need to draw the line somewhere, and I think that the PR is in a good shape for review 2.

Thank you for your assistance so far!

